PR #33, opened 2020-03-17
lists.less: Adjust caption font size for list-items

flourish86:
The font sizes of list item output preview in the original monitoring module is a little smaller than in the current module.

I thought it would be a bad idea since it makes the eye focus more on the headings. and looks a little cleaner.

## Original Icinga monitoring module

Screenshot 2020-03-17 at 16 41 59

## Current

Screenshot 2020-03-17 at 17 02 51

### DIFF
--- a/public/css/lists.less
+++ b/public/css/lists.less
@@ -91,9 +91,15 @@
     p {
       display: inline-block;
     }
+
+    &.plugin-output, .plugin-output {
+      font-size: 11/12em;
+      line-height: 1.5*12/11em;
+    }
   }
 
   header {
+    align-items: baseline;
     display: flex;
     justify-content: space-between;
 


lippserd:
I think you've mixed current with original 😆 

Could you add a screenshot of Icinga DB with and without this change? I'm not quite sure which state the current screenshot is showing.

flourish86:
> I think you've mixed current with original 😆
> 
> Could you add a screenshot of Icinga DB with and without this change? I'm not quite sure which state the current screenshot is showing.

Liaarr! No, kidding. Was my fault, I posted some totally unrelated screenshot.

Here's a screenshot of the change:

Screenshot 2020-03-17 at 17 04 56

flourish86:
The fonts seemed a little too small for me, so I re-adjusted them again.

Also now only plugin-output and lists are now smaller, since the monospaced font is larger to our standard font.

## Services
with plugin output that has been resized.

Screenshot 2020-03-18 at 10 55 49

## Donwtimes
With default font. Font size has been reset.

Screenshot 2020-03-18 at 10 55 32

PR #40, opened 2020-03-27
Ellipsize downtime list caption

### LINKED ISSUE #34: Ellipsisize downtime list caption
Screenshot 2020-03-17 at 16 45 53

flourish86:
fixes #34


flourish86:
With this commit ellipsisizing will with non-markup output.

Maybe markdown output should have different markup, so we can address it differently with css.

nilmerg:
The master currently does not render block elements in downtime comments anymore. It's now only flow content that's rendered. Please have a look at this again with this in mind. Thanks.

PR #69, opened 2020-04-21
HistoryListItem: Fix title for state change

### LINKED ISSUE #68: Hard state changes when it should not
Hi, I've been testing icingadb for a while now, and it works great. But I have one concern:

Why is icingadb reporting a hard-state change when only the soft-state changed once (1/3)? 
Screenshot 2020-04-03 at 22 18 54

In my opinion this should be a soft-state change twice.

Version: rc1

flourish86:
resolves #68

### DIFF
--- a/library/Icingadb/Widget/ItemList/HistoryListItem.php
+++ b/library/Icingadb/Widget/ItemList/HistoryListItem.php
@@ -231,7 +231,8 @@ class HistoryListItem extends CommonListItem
 
                 break;
             case 'state_change':
-                $state = $this->item->state === 'hard' ? $this->item->state->hard_state : $this->item->state->soft_state;
+                $state = $this->item->state === 'hard' ?
+                    $this->item->state->hard_state : $this->item->state->soft_state;
                 if ($state === 0) {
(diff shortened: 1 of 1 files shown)


nilmerg:
And, while you're at it, please rebase with master.

nilmerg:
Are you still going to fix this?

library/Icingadb/Widget/ItemList/HistoryListItem.php
```diff
@@ -231,10 +231,20 @@ protected function assembleTitle(BaseHtmlElement $title)
 
                 break;
             case 'state_change':
-                if ($this->item->state->state_type === 'hard') {
-                    $title->add(t('Hard state changed'));
+                $state = $this->item->state === 'hard' ?
+                    $this->item->state->hard_state : $this->item->state->soft_state;
```
nilmerg, line 235: If you need to break a line with ternary operators (? :) please break each result with its respective operator.

```suggestion
                $state = $this->item->state === 'hard'
                    ? $this->item->state->hard_state
                    : $this->item->state->soft_state;
```

PR #75, opened 2020-05-04
CSS: Fix sort direction control layout

flourish86:


nilmerg:
This doesn't need to be fixed here and not this way. I've created https://github.com/Icinga/ipl-web/pull/3 instead.

PR #95, opened 2020-05-25
lists.less: Fix downtime-list caption ellipsisizing

### LINKED ISSUE #34: Ellipsisize downtime list caption

flourish86:
fixes #34

### DIFF
--- a/public/css/lists.less
+++ b/public/css/lists.less
@@ -135,7 +135,7 @@
   padding-right: 1em;
 }
 
-.common .list-item {
+:not(.downtime-list).common .list-item {
   .caption {
     .line-clamp();
   }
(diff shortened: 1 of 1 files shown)


public/css/lists.less
```diff
@@ -135,7 +135,7 @@
   padding-right: 1em;
 }
 
-.common .list-item {
+:not(.downtime-list).common .list-item {
   .caption {
```
nilmerg, line 139: I think this can be solved without an exclusion, but rather with an inclusion. It's only plugin output after all that's supposed to have multiple lines. It also solves the issue that plain comments are still not ellipsized.

```suggestion
.common .list-item {
  .caption .plugin-output {
```

PR #111, opened 2020-07-16
common.less: Break long plugin output with no whitespace

flourish86:


flourish86:
Fixes layout break when outputs have long text portions with no white space.

Screenshot 2020-07-16 at 10 11 58

nilmerg (approved):

PR #120, opened 2020-08-12
Change label of note url

### LINKED ISSUE #67: Action and note urls should use parts of their url as label
Currently action/note urls are shown this way: 
Bildschirmfoto von 2020-04-21 11-16-14

The monitoring module shows actions the same, but notes differently:
Bildschirmfoto von 2020-04-21 11-15-54

I propose to show all of them as it's done in the monitoring module for notes. Probably with a maximum char length though.

ghost:
fixes #67

### DIFF
--- a/library/Icingadb/Widget/Detail/ObjectDetail.php
+++ b/library/Icingadb/Widget/Detail/ObjectDetail.php
@@ -75,10 +75,10 @@ class ObjectDetail extends BaseHtmlElement
                     'title' => t('Link opens in new window')
                 ]) . (string) new Text(" " . $url),
                 [
-                    'renderer' => array(
+                    'renderer' => [
                         'NavigationItemRenderer',
                         'escape_label' => false
-                    ),
+                    ],
                     'target'   => '_blank',
                     'url'      => $url
                 ]
@@ -228,10 +228,10 @@ class ObjectDetail extends BaseHtmlElement
                     'title' => t('Link opens in new window')
                 ]) . (string) new Text(" " . $url),
                 [
-                    'renderer' => array(
+                    'renderer' => [
                         'NavigationItemRenderer',
                         'escape_label' => false
-                    ),
+                    ],
                     'target'   => '_blank',
                     'url'      => $url
                 ]
@@ -256,6 +256,7 @@ class ObjectDetail extends BaseHtmlElement
         if (empty($content)) {
             return null;
         }
+        
         array_unshift($content, Html::tag('h2', t('Notes')));
 
         return $content;
--- a/library/Icingadb/Widget/Detail/ObjectDetail.php
+++ b/library/Icingadb/Widget/Detail/ObjectDetail.php
@@ -225,7 +225,6 @@ class ObjectDetail extends BaseHtmlElement
                 (string) new Icon('forward', [
                     'title' => t('Link opens in new window')
                 ]) . (string) new Text(" " . $url),
-
                 [
                     'renderer' => array(
                         'NavigationItemRenderer',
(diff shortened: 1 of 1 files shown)


library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -70,12 +71,14 @@ protected function createActions()
 
         foreach ($this->compatObject->getActionUrls() as $i => $url) {
             $navigation->addItem(
-                sprintf(t('Action %d'), $i + 1),
+                (string) new Icon('forward', [
+                    'title' => t('Link opens in new window')
+                ]) . (string) new Text(" " . $url),
                 [
-                    'renderer' => [
+                    'renderer' => array(
```
nilmerg, line 78: Why did you change this back to the legacy syntax? We prefer brackets instead of this so please revert this.

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -221,9 +224,14 @@ protected function createNotes()
 
         foreach ($this->compatObject->getNotesUrls() as $i => $url) {
             $navigation->addItem(
-                sprintf(t('Note %d'), $i + 1),
+                (string) new Icon('forward', [
+                    'title' => t('Link opens in new window')
+                ]) . (string) new Text(" " . $url),
                 [
-                    'renderer' => 'NavigationItemRenderer',
+                    'renderer' => array(
```
nilmerg, line 231: Use brackets here please.

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -248,7 +256,6 @@ protected function createNotes()
         if (empty($content)) {
             return null;
         }
-
```
nilmerg, line 251: Why did you remove this empty line? We require explicit empty lines after control structures.

nilmerg (changes requested):

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -70,7 +71,9 @@ protected function createActions()
 
         foreach ($this->compatObject->getActionUrls() as $i => $url) {
             $navigation->addItem(
-                sprintf(t('Action %d'), $i + 1),
+                (string) new Icon('forward', [
+                    'title' => t('Link opens in new window')
+                ]) . (string) new Text(" " . $url),
```
nilmerg, line 76: Maybe we should only show the url without query here. Urls can get very long and the path itself is usually the most interesting part anyway. So please only show the url path in the label.

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -221,9 +224,14 @@ protected function createNotes()
 
         foreach ($this->compatObject->getNotesUrls() as $i => $url) {
             $navigation->addItem(
-                sprintf(t('Note %d'), $i + 1),
+                (string) new Icon('forward', [
+                    'title' => t('Link opens in new window')
+                ]) . (string) new Text(" " . $url),
```
nilmerg, line 229: Here, also only the path please.

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -248,7 +256,7 @@ protected function createNotes()
         if (empty($content)) {
             return null;
         }
-
+
```
nilmerg, line 259: Does this only look like as if there are spaces or are there actually spaces? If so, please remove them, this should be an *empty* line.

PR #122, opened 2020-08-20
Bugfix/right datatype gives an error

### LINKED ISSUE #119: Multiple right Data type gives an error
Multiple right data types gives an error, the wrong not.

## 1. Reproduce error:
1. Create hosts.conf in custom.conf.d directory.
2. Set variable notes_url and action_url as a string, described like in the docs.
3. Restart Icinga web and see the error in the host.
fehlerfall1

## 2. Expected error in this case:
1. Create hosts.conf in custom.connf.d directory.
2. Set variable notes_url as a string inside an array, not like in the docs.
fotomitErfolg1

## 3. Expected behavior:
Expected no error with the multiple data type string.
fehlerfallmit1  

Expected an error with an array, for example on notes_url, meaning using wrong data type.
fehlerMit3

## 4. Environment

- Icinga DB Web version : 1.0.0-rc1
- Icinga Web 2 version : 2.8.1
- Web browser : Chrome
- Icinga 2 version : r2.12.0-rc1-1
- Icinga DB version : v1.0.0-rc1
- PHP version used : 7.4.8
- Server operating system and version : ubuntu 18.04

pmloei:
fixes #119

### DIFF
--- a/library/Icingadb/Compat/CompatHost.php
+++ b/library/Icingadb/Compat/CompatHost.php
@@ -12,6 +12,7 @@ class CompatHost extends Host
 
     private $legacyColumns = [
         'host_action_url'                       => ['path' => ['action_url', 'action_url']],
+        'action_url'                            => ['path' => ['action_url', 'action_url']],
         'host_address'                          => ['path' => ['address']],
         'host_address6'                         => ['path' => ['address6']],
         'host_alias'                            => ['path' => ['display_name']],


library/Icingadb/Compat/CompatObject.php
```diff
@@ -94,6 +94,10 @@ public function fetchCustomvars()
 
     public function __get($name)
     {
+        if ($name == 'action_url') {
+            $name = 'host_action_url';
+        }
+
```
nilmerg, line 100: That's unfortunately the wrong solution. Please register `action_url` as a legacy column instead.

nilmerg (changes requested):
Left a comment. Also, please remove the commits from this branch that belong to the other issue.

pmloei:
> Left a comment. Also, please remove the commits from this branch that belong to the other issue.

Commits removed.

nilmerg (approved):

PR #153, opened 2020-11-24
Remove obsolete event box

nilmerg:
@flourish86 You've made this obsolete with https://github.com/Icinga/icingadb-web/commit/c69cd5120492cdefee378093c9d314777586e164, are there any plans to utilize this widget somehow else or is it safe to remove it?


flourish86:
@nilmerg As discussed. Here is the original mockup of the detail view. It included a concept to "intelligently" list recent historic events instead of just simply the current plugin output.

Redesign

PR #170, opened 2021-03-17
Introduce implementation for hook X509/Sni in icingadb

raviks789:
The hook X509/Sni is implemented for icingadb in this commit.

### DIFF
--- a/library/Icingadb/ProvidedHook/X509/Sni.php
+++ b/library/Icingadb/ProvidedHook/X509/Sni.php
@@ -7,7 +7,6 @@ use Icinga\Module\Icingadb\Common\Auth;
 use Icinga\Module\Icingadb\Common\Database;
 use Icinga\Module\Icingadb\Model\Host;
 use Icinga\Module\X509\Hook\SniHook;
-use ipl\Orm\Compat\FilterProcessor;
 use ipl\Web\Filter\QueryString;
 
 class Sni extends SniHook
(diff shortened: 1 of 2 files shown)


library/Icingadb/ProvidedHook/X509/Sni.php
```diff
@@ -0,0 +1,56 @@
...
+            'host_name'      => 'name',
+            'host_address'   => 'address',
+            'host_address6'  => 'address6'
+        ];
+
+        $queryHost = $queryHost->columns($hostStatusCols);
+
+        $this->applyRestrictions($queryHost);
+
+        $filter = Filter::where('name', 'dummy-1');
```
nilmerg, line 37: You forgot to remove your test :wink:

library/Icingadb/ProvidedHook/X509/Sni.php
```diff
@@ -0,0 +1,56 @@
...
+        if ($filter !== null) {
+            $queryString = $filter->toQueryString();
+            $filterCondition = QueryString::parse($queryString);
+            FilterProcessor::apply($filterCondition, $queryHost);
```
nilmerg, line 41: The `FilterProcessor` class is now not necessary anymore. Use this instead:

```suggestion
            $queryHost->filter($filterCondition);
```

nilmerg (changes requested):

library/Icingadb/ProvidedHook/X509/Sni.php
raviks789: Thank you, I will remove this test case.

library/Icingadb/ProvidedHook/X509/Sni.php
raviks789: It makes sense. I will make the changes.

nilmerg (approved):

PR #195, opened 2021-05-05
Add object inspection

### LINKED ISSUE #117: Object inspection like in the director
For the "inspect" button you need to install director, this should be a feature of icingaweb2 and not director. Some user dont want to use and maintain director package just for this feature

Inspect feature is part of icingaweb2 and not director only

### LINKED ISSUE #181: Show shell check-command in ui
While debugging failing check results it is always a good start to look at the actual check command to understand better what it does. In Icinga 1 this was easy, as the configuration page provides the actual check command that is executed by icinga. Not so in icinga2. It is possible to get that command from the json http api with something like 

```shell
curl -ksu root:pass https://icinga:5665/v1/objects/services
```

But that is really not a good solution, especially if API access isn't enabled / available.

I would like to have a button somewhere in this UI, preferable at the top with the other buttons that shows the shell command.

image

Figuring out the command from the api as described above is cumbersome and error prone, but seems to be the only way right now.

nilmerg:
Includes Icinga 2 API results and Icinga Redis results
Screenshot from 2021-05-11 16-48-07 Screenshot from 2021-05-11 16-47-51

resolves #117
resolves #181

### DIFF
--- a/application/controllers/HostController.php
+++ b/application/controllers/HostController.php
@@ -5,6 +5,8 @@
 namespace Icinga\Module\Icingadb\Controllers;
 
 use Icinga\Exception\NotFoundError;
+use Icinga\Module\Icingadb\Command\Object\GetObjectCommand;
+use Icinga\Module\Icingadb\Command\Transport\CommandTransport;
 use Icinga\Module\Icingadb\Common\CommandActions;
 use Icinga\Module\Icingadb\Common\HostLinks;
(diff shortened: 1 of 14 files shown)


nilmerg:
Well, icons in tabs don't work without changes in the tab implementation as labels get always escaped. I'll use `Source` as label, as it's my favorite as well.

PR #198, opened 2021-05-07
Re-add host tabs to service detail

### LINKED ISSUE #115: ServiceList: Clicking on host makes service item active
## Describe the bug

When clicking the host on a service list item the item (which represents a service) will become "active". This is wrong since this results in the host detail view to appear.

## To Reproduce

1. Go to a service list
2. Click the host name in the item's title

## Expected behavior

Two options:
1) The service list item should not be shown as active
2) In the legacy monitoring module there was always a host tab in services. This could be re-introduced and after a click on the host name could be activated instead. This would be the preferred version

## Screenshots

Screenshot 2020-08-04 at 16 08 45

## Your Environment

Screenshot 2020-08-04 at 16 10 51

nilmerg:
Screenshot from 2021-05-07 15-42-49

When in service detail view, the `Host` and `Services` tabs are displayed now. This is now the same as in the monitoring module, with a single exception:

* The `History` tab doesn't change its context anymore
  * Clicking on `Host` while looking at a service and then clicking on `History` will show the service history, **not** the host history
  * This means that there is no click-path from the service detail view to the host's history in any way
  * I did this as I think the tabs should all keep their context if they happen to have multiple meanings
    * Splitting the `History` into `Host History` and `Service History` is another solution, but that's too much tabs then, right?

Also, what I'm unsure of is whether we should show the `Services` tab when in service context:

a) Keep it as is now (i.e. The `Services` tab is always there)
b) Only show the `Services` tab when in host context, hide it completely in service context

@lippserd, @flourish86 Please share your thoughts regarding the `History` and `Services` tab behavior

resolves #115

### DIFF
--- a/application/controllers/HostController.php
+++ b/application/controllers/HostController.php
@@ -8,6 +8,7 @@ use Icinga\Exception\NotFoundError;
 use Icinga\Module\Icingadb\Common\CommandActions;
 use Icinga\Module\Icingadb\Common\HostLinks;
 use Icinga\Module\Icingadb\Common\Links;
+use Icinga\Module\Icingadb\Common\ServiceLinks;
 use Icinga\Module\Icingadb\Model\History;
 use Icinga\Module\Icingadb\Model\Host;
 use Icinga\Module\Icingadb\Model\Service;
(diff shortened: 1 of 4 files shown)


nilmerg:
Superseded by #200

PR #201, opened 2021-05-11
Perfdata: Add support for more UoMs

### LINKED ISSUE #71: Support for more UoMs in perfdata
Check plugins currently have only a limited set of UoMs (Units of measurement) available which Icinga understands.

With Icinga v2.13 support for more is added. We have to support them as well.

## Additional context
https://github.com/Icinga/icinga2/issues/7225

Yonas-net:
resolves #71

### DIFF
--- a/library/Icingadb/Util/PerfData.php
+++ b/library/Icingadb/Util/PerfData.php
@@ -6,7 +6,6 @@ namespace Icinga\Module\Icingadb\Util;
 
 use Icinga\Exception\ProgrammingError;
 use Icinga\Module\Monitoring\Object\Service;
-use Icinga\Util\Format;
 use Icinga\Web\Widget\Chart\InlinePie;
 use InvalidArgumentException;
 
@@ -143,7 +142,7 @@ class PerfData
      */
     public function isSeconds()
     {
-        return in_array($this->unit, array('s', 'ms', 'us'));
+        return $this->unit === 's';
     }
 
     /**
@@ -153,7 +152,7 @@ class PerfData
      */
     public function isTemperature()
     {
-        return in_array($this->unit, array('°c', '°f'));
+        return in_array($this->unit, array('C', 'F', 'K'));
     }
 
     /**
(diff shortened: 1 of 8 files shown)


Yonas-net:
Depends on [#4369](https://github.com/Icinga/icingaweb2/pull/4369)

library/Icingadb/Util/PerfData.php
```diff
@@ -307,8 +451,8 @@ protected function parse()
         $parts = explode(';', $this->perfdataValue);
 
         $matches = array();
-        if (preg_match('@^(-?\d+(\.\d+)?)([a-zA-Z%°]{1,2})$@u', $parts[0], $matches)) {
-            $this->unit = strtolower($matches[3]);
+        if (preg_match('@^(-?\d+(\.\d+)?)([a-zA-Z%°]{1,3})$@u', $parts[0], $matches)) {
+            $this->unit = $matches[3];
             $this->value = self::convert($matches[1], $this->unit);
```
nilmerg, line 456: I'm not sure if values need to be converted here still. If we expect the input data as being normalized, there's no need to. Also, doesn't the formatting in method `format()` expect the normalized values from Icinga?

library/Icingadb/Util/PerfData.php
```diff
@@ -449,15 +643,67 @@ protected function format($value)
...
+            return Format::ohms($value);
+        }
+        if ($this->isGrams()) {
+            return Format::grams($value);
+        }
+        if ($this->isLiters()) {
+            return Format::liters($value);
+        }
         if ($this->isTemperature()) {
             return (string)$value . strtoupper($this->unit);
```
nilmerg, line 704: The `strtoupper` call isn't necessary anymore.

library/Icingadb/Util/PerfData.php
```diff
@@ -449,15 +643,67 @@ protected function format($value)
         if ($this->isPercentage()) {
```
nilmerg, line 643: * Consider to use a `switch (true) { case $this->isPercentage(): ... }` here
  * There are now so much similar control structures, it makes sense to condense them
* Also please put a space between casts and variables everywhere (i.e. `(string) $value`)

library/Icingadb/Util/Format.php
```diff
@@ -0,0 +1,275 @@
+<?php
+
+namespace Icinga\Module\Icingadb\Util;
+
+use DateTime;
+
+class Format
```
nilmerg, line 7: Please name it `PerfDataFormat` and remove all unused methods. (e.g. `secondsByMonth`)

nilmerg (changes requested):

library/Icingadb/Util/PerfData.php
Yonas-net: Ohh, you're right. I wonder if this is also the case for the thresholds, so are the thresholds normalized as well or does this only affect the plugin output?

library/Icingadb/Util/PerfData.php
```diff
@@ -166,14 +165,159 @@ public function isPercentage()
...
+
+    /**
+     * Get whether this data's value is in watt seconds
+     *
+     * @return bool
+     */
+    public function isWattSeconds()
+    {
+        return in_array($this->unit, Format::getTimeWattPrefix()['WattSeconds']);
+    }
```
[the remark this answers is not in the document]
Yonas-net: As I understand it, they are different units. For example these units ``mWm, uWm`` are normalized to ``Wm`` and not to ``Wh``. This also affects the different ampere units``Ah, Am, As``. But I might be wrong.

library/Icingadb/Util/PerfData.php
nilmerg: That is a question for @Al2Klimov :wink:

library/Icingadb/Util/PerfData.php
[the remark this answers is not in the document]
nilmerg: the [table](https://github.com/Icinga/icinga2/pull/8759/files#diff-b5d3095868e71d7e5d627a517bab1e8e40198c49d0536f43e36b60ede5488278R677-R679) shows that for these units the metric backends get *watt-hours*, so it should be `Wh` for them if normalized.

library/Icingadb/Util/PerfData.php
Al2Klimov: Reading about _plugin output normalization_ the first time. 😉

And – regarding the thresholds:

https://github.com/Icinga/icinga2/blob/63b53bfda92930b2332035860082dbcaa5cff9e7/test/icinga-perfdata.cpp#L46-L53

Yonas-net: > Reading about _plugin output normalization_ the first time. 😉

My mistake.

library/Icingadb/Util/PerfDataFormat.php
```diff
@@ -0,0 +1,138 @@
+<?php
+
+/* Icinga DB Web | (c) 2021 Icinga GmbH | GPLv2+ */
```
nilmerg, line 3: Please change this to `GPLv2` (Remove the plus). I know, the other PerfData files also use it, but it's wrong as well.

library/Icingadb/Util/PerfDataFormat.php
```diff
@@ -0,0 +1,138 @@
...
+
+        if ($absValue < 60) {
+            return self::formatForUnits($value, self::$secondPrefix, self::$generalBase);
+        } elseif ($absValue < 3600) {
+            return sprintf('%0.2f m', $value / 60);
+        } elseif ($absValue < 86400) {
+            return sprintf('%0.2f h', $value / 3600);
+        }
+
+        // TODO: Do we need weeks, months and years?
```
nilmerg, line 107: I'd say, after many weeks and months and years: No.

Don't keep the comment, remove it :wink:

nilmerg (changes requested):
Nice! Two minor changes left and it's ready!

nilmerg (dismissed):
LGTM! Can be merged once tested with actual normalized data. (Blocked by ~~https://github.com/Icinga/icinga2/pull/8761~~ https://github.com/Icinga/icingadb/pull/308 in the meantime)

nilmerg (approved):

Issue #143, opened 2020-10-20
Downtime List: Provide additional information in detailed view mode


flourish86:
## Describe the bug
There's no difference in display, when switching between view mode default and detailed.

## Expected behavior
There should be additional information in detailed view mode, like multiline comment text in the caption or alternatively no detailed view mode.

### Current Default view mode

![Screen Shot 2020-10-20 at 15 26 15](https://user-images.githubusercontent.com/6977434/96592674-ee05c400-12e8-11eb-9e32-de00e3ea351a.png)

### Detailed view mode

![Screen Shot 2020-10-20 at 15 26 09](https://user-images.githubusercontent.com/6977434/96592703-f78f2c00-12e8-11eb-8315-87c5814b81e5.png)

flourish86:
I wondered at first, because currently the downtime list items show 2 lines of caption in the default view.

The screenshots only show one, so I guess, that it was intended that way originally. Was this an intentional change?

I think it would look more balanced with only one. Also the visual would align perfectly with the lower height.

This is the current version (ignore the crazy data)

![Screen Shot 2021-05-12 at 10 53 29](https://user-images.githubusercontent.com/6977434/117947420-780e8c00-b310-11eb-82ee-4d9db98879a6.png)

nilmerg:
That's probably a side-effect of the clean-up I made in #184. Though, the elllipsis looks fine on my end:

![Screenshot from 2021-05-12 11-16-25](https://user-images.githubusercontent.com/16668527/117950639-ae99d600-b313-11eb-99ab-5e640fdabddf.png)

nilmerg:
Oh, ehm, and I don't see why downtimes should not behave the same as other lists with comments/output. (regarding the 2 lines)

flourish86:
This can be closed, right?

nilmerg:
Seems so

PR #202, opened 2021-05-11
CSS: Display 5 lines of caption in .downtime-list.detailed

flourish86:
refs #143

### DIFF
--- a/public/css/list/downtime-list.less
+++ b/public/css/list/downtime-list.less
@@ -40,11 +40,6 @@
 
 .downtime-list .list-item,
 .downtime-detail .list-item {
-  .caption {
-    .text-ellipsis();
-    height: 1.5em;
-  }
-
   .progress {
     height: 1/6em;
     margin-bottom: 1/3em;
@@ -92,16 +87,4 @@
       font-size: 1em;
     }
   }
-
-  .caption {
-    .text-ellipsis()
-  }
 }
-
-.downtime-list.detailed .list-item {
-  .caption {
-    height: auto;
-    max-height: 4.5em;
-  }
-}
-
--- a/public/css/list/item-list.less
+++ b/public/css/list/item-list.less
@@ -61,9 +61,7 @@
       text-overflow: ellipsis;
       overflow: hidden;
 
-      .plugin-output {
-        .line-clamp();
-      }
+      .line-clamp();
     }
 
     header {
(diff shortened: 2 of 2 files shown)


public/css/list/downtime-list.less
```diff
@@ -98,10 +99,10 @@
...
-.downtime-list.detailed .list-item {
+.item-list.downtime-list.detailed .list-item {
   .caption {
     height: auto;
-    max-height: 4.5em;
+    max-height: 7.5em;
+    .line-clamp(5);
   }
 }
-
```
nilmerg, line 107: Please don't remove empty lines from the end of files.

nilmerg (changes requested):
Have you taken a look at the comment list as well? Does it suffer from the same?

public/css/list/item-list.less
```diff
@@ -58,12 +58,8 @@
 
     .caption {
       height: 3em;
-      text-overflow: ellipsis;
-      overflow: hidden;
 
-      .plugin-output {
-        .line-clamp();
-      }
```
nilmerg, line 66: If the line clamp is now not (only) applied to the plugin output anymore, please also adjust the reset on line 90 so that it has an effect again.

public/css/list/item-list.less
```diff
@@ -169,4 +165,9 @@
     max-height: 7.5em; /* 5 lines */
     position: relative;
   }
+
+  .caption,
+  .caption > * {
```
nilmerg, line 170: Why is this also applied to the plugin output but only in detailed mode? Doesn't it need to be applied always anyway? If not, please remove.

public/css/list/list-item.less
```diff
@@ -23,6 +23,8 @@
   }
 
   .caption {
+    overflow: hidden;
+
```
nilmerg, line 27: Please remove this. For one thing this is the style section of this file and this doesn't belong here. Also, you've removed the same instruction in file `item-list.less`. Why not keep it there instead?

nilmerg (changes requested):

public/css/list/list-item.less
flourish86: ✅

public/css/list/item-list.less
flourish86: I'll change the `*` from line 170 to `.plugin-output`.

`.line-clamp` itself should always be applied, but in detailed it's 4 lines, otherwise 2.

public/css/list/item-list.less
flourish86: I think, we can remove the `line-clamp("reset")` property from line 88, since we always have a max no of lines to display, don't we?

It's just 4 in .detailed instead of 2.

nilmerg (approved):

PR #203, opened 2021-05-11
Action-List: Add selected items to the bottom status-bars

### LINKED ISSUE #105: Bottom StatusBars in Lists
1) Status bars in lists currently lack the information how many items are selected.
2) Therefore also downtime and comment lists should have a status bar, as well as every other list with multi select capabilities

And this is how it should look in host lists.

No items selected

Desktop

With selected items

Desktop Copy

Yonas-net:
resolves #105

### DIFF
--- a/library/Icingadb/Common/DetailActions.php
+++ b/library/Icingadb/Common/DetailActions.php
@@ -48,6 +48,9 @@ trait DetailActions
         $this->getAttributes()
             ->registerAttributeCallback('class', function () {
                 return $this->getDetailActionsDisabled() ? null : 'action-list';
+            })
+            ->registerAttributeCallback('data-icinga-multiselect-count-label', function () {
+                return $this->getDetailActionsDisabled() ? null : t('%d Item(s) selected');
             });
 
         return $this;
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -212,25 +212,38 @@ div.show-more {
   }
 }
 
-.status-bar {
+.footer {
   display: flex;
-  justify-content: space-between;
-  align-items: center;
-  padding: .25em 1em;
-  box-shadow: 0 -1px 0 0px @gray-lighter;
-  font-size: .857em;
   z-index: 1000;
-  line-height: 1.7;
+
+  .selection-count {
+    flex: 1 1 auto;
+
+    .selected-items {
+      font-size: 1.25em;
+      color: @gray;
+    }
+  }
+
+  .status-bar, .selection-count {
+    font-size: .857em;
+    box-shadow: 0 -1px 0 0px @gray-lighter;
+    padding: .25em 1em;
+    line-height: 1.7;
+  }
+}
+
+.status-bar {
+  margin-left: auto;
 
   .item-count {
-    margin-left: auto;
     font-size: 1.25em;
     color: @gray;
   }
 
   .state-badges {
-    margin: 0;
-    margin-left: .417em;
+    display: inline-block;
+    margin: 0 0 0 .417em;
   }
 }
 
--- a/public/js/action-list.js
+++ b/public/js/action-list.js
@@ -55,6 +55,8 @@
         event.stopImmediatePropagation();
         event.stopPropagation();
 
+        var container = $list.closest('.container');
+
         if ($list.is('[data-icinga-multiselect-url]')) {
             if (event.ctrlKey || event.metaKey) {
                 $item.toggleClass('active');
@@ -82,14 +84,28 @@
                 $list.find('.list-item.active').removeClass('active');
                 $item.addClass('active');
             }
+
+            // For items that do not have a bottom status bar like Downtimes, Comments...
+            if (! container.children('.footer').length) {
+                container.append('<div class="footer" data-action-list-automatically-added></div>');
+            }
         } else {
             $list.find('.list-item.active').removeClass('active');
             $item.addClass('active');
         }
 
         $activeItems = $list.find('.list-item.active');
+        var footer = container.children('.footer');
 
         if ($activeItems.length === 0) {
+            if (footer.length) {
+                if (typeof footer.data('action-list-automatically-added') !== 'undefined') {
+                    footer.remove();
+                } else {
+                    footer.children('.selection-count').remove();
+                }
+            }
+
             if (_this.icinga.loader.getLinkTargetFor($target).attr('id') === 'col2') {
                 _this.icinga.ui.layout1col();
             }
@@ -104,6 +120,20 @@
                 });
 
                 url = $list.attr('data-icinga-multiselect-url') + '?(' + filters.toArray().join('|') + ')';
+            }
+
+            if ($list.is('[data-icinga-multiselect-url]')) {
+                if (! footer.children('.selection-count').length) {
+                    footer.prepend('<div class="selection-count"></div>');
+                }
+
+                var label = $list.data('icinga-multiselect-count-label').replace('%d', $activeItems.length);
+                var selectedItems = footer.find('.selection-count > .selected-items');
+                if (selectedItems.length) {
+                    selectedItems.text(label);
+                } else {
+                    footer.children('.selection-count').append('<span class="selected-items">' + label + '</span>');
+                }
             }
 
             _this.icinga.loader.loadUrl(
@@ -135,6 +165,16 @@
                     '[data-icinga-detail-filter="' + detailUrl.query.slice(1) + '"]'
                 ).removeClass('active');
             }
+
+            var footer = $list.closest('.container').children('.footer');
+
+            if (footer.length) {
+                if (typeof footer.data('action-list-automatically-added') !== 'undefined') {
+                    footer.remove();
+                } else {
+                    footer.children('.selection-count').remove();
+                }
+            }
         }
     };
 
@@ -161,6 +201,21 @@
                 $list.find(
                     '[data-icinga-detail-filter="' + detailUrl.query.slice(1) + '"]'
                 ).addClass('active');
+            }
+        }
+
+        if ($list.length && $list.is('[data-icinga-multiselect-url]')) {
+            var $activeItems = $list.find('.list-item.active');
+
+            if ($activeItems.length) {
+                if (! $target.children('.footer').length) {
+                    $target.append('<div class="footer" data-action-list-automatically-added></div>');
+                }
+
+                var label = $list.data('icinga-multiselect-count-label').replace('%d', $activeItems.length);
+                $target.children('.footer').prepend(
+                    '<div class="selection-count"><span class="selected-items">' + label + '</span></div>'
+                );
             }
         }
     };


Yonas-net:
Bildschirmfoto 2021-05-12 um 09 07 16

Bildschirmfoto 2021-05-12 um 09 08 22

public/js/action-list.js
```diff
@@ -82,14 +82,25 @@
                 $list.find('.list-item.active').removeClass('active');
                 $item.addClass('active');
             }
+
+            // For items that do not have a bottom status bar like Downtimes, Comments...
+            if ($('#col1.container').find('.footer > .status-bar').length === 0) {
+                $('#col1').append('<div id="automatically-added" class="footer"><div class="status-bar"></div></div>');
```
nilmerg, line 88: * Please use `$list.closest('.container')` instead of `$('#col1.container')` or `$('#col1')`
  * It's not guaranteed that the list the user is interacting with resides in `#col1`
* Please don't check for/add the `.status-bar` container, just check if there's a footer
  * Add the selection count to its own container instead
* Don't add an id to the footer
  * An id is only required based on the content. No content, no id

public/js/action-list.js
```diff
@@ -106,6 +117,13 @@
                 url = $list.attr('data-icinga-multiselect-url') + '?(' + filters.toArray().join('|') + ')';
             }
 
+            // If the item exists, remove it first before reading it
+            if (bottomStatusBar.find('.selected-items').length !== 0) {
+                bottomStatusBar.find('.selected-items').remove();
```
nilmerg: Instead of removing, why don't you update it? If you're already checking its existence, this is the more efficient way I think.

public/js/action-list.js
```diff
@@ -106,6 +117,13 @@
...
+            }
+
+            bottomStatusBar.prepend('<span class="selected-items">' + $activeItems.length + ' Items selected</span>');
```
nilmerg, line 120: The class doesn't exist. Only use classes you'll need. Though, if you're not adding this to the status bar but directly to the footer, the class is required. So.. add it ;)

public/js/action-list.js
```diff
@@ -106,6 +117,13 @@
                 url = $list.attr('data-icinga-multiselect-url') + '?(' + filters.toArray().join('|') + ')';
             }
 
+            // If the item exists, remove it first before reading it
+            if (bottomStatusBar.find('.selected-items').length !== 0) {
```
nilmerg: Consider processing this as a boolean: `if (seq.length) {` or `if (! seq.length) {`

nilmerg (changes requested):
Left some comments.

Also, I'd like you to enable localization. For this, render the label to use on the server and pass it to JS by use of a `data-` attribute. (e.g. `data-icinga-multiselect-count-label`) Include a placeholder for the number in the localized label so that translators can move it. You'll then need to replace that by use of JS.

library/Icingadb/Widget/BaseListItem.php
```diff
@@ -39,6 +39,9 @@ public function __construct($item, BaseItemList $list)
         $this->list = $list;
 
         $this->addAttributes($this->baseAttributes);
+        $this->addAttributes([
+            'data-icinga-multiselect-count-label'   => t('%d Items selected')
+        ]);
```
nilmerg, line 44: Take a look at the `DetailActions` trait. There, register this attribute in `initializeDetailActions()`. Make sure to also use a callback to avoid registering it if it's not necessary.

public/js/action-list.js
```diff
@@ -82,14 +82,26 @@
                 $list.find('.list-item.active').removeClass('active');
                 $item.addClass('active');
             }
+
+            // For items that do not have a bottom status bar like Downtimes, Comments...
+            if (! $list.closest('.container').find('.footer').length) {
```
nilmerg, line 87: You should probably also use `children` instead of `find`. The latter traverses the entire DOM tree down, which may lead to false-positives.

public/js/action-list.js
```diff
@@ -82,14 +82,26 @@
...
+                $list.closest('.container').append('<div class="footer automatically-added"></div>');
+            }
         } else {
             $list.find('.list-item.active').removeClass('active');
             $item.addClass('active');
         }
 
         $activeItems = $list.find('.list-item.active');
+        var bottomStatusBar = $('.footer');
```
nilmerg, line 96: This way you not only find a `.footer` in the current container, but also in the menu and other containers. Selecting elements with such a broad rule isn't good practice. Preferably, do it the same as on line 87. You may also consider now to cache the result of `$list.closest('.container')` :wink:

public/js/action-list.js
```diff
@@ -82,14 +82,26 @@
                 $list.find('.list-item.active').removeClass('active');
                 $item.addClass('active');
             }
+
+            // For items that do not have a bottom status bar like Downtimes, Comments...
+            if (! $list.closest('.container').find('.footer').length) {
+                $list.closest('.container').append('<div class="footer automatically-added"></div>');
```
nilmerg, line 88: Don't add a class if you don't require it for styling. If you need to flag a particular DOM element, use data attributes instead. (e.g. `<div class="footer" data-action-list-automatically-added></div>`)

public/js/action-list.js
```diff
@@ -106,6 +118,17 @@
                 url = $list.attr('data-icinga-multiselect-url') + '?(' + filters.toArray().join('|') + ')';
             }
 
+            if (! bottomStatusBar.find('.selection-count').length) {
+                bottomStatusBar.prepend('<div class="selection-count"></div>');
+            }
+
+            var $label = $activeItems.data('icinga-multiselect-count-label').replace('%d', $activeItems.length);
```
nilmerg, line 125: Shouldn't `$activeItems` be `$list` instead? The label attribute is on the list after all.

public/css/common.less
```diff
@@ -212,15 +212,37 @@ div.show-more {
   }
 }
 
+.footer {
+  div:first-child:nth-last-child(2) ~ div {
```
nilmerg, line 216: I don't quite get what this is meant for. Could you explain this or better, simplify it?

nilmerg (changes requested):

library/Icingadb/Widget/BaseListItem.php
Yonas-net: However, I wanted to add this attribute to the individual selectable list items rather than to the `ul`.

public/js/action-list.js
Yonas-net: If the attribute is not added in `DetailActions`, then it is so correct because the individual active elements have this attribute and not the list.

public/css/common.less
Yonas-net: Since you told me to use custom containers for these selected elements, I had to do a lot of research on how to style the footer without changing the previous style. I then found out that you can style it any way you want based on the number of elements.  This line only changes the style of the footer if two div elements are added, otherwise it keeps the previous style. Why I do this is because now we add and remove a new container with JS which we need for these selected elements as well as a fixed container for the status bar, then this technique is very handy. I couldn't find any simplified other solution unfortunately.

Yonas-net: If you would consider that as simplified, it also works that way:
```css
div:nth-last-child(2) ~ div ...
```

library/Icingadb/Widget/BaseListItem.php
nilmerg: To the list items? The text is *about* the list items in the context of the list where they're a part of.

public/js/action-list.js
nilmerg: But it doesn't make any sense that the list items have this label, the label isn't even associated to a particular list item, it only shows a number. Besides, `$activeItems` isn't a list item to begin with, it's a jQuery set of `li`.

public/css/common.less
nilmerg: I'd rather make the entire footer a flex container instead of the `.status-bar`.

So, add `display: flex;` to the footer. `flex: 1 1 auto` to the `.selection-count`. `margin-left: auto;` to the `.status-bar`.

You should also remove all flexbox related styles from `.status-bar` and apply `display: inline-block;` to the contained `.state-bagdes` list instead.

This way your selection count is also on the very left and the previous layout keeps unchanged. It's also more readable and on point.

Yonas-net: Indeed, the two dives then don't get the same width but this way also works. Ty.

library/Icingadb/Common/DetailActions.php
```diff
@@ -48,6 +48,9 @@ public function initializeDetailActions()
         $this->getAttributes()
             ->registerAttributeCallback('class', function () {
                 return $this->getDetailActionsDisabled() ? null : 'action-list';
+            })
+            ->registerAttributeCallback('data-icinga-multiselect-count-label', function () {
+                return t('%d Items selected');
```
nilmerg, line 53: Please add the disabled check as well. Just like the other callbacks do.

public/css/common.less
```diff
@@ -212,15 +212,29 @@ div.show-more {
...
-  box-shadow: 0 -1px 0 0px @gray-lighter;
-  font-size: .857em;
-  z-index: 1000;
-  line-height: 1.7;
+
+  .selection-count {
+    flex: 1 1 auto;
+
+    .selected-items {
+      margin-right: auto;
```
nilmerg, line 222: The margin has no effect, it's an inline element. Please remove.

public/css/common.less
```diff
@@ -212,15 +212,29 @@ div.show-more {
...
 
-.status-bar {
+.footer {
   display: flex;
-  justify-content: space-between;
-  align-items: center;
-  padding: .25em 1em;
-  box-shadow: 0 -1px 0 0px @gray-lighter;
-  font-size: .857em;
-  z-index: 1000;
```
nilmerg, line 222: Please re-add the `z-index`. There's no need to remove it.

nilmerg: Though, you should add it to the footer instead since it's not only the status bar anymore which is part of it.

public/css/common.less
```diff
@@ -212,15 +212,29 @@ div.show-more {
...
+    padding: .25em 1em;
+    line-height: 1.7;
+  }
+}
+
+.status-bar {
+  margin-left: auto;
 
   .item-count {
     margin-left: auto;
```
nilmerg, line 240: Since the status bar itself has the margin now, you can remove it here.

public/js/action-list.js
```diff
@@ -82,14 +82,31 @@
...
+                );
+            }
         } else {
             $list.find('.list-item.active').removeClass('active');
             $item.addClass('active');
         }
 
         $activeItems = $list.find('.list-item.active');
+        var container = $list.closest('.container');
+        var bottomStatusBar = container.find('.footer');
```
nilmerg, line 99: Why not move this further up and use it already in the check on line 87?

public/js/action-list.js
```diff
@@ -82,14 +82,31 @@
...
 
         if ($activeItems.length === 0) {
+            if (bottomStatusBar.length) {
+                if (bottomStatusBar.data('action-list-automatically-added')) {
+                    bottomStatusBar.data('action-list-automatically-added').remove();
```
nilmerg, line 104: Why is the `remove` call chained with the `data` call? This is wrong I guess.

public/js/action-list.js
nilmerg: And remember to only use `find` if you don't know or care about an elements position in the tree. The footer however should only be a direct child of the container, so `children` should suffice.

public/js/action-list.js
```diff
@@ -82,14 +82,31 @@
...
+                } else {
+                    bottomStatusBar.find('.selection-count').remove();
```
nilmerg, line 106: Again, please use `children`

public/js/action-list.js
```diff
@@ -106,6 +123,17 @@
                 url = $list.attr('data-icinga-multiselect-url') + '?(' + filters.toArray().join('|') + ')';
             }
 
+            if (! bottomStatusBar.find('.selection-count').length) {
```
nilmerg, line 126: `children` :wink:

public/js/action-list.js
nilmerg: Oh, aaand, you should rename the variable. It's the footer after all, not the status bar.

public/js/action-list.js
```diff
@@ -106,6 +123,17 @@
...
+                bottomStatusBar.prepend('<div class="selection-count"></div>');
+            }
+
+            var label = $list.data('icinga-multiselect-count-label').replace('%d', $activeItems.length);
+            if (bottomStatusBar.find('.selection-count').find('.selected-items').length) {
```
nilmerg, line 131: This would be a use-case for `find` now. If it would only be **one** `find` call, though. Please use `bottomStatusBar.find('.selection-count > .selected-items').length` instead, and cache it and re-use the result in the next line.

public/js/action-list.js
```diff
@@ -106,6 +123,17 @@
...
+                bottomStatusBar.find('.selection-count').find('.selected-items').text(label);
+            } else {
+                bottomStatusBar.find('.selection-count').append('<span class="selected-items">' + label + '</span>');
```
nilmerg, line 134: Again, `children` :wink:

nilmerg (changes requested):

library/Icingadb/Common/DetailActions.php
```diff
@@ -48,6 +48,9 @@ public function initializeDetailActions()
         $this->getAttributes()
             ->registerAttributeCallback('class', function () {
                 return $this->getDetailActionsDisabled() ? null : 'action-list';
+            })
+            ->registerAttributeCallback('data-icinga-multiselect-count-label', function () {
+                return $this->getDetailActionsDisabled() ? null : t('%d Items selected');
```
nilmerg, line 53: Please use `Item(s)` instead. Looks better if only one item is selected.

nilmerg (changes requested):
Looks mostly good now. Good job!

Though there's one last thing: An autorefresh hides the selection count again. To counteract this, please re-add it in method `onRendered` of the behavior. Maybe implement a method if it suits, which maintains the selection count.

Yonas-net:
> Maybe implement a method if it suits, which maintains the selection count.

I actually decided against implanting a new method because it's just not worth duplicating the ``onRendered`` event for just a few lines of code.

public/js/action-list.js
```diff
@@ -106,6 +122,18 @@
...
+            }
+
+            var label = $list.data('icinga-multiselect-count-label').replace('%d', $activeItems.length);
+            var selectedItems = footer.find('.selection-count > .selected-items');
+            if (selectedItems.length) {
+                selectedItems.text(label);
+            } else {
+                footer.children('.selection-count').append('<span class="selected-items">' + label + '</span>');
+            }
+
```
nilmerg, line 136: During testing I noticed that this runs also for non-multiselect lists. This has the side effect that if there's no native footer, line 126 prepends its HTML to an empty set. (`footer`) The set is empty, because the automatically added footer is only added in case it's a multiselect list.

Since it's useless anyway to show the selection count for non-multiselect lists, the entire block here should only run for multiselect lists. Then it's also guaranteed that a footer is found.

public/js/action-list.js
```diff
@@ -163,6 +191,22 @@
                 ).addClass('active');
             }
         }
+
+        var $activeItems = $list.find('.list-item.active');
+
+        if ($activeItems.length) {
+            var container = $list.closest('.container');
```
nilmerg, line 198: Here in `onRendered`, `$target` is already the container. You don't have to query that again.

public/js/action-list.js
```diff
@@ -163,6 +191,22 @@
...
+
+            if (! container.children('.footer').length) {
+                container.append('<div class="footer" data-action-list-automatically-added></div>');
+            }
+
+            var footer = container.children('.footer');
+            var label = $list.data('icinga-multiselect-count-label').replace('%d', $activeItems.length);
+
+            footer.prepend('<div class="selection-count"></div>');
+            footer.children('.selection-count').append('<span class="selected-items">' + label + '</span>');
```
nilmerg, line 208: Please combine both lines. You can include the span with the `.selection-count` div already.

public/js/action-list.js
```diff
@@ -163,6 +191,22 @@
                 ).addClass('active');
             }
         }
+
+        var $activeItems = $list.find('.list-item.active');
+
+        if ($activeItems.length) {
```
nilmerg, line 197: Show the selection count here only for multiselect lists as well.

nilmerg (changes requested):
Just did the first test, noticed some issues. See my comments for this.

Also another thing which hadn't on screen until now: Closing the right column resets the active state of the list items, but the selection-count isn't removed. Please implement this in method `onColumnClose`.

public/js/action-list.js
```diff
@@ -106,6 +122,20 @@
                 url = $list.attr('data-icinga-multiselect-url') + '?(' + filters.toArray().join('|') + ')';
             }
 
+            if (footer.length) {
```
nilmerg, line 125: That a footer exists, doesn't mean it's a multiselect list. Please check explicitly for a multiselect list.

nilmerg (changes requested):

nilmerg (approved):
LGTM!

PR #204, opened 2021-05-17
Add Service Grid

### LINKED ISSUE #118: Add Service Grid
Screenshot from 2020-08-10 13-33-30

raviks789:
resolves #118

### DIFF
--- a/application/controllers/ServicesController.php
+++ b/application/controllers/ServicesController.php
@@ -7,6 +7,7 @@ namespace Icinga\Module\Icingadb\Controllers;
 use GuzzleHttp\Psr7\ServerRequest;
 use Icinga\Module\Icingadb\Common\CommandActions;
 use Icinga\Module\Icingadb\Common\Links;
+use Icinga\Module\Icingadb\Data\PivotTable;
 use Icinga\Module\Icingadb\Model\Service;
 use Icinga\Module\Icingadb\Model\ServicestateSummary;
 use Icinga\Module\Icingadb\Util\FeatureStatus;
@@ -18,7 +19,9 @@ use Icinga\Module\Icingadb\Widget\ServiceList;
 use Icinga\Module\Icingadb\Widget\ServiceStatusBar;
 use Icinga\Module\Icingadb\Widget\ShowMore;
 use Icinga\Module\Icingadb\Widget\ViewModeSwitcher;
+use ipl\Html\Form;
 use ipl\Stdlib\Filter;
+use ipl\Web\Compat\CompatForm;
 use ipl\Web\Control\LimitControl;
 use ipl\Web\Control\SortControl;
 use ipl\Web\Filter\QueryString;
@@ -196,6 +199,139 @@ class ServicesController extends Controller
 
         $this->getDocument()->add($editor);
         $this->setTitle(t('Adjust Filter'));
+    }
+
+    public function gridSearchEditorAction()
+    {
+        $query = Service::on($this->getDb())
+            ->with([
+                'state',
+                'host',
+                'host.state'
+            ]);
+        $editor = $this->createSearchEditor($query, [
+            LimitControl::DEFAULT_LIMIT_PARAM,
+            SortControl::DEFAULT_SORT_PARAM,
+            ViewModeSwitcher::DEFAULT_VIEW_MODE_PARAM
+        ]);
+
+        $this->getDocument()->add($editor);
+        $this->setTitle(t('Adjust Grid Filter'));
+    }
+
+    public function gridAction()
+    {
+        $compact = $this->view->compact;
+
+        $db = $this->getDb();
+        $this->setTitle(t('Service Grid'));
+
+        $query = Service::on($db)->with([
+            'state',
+            'host',
+            'host.state'
+        ]);
+
+        $this->handleSearchRequest($query);
+        $gridcols = [
+            'host_name' => 'host.display_name',
+            'host_display_name' => 'host.name',
+            'service_name' => 'service.name',
+            'service_display_name' => 'service.display_name',
+            'service_handled' => 'service.state.is_handled',
+            'service_output' => 'service.state.output',
+            'service_state' => 'service.state.soft_state'
+        ];
+        $flipped = $this->params->shift('flipped', false);
+
+        $pivotFilter = (bool) $this->params->shift('problems', false) ?
+            Filter::equal('service.state.is_problem', 'y') : null;
+
+        $problemToggle = (new CompatForm())->addAttributes(['class' => 'inline']);
+
+        $problemToggle->addElement('checkbox', 'problems', [
+            'class'     => 'autosubmit',
+            'id'        => $this->getRequest()->protectId('problems'),
+            'label'     => $this->translate('Problems Only'),
+            'value'     => $pivotFilter !== null
+        ]);
+
+        $problemToggle->on(Form::ON_SUCCESS, function (Form $form) {
+            if (! $form->getElement('problems')->isChecked()) {
+                $this->redirectNow(Url::fromRequest()->remove('problems'));
+            } else {
+                $this->redirectNow(Url::fromRequest()->setParams($this->params->add('problems')));
+            }
+        });
+
+        $problemToggle->handleRequest(ServerRequest::fromGlobals());
+
+        $this->addControl($problemToggle);
+
+        $sortControl = $this->createSortControl($query, [
+            'host.name'     => $this->translate('Host Name'),
+            'service.name'  => $this->translate('Service Name')
+        ]);
+
+        if ($flipped) {
+            $xAxisCol = 'host_name';
+            $yAxisCol = 'service_name';
+        } else {
+            $xAxisCol = 'service_name';
+            $yAxisCol = 'host_name';
+        }
+
+        $pivot = (new PivotTable($query, $xAxisCol, $yAxisCol, $gridcols))
+            ->setXAxisFilter($pivotFilter)
+            ->setYAxisFilter($pivotFilter ? clone $pivotFilter : null)
+            ->setXAxisHeader($xAxisCol)
+            ->setYAxisHeader($yAxisCol);
+
+        $this->view->horizontalPaginator = $pivot->paginateXAxis();
+        $this->view->verticalPaginator = $pivot->paginateYAxis();
+        list($pivotData, $pivotHeader) = $pivot->toArray();
+        $this->view->pivotData = $pivotData;
+        $this->view->pivotHeader = $pivotHeader;
+
+        $searchBar = $this->createSearchBar($query, [
+            LimitControl::DEFAULT_LIMIT_PARAM,
+            $sortControl->getSortParam(),
+            'flipped',
+            'page'
+        ]);
+
+        $continueWith = $this->createContinueWith(
+            Url::fromPath('icingadb/services/grid')->addParams($this->getAllParams()),
+            $searchBar
+        );
+
+        if ($searchBar->hasBeenSent() && ! $searchBar->isValid()) {
+            if ($searchBar->hasBeenSubmitted()) {
+                $filter = $this->getFilter();
+            } else {
+                $this->addControl($searchBar);
+                $this->sendMultipartUpdate();
+                return;
+            }
+        } else {
+            $filter = $searchBar->getFilter();
+        }
+
+        $this->filter($query, $filter);
+
+        $this->addControl($sortControl);
+        $this->addControl($searchBar);
+        $this->view->controls = $this->controls;
+
+        if ($flipped) {
+            $this->render('grid-flipped');
+        }
+
+        if (! $searchBar->hasBeenSubmitted() && $searchBar->hasBeenSent()) {
+            $this->sendMultipartUpdate($continueWith);
+        }
+
+        $this->setAutorefreshInterval(10);
     }
 
     public function fetchCommandTargets()
--- a/application/controllers/ServicesController.php
+++ b/application/controllers/ServicesController.php
@@ -11,6 +11,7 @@ use Icinga\Module\Icingadb\Data\PivotTable;
 use Icinga\Module\Icingadb\Model\Service;
 use Icinga\Module\Icingadb\Model\ServicestateSummary;
 use Icinga\Module\Icingadb\Util\FeatureStatus;
+use Icinga\Module\Icingadb\Web\Control\ProblemToggle;
 use Icinga\Module\Icingadb\Web\Control\SearchBar\ObjectSuggestions;
 use Icinga\Module\Icingadb\Web\Controller;
 use Icinga\Module\Icingadb\Widget\Detail\MultiselectQuickActions;
@@ -19,9 +20,7 @@ use Icinga\Module\Icingadb\Widget\ServiceList;
 use Icinga\Module\Icingadb\Widget\ServiceStatusBar;
 use Icinga\Module\Icingadb\Widget\ShowMore;
 use Icinga\Module\Icingadb\Widget\ViewModeSwitcher;
-use ipl\Html\Form;
 use ipl\Stdlib\Filter;
-use ipl\Web\Compat\CompatForm;
 use ipl\Web\Control\LimitControl;
 use ipl\Web\Control\SortControl;
 use ipl\Web\Filter\QueryString;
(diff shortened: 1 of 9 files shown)


public/css/service-grid.less
```diff
@@ -0,0 +1,75 @@
+/*! Icinga Web 2 | (c) 2015 Icinga Development Team | GPLv2+ */
```
nilmerg, line 1: Please remove the license header.

And move the entire file to `public/css/view` (new directory)

application/controllers/ListController.php
```diff
@@ -0,0 +1,342 @@
...
+        $this->setupSortControl(array(
+            'servicegroup_alias'    => $this->translate('Service Group Name'),
+            'services_severity'     => $this->translate('Severity'),
+            'services_total'        => $this->translate('Total Services')
+        ), $serviceGroups, ['services_severity' => 'desc']);
+
+        $this->view->serviceGroups = $serviceGroups;
+    }
+
+    public function servicegridAction()
```
nilmerg, line 197: I think this should be a part of the `ServicesController`. (i.e. `icingadb/services/grid`)

This has the advantage that you can use nearly (minus e.g. the view-mode-switcher) the same instructions as the `indexAction` to set up the controls. (Because the grid is also using `Icinga\Module\Icingadb\Model\Service` as base model)

application/controllers/ServicesController.php
```diff
@@ -199,6 +202,85 @@ public function searchEditorAction()
         $this->setTitle(t('Adjust Filter'));
     }
 
+    public function gridAction()
+    {
+        $db = $this->getDb();
+
+        if ($this->params->has('noscript_apply')) {
+            $this->redirectNow($this->getRequest()->getUrl()->without('noscript_apply'));
+        }
```
nilmerg, line 211: We don't support no-js environments in Icinga DB Web. Remove this.

application/controllers/ServicesController.php
```diff
@@ -199,6 +202,85 @@ public function searchEditorAction()
...
+            'host.name',
+            'service.name',
+            'service.display_name',
+            'service.state.is_handled',
+            'service.state.output',
+            'service.state.soft_state'
+        ];
+
+        $filter = (bool) $this->params->shift('problems', false) ?
+            \Icinga\Data\Filter\Filter::where('service.state.is_problem', 1) : null;
```
nilmerg, line 232: Please use `ipl\Stdlib\Filter` instead.

application/controllers/ServicesController.php
```diff
@@ -199,6 +202,85 @@ public function searchEditorAction()
...
+
+        $this->filter($query);
+        $query->getSelectBase();
```
nilmerg, line 235: That's a noop. Please remove.

application/controllers/ServicesController.php
```diff
@@ -199,6 +202,85 @@ public function searchEditorAction()
...
+        $problemToggle->setUidDisabled();
+        $problemToggle->setTokenDisabled();
+        $problemToggle->setAttrib('class', 'filter-toggle inline icinga-controls');
+        $problemToggle->addElement('checkbox', 'problems', [
+            'disableHidden' => true,
+            'autosubmit'    => true,
+            'value'         => $filter !== null,
+            'label'         => $this->translate('Problems Only'),
+            'decorators'    => ['ViewHelper', ['Label', ['placement' => 'APPEND']]]
+        ]);
```
nilmerg, line 249: Please use `ipl\Html\Form` instead.

application/controllers/ServicesController.php
```diff
@@ -199,6 +202,85 @@ public function searchEditorAction()
...
+
+        $query = new SimpleQuery($query->columns($gridcols)->assembleSelect());
```
nilmerg, line 251: The `PivotTable` below should be able to handle `ipl\Orm\Query`. We don't use `SimpleQuery` in Icinga DB Web anymore.

application/controllers/ServicesController.php
```diff
@@ -199,6 +202,85 @@ public function searchEditorAction()
...
+            $pivot = $pivot
+                ->setXAxisFilter($filter)
+                ->setYAxisFilter($filter ? clone $filter : null)
+                ->setXAxisHeader('service.name')
+                ->setYAxisHeader('host.name');
+        }
+        $this->setupSortControl(array(
+            'host_display_name'     => $this->translate('Hostname'),
+            'service_display_name'  => $this->translate('Service Name')
+        ), $pivot);
```
nilmerg, line 273: You should use `$this->createSortControl()` here.

nilmerg (commented):
Left a few comments.

Remember that we only want to re-use view related stuff. That's CSS and the `.phtml` view scripts. Everything else (controller code, queries, etc) have to make use of the ipl and shouldn't rely upon the legacy code from Icinga Web 2.

The `PivotTable` class is still nearly identical to the one used in Icinga Web 2, so I didn't take ta look at it. Please make sure to adjust it so it doesn't need anything from Icinga Web 2 and can handle `ipl\Orm\Query` instances. Maybe the class itself should be one.

application/controllers/ServicesController.php
nilmerg: With that I've meant the `getSelectBase()` call only :wink:

library/Icingadb/Web/Adapter/QueryAdapter.php
```diff
@@ -0,0 +1,85 @@
+<?php
+/* Icinga Web 2 | (c) 2013 Icinga Development Team | GPLv2+ */
+namespace Icinga\Module\Icingadb\Web\Adapter;
+
+use ipl\Stdlib\Contract\Paginatable;
+use Zend_Paginator_Adapter_Interface;
+
+class QueryAdapter implements Zend_Paginator_Adapter_Interface
```
nilmerg, line 8: Why is that necessary? Is it because of the `joystickPagination` view script? If so, why don't you just copy this view script and adjust it as well to make it work with a `ipl\Stdlib\Contract\Paginatable`?

raviks789: Yes, it is because of the `joystickPagination`. Ok, I will do as you suggested.

nilmerg: Then please make it so that `PivotTable` doesn't require this. We have `ipl\Stdlib\Contract\Paginatable` as successor for that.

We don't want to use any Zend_* classes/interfaces in this module.

raviks789: Ok

nilmerg:
The grid is very slow, runs always into the execution timeout for me. That's due to a missing limit on the used database queries. Setup a default limit on the query you pass to the `PivotTable` and the slowness should be gone.

library/Icingadb/Widget/JoystickPaginator.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Icinga\Module\Icingadb\Widget;
+
+use ipl\Orm\Query;
+
+class JoystickPaginator
```
nilmerg, line 7: I guess you added this as an optimization? The thing is, I don't quite see a justification for this being outsourced. The joystick paginator is subject to change with the new style for the servicegrid, so this is also likely subject to change as well. I think, leaving the calculations in the view script is fine for now. :wink:

library/Icingadb/Data/PivotTable.php
```diff
@@ -0,0 +1,450 @@
...
+
+namespace Icinga\Module\Icingadb\Data;
+
+use Icinga\Application\Icinga;
+use Icinga\Module\Icingadb\Widget\JoystickPaginator;
+use ipl\Orm\Query;
+use ipl\Stdlib\Contract\Paginatable;
+use ipl\Stdlib\Filter;
+
+class PivotTable
```
nilmerg, line 11: Still didn't have a detailed look here, but will soon. In the meantime, please make sure the code style is correct and remove any unused parts. (e.g. the order stuff)

Code style to look out for:
* array syntax
* empty lines after control structures if the body continues
* indentation

nilmerg (changes requested):

application/controllers/ServicesController.php
```diff
@@ -199,6 +203,95 @@ public function searchEditorAction()
...
+        ]);
+
+        $gridcols = [
+            'host_name' => 'host.display_name',
+            'host_display_name' => 'host.name',
+            'service_name' => 'service.name',
+            'service_display_name' => 'display_name',
+            'service_handled' => 'service.state.is_handled',
+            'service_output' => 'service.state.output',
+            'service.state.soft_state'
```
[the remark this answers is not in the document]
raviks789: It is easier to get axis keys this way using $row->{$this->yAxisColumn} or $row->{$this->xAxisColumn} when the axis column is 'host.name'.

application/controllers/ServicesController.php
```diff
@@ -7,7 +7,9 @@
 use GuzzleHttp\Psr7\ServerRequest;
 use Icinga\Module\Icingadb\Common\CommandActions;
 use Icinga\Module\Icingadb\Common\Links;
+use Icinga\Module\Icingadb\Data\PivotTable;
 use Icinga\Module\Icingadb\Model\Service;
+use Icinga\Module\Icingadb\Model\ServiceGrid;
```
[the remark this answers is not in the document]
raviks789: Need to remove it. Was trying to build a separate Model for ServiceGrid, but realised that it was unnecessary.

application/controllers/ServicesController.php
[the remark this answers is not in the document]
nilmerg: I see. Why is `service.state.soft_state` not aliased then?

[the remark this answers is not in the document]
raviks789: I did not use alias there because in `grid.phtml` it would be easier to use `$service->state->getStateText()`. But now I have changed it. I have given alias even for `service.state.soft_state`.

library/Icingadb/Data/PivotTable.php
```diff
@@ -0,0 +1,410 @@
...
+            $xCol = explode('.', $this->gridcols[$this->xAxisColumn]);
+
+            if ($xCol[0] !== $table) {
+                $groupCols = array_unique([
+                    $this->xAxisColumn => $table . '_' . $this->gridcols[$this->xAxisColumn],
+                    $xAxisHeader => $table . '_' . $this->gridcols[$xAxisHeader]
+                ]);
+            } else {
+                $groupCols = $columns;
+            }
```
nilmerg, line 215: What's the use-case for this? You don't seem to need this?

library/Icingadb/Data/PivotTable.php
```diff
@@ -0,0 +1,410 @@
...
+
+            if ($groupCols !== null) {
```
nilmerg, line 217: `$groupCols` is never null.

library/Icingadb/Data/PivotTable.php
```diff
@@ -0,0 +1,410 @@
...
+     *
+     * @return array
+     */
+    public function toArray()
+    {
+        if (
+            ($this->xAxisFilter === null && $this->yAxisFilter === null)
+            || ($this->xAxisFilter !== null && $this->yAxisFilter !== null)
+        ) {
+            $xAxis = $this->getAxisKeys($this->queryXAxis(), $this->xAxisColumn);
```
nilmerg, line 352: You should use PDO's native fetchPairs functionality here:

```php
$xAxis = $this->queryXAxis()->getDb()->fetchPairs($this->queryXAxis()->assembleSelect());
```

Do the same everywhere where you use `$this->getAxisKeys()`.

library/Icingadb/Data/PivotTable.php
```diff
@@ -0,0 +1,410 @@
...
+    protected function queryXAxis()
+    {
+        if ($this->xAxisQuery === null) {
+            $this->xAxisQuery = clone $this->baseQuery;
+            $xAxisHeader = $this->getXAxisHeader();
+            $table = $this->xAxisQuery->getModel()->getTableName();
+            $columns = array_unique([
+                $this->xAxisColumn => $this->gridcols[$this->xAxisColumn],
+                $xAxisHeader => $this->gridcols[$xAxisHeader]
+            ]);
```
nilmerg, line 204: `array_unique` shouldn't be applied to the select columns. I've suggested to use PDO's fetchPairs functionality below. This won't work with less than two select columns. The group columns however don't need to contain duplicates, there `array_unique` is appropriate.

library/Icingadb/Data/PivotTable.php
```diff
@@ -0,0 +1,410 @@
...
+
+        $pivotData = [];
+        $pivotHeader = [
+            'cols'  => $xAxis,
+            'rows'  => $yAxis
+        ];
+
+        if (! empty($xAxis) && ! empty($yAxis)) {
+            $this->baseQuery->filter(Filter::equal($this->gridcols[$this->xAxisColumn], $xAxisKeys));
+            $this->baseQuery->filter(Filter::equal($this->gridcols[$this->yAxisColumn], $yAxisKeys));
```
nilmerg, line 382: Isn't it possible to use `$this->xAxisColumn` as column in the filter? Aliases are usually available for use in a WHERE.

library/Icingadb/Data/PivotTable.php
```diff
@@ -0,0 +1,410 @@
...
+    protected $xAxisHeader;
+
+    /**
+     * Y-axis header column
+     *
+     * @var string|null
+     */
+    protected $yAxisHeader;
+
+    protected $gridcols = [];
```
nilmerg, line 75: After I've seen what you use this for, I think it's not necessary. SQL accepts aliases as alternative in most places, so the `PivotData` class shouldn't need to know the real column name I believe.

Please try to avoid this and remove it if successful. If not successful, please tell me the reason.

application/controllers/ServicesController.php
```diff
@@ -198,6 +201,91 @@ public function searchEditorAction()
...
+
+        $this->addControl($problemToggle);
+
+        $sortControl = $this->createSortControl($query, [
+            'host.name'     => $this->translate('Host Name'),
+            'service.name'          => $this->translate('Service Name')
+        ]);
+
+        $searchBar = $this->createSearchBar($query, [
+            $sortControl->getSortParam(),
```
nilmerg, line 256: You should also add all the params specific to this action. Otherwise they show up in the search bar after using the joystick, flipping the grid or using the toggle slider.

application/controllers/ServicesController.php
```diff
@@ -198,6 +201,91 @@ public function searchEditorAction()
         $this->setTitle(t('Adjust Filter'));
     }
 
+    public function gridAction()
```
nilmerg, line 204: This action is still missing some processing for the searchbar. Try adding a filter with it, you should notice it isn't applied. The searchbar requires specific handling by an action, check what the `servicesAction` does with it and perform the same here.

nilmerg (changes requested):
Left a few comments.

There's also something weird with the ordering after `Service Name`. This should apply to the axis where the service name is being used, currently it applies to the axis where hosts are shown.

library/Icingadb/Data/PivotTable.php
```diff
@@ -0,0 +1,398 @@
...
+            $xAxisHeader = $this->getXAxisHeader();
+            $table = $this->xAxisQuery->getModel()->getTableName();
+            $xCol = explode('.', $this->gridcols[$this->xAxisColumn]);
+            $columns = [
+                $this->xAxisColumn => $this->gridcols[$this->xAxisColumn],
+                $xAxisHeader => $this->gridcols[$xAxisHeader]
+            ];
+
+            if ($xCol[0] !== $table) {
+                $groupCols = array_unique([
```
nilmerg, line 209: Please add `// TODO: This shouldn't be required. Refactor this once ipl\Orm\Query has support for group by rules!`

And add the same todo in method `queryYAxis`.

library/Icingadb/Data/PivotTable.php
```diff
@@ -0,0 +1,398 @@
...
+    protected $xAxisHeader;
+
+    /**
+     * Y-axis header column
+     *
+     * @var string|null
+     */
+    protected $yAxisHeader;
+
+    protected $gridcols = [];
```
nilmerg, line 75: Please add a docblock here.

application/controllers/ServicesController.php
```diff
@@ -18,8 +19,11 @@
 use Icinga\Module\Icingadb\Widget\ServiceStatusBar;
 use Icinga\Module\Icingadb\Widget\ShowMore;
 use Icinga\Module\Icingadb\Widget\ViewModeSwitcher;
+use ipl\Html\Form;
 use ipl\Stdlib\Filter;
+use ipl\Web\Compat\CompatForm;
 use ipl\Web\Control\LimitControl;
+use ipl\Web\Control\PaginationControl;
```
nilmerg, line 26: This is unused, please remove.

application/controllers/ServicesController.php
```diff
@@ -198,6 +202,105 @@ public function searchEditorAction()
...
+
+        if ($flipped) {
+            $this->render('grid-flipped');
+        }
+
+        if (! $searchBar->hasBeenSubmitted() && $searchBar->hasBeenSent()) {
+            $this->sendMultipartUpdate($continueWith);
+        }
+
+        $this->setAutorefreshInterval(10);
```
nilmerg, line 301: Please increase this to 30 seconds. The grid may become very large quickly, refreshing every 10th second is too often then.

nilmerg (changes requested):
* The ordering after `service.name` still doesn't work
* Clicking on `Load More` results in an error

nilmerg (changes requested):

nilmerg (approved):

nilmerg (approved):

PR #206, opened 2021-05-25
Align all controls on the same baseline

### LINKED ISSUE #130: List Controls break line
## Describe the bug

List Controls break at a viewport width of ~1290

## Suggestion

Set a max-width to sort control to make the view mode switcher to prevent the view mode switcher from going to the next line.

```max-width: 6em;``` should be appropriate.

## Screenshots

Screenshot 2020-10-19 at 17 28 50

### LINKED ISSUE #131: Compact View: Fix view mode switcher height
## Describe the bug

List view's View Mode Switcher is too tall in compact view.

## Suggestion

Maybe adjust icon size(?)

## Screenshots
Screenshot 2020-10-19 at 17 38 15

### LINKED ISSUE #132: Align control elements
## Describe the bug
Control elements should be properly aligned vertically

## Screenshots
Screenshot 2020-10-19 at 17 53 05

### LINKED ISSUE #133: Make control container height match
## Describe the bug
Control containers should have the same height. At least when there is a similar amount of rows. Row height should be unified to 30px (2.5em), which is the result of 24px (2em) element height plus 6 px (.5em) margin.

Pixel values refer to default view.

## Screenshots
Screenshot 2020-10-19 at 17 55 51

### LINKED ISSUE #141: Fix Downtime Detail Header Design
## Describe the bug

* Remove top border
* Adjust spacings
* Align to controls grid

### Current State

Screen Shot 2020-10-20 at 14 23 57

#### Apply one row grid

Control Grid One Row

#### Apply two row grid

Control Grid Two Rows

#### Whole Screen

Desktop

Desktop Copy

### LINKED ISSUE #185: Collapsible controls
Since we have a separation line between controls and content, there's a visual mis-match if multiple columns with differently sized controls are shown simultaneously.

Screenshot from 2021-04-23 10-23-26

There are already the following issues which also highlight the problem:

* Fix Downtime Detail Header Design #141
* Align control elements #132
* List Controls break line #130
* Make control container height match #133

## The Problem

I also see the visual mis-match, but I don't think this is a battle we can win. We can of course define strict heights and row separations in controls, but this way we highly limit ourselves in a way which controls we can show there. It's also nearly impossible to make this match on every screen size. The risk that this will never be fixed entirely is high and we probably fix it once, twice, a third time and then a new feature/module appears and it's broken again.

So to me this is a question of cost/effort vs. benefits and not any of the proposed solutions in the linked issues are viable.

## The Solution?

Let's grant ourselves the freedom of (nearly) doing whatever we want in controls and ignore the not aligned separation lines. :joy: I'd be fine with this, as I'm not too bothered by this *issue* and don't think of it as a sign of quality.

If that's not an option, I'd suggest to give the controls a rather low `min-height` (so that it's effectively only a single line) and only show selected controls (TBD) always. Other controls need to be expanded, in a way. We talked about this already if I correctly remember.

--

@lippserd, @flourish86 please give your thoughts on the matter.

### PR DESCRIPTION
**Pull request by flourish86**
The alignment should now works for all views.

fixes #130
fixes #131
fixes #132
fixes #133
fixes #141
closes #185

### DIFF
--- a/library/Icingadb/Widget/ShowMore.php
+++ b/library/Icingadb/Widget/ShowMore.php
@@ -40,6 +40,15 @@ class ShowMore extends BaseHtmlElement
         return $this->label ?: t('Show More');
     }
 
+    public function renderUnwrapped()
+    {
+        if ($this->resultSet->hasMore()) {
+            return parent::renderUnwrapped();
+        }
+
+        return '';
+    }
+
     protected function assemble()
     {
         if ($this->resultSet->hasMore()) {
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -153,6 +153,16 @@ div.show-more {
       height: 100%;
       padding-right: 0;
     }
+
+    .link-button {
+      min-width: 2em;
+      text-align: center;
+      .rounded-corners();
+
+      &:hover {
+        background: @gray-lighter;
+      }
+    }
   }
 
   .show-more {
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -102,7 +102,7 @@ div.show-more {
 
   .filter {
     display: flex;
-    width: 100%;
+    min-width: 100%;
 
     .search-bar {
       flex: 1 1 auto;
@@ -111,6 +111,10 @@ div.show-more {
     .search-editor-opener,
     .continue-with {
       margin-left: .5em;
+
+      &:last-child {
+        margin-right: -.5em;
+      }
     }
   }
 
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -112,15 +112,6 @@ div.show-more {
     .search-bar {
       flex: 1 1 auto;
     }
-
-    .search-editor-opener,
-    .continue-with {
-      margin-left: .5em;
-
-      &:last-child {
-        margin-right: -.5em;
-      }
-    }
   }
 
   .pagination-control {
@@ -148,13 +139,17 @@ div.show-more {
     padding-top: .25em;
   }
 
+  .control-button {
+    margin-left: .5em;
+
+    &:last-child {
+      margin-right: -.5em;
+    }
+  }
+
   .sort-control {
     display: flex;
     justify-content: flex-end;
-
-	.control-button {
-      margin-right: 0;
-    }
 
     // Icinga Web 2 override
     label {
@@ -164,7 +159,10 @@ div.show-more {
     // Icinga Web 2 override
     .spinner {
       height: 100%;
-      padding-right: 0;
+    }
+
+    .control-button {
+      margin: 0;
     }
 
     .link-button {
@@ -180,7 +178,6 @@ div.show-more {
 
   .show-more {
     margin-top: .25em;
-    margin-left: 6em;
   }
 }
 
@@ -317,4 +314,3 @@ div.show-more {
 .servicegropup-list > col.object-statistics-graph {
   padding-right: .25em;
 }
-
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -95,6 +95,11 @@ div.show-more {
     margin-bottom: .5em;
   }
 
+  // ToDo: .5em spacing (`margin-bottom`) should be applied to `.collapsible` content?
+  > .collapsible {
+    margin-bottom: 0
+  }
+
   &.overdue,
   &.overdue .tabs li.active a {
     background-color: @gray-lighter;
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -159,10 +159,6 @@ div.show-more {
     margin-top: .25em;
     margin-left: 6em;
   }
-
-  .quick-actions {
-    margin-bottom: .5em;
-  }
 }
 
 .content > h2:first-child,
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -95,8 +95,7 @@ div.show-more {
     margin-bottom: .5em;
   }
 
-  // ToDo: .5em spacing (`margin-bottom`) should be applied to `.collapsible` content?
-  > .collapsible {
+  #host-meta-info {
     margin-bottom: 0
   }
 
@@ -133,6 +132,10 @@ div.show-more {
     border-top: none;
   }
 
+  .list-item .visual {
+    margin-top: 0;
+  }
+
   .list-item .visual,
   .list-item .main {
     padding-bottom: .25em;
@@ -142,7 +145,7 @@ div.show-more {
   .control-button {
     margin-left: .5em;
 
-    &:last-child {
+    &.continue-with {
       margin-right: -.5em;
     }
   }
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -167,16 +167,6 @@ div.show-more {
     .control-button {
       margin: 0;
     }
-
-    .link-button {
-      min-width: 2em;
-      text-align: center;
-      .rounded-corners();
-
-      &:hover {
-        background: @gray-lighter;
-      }
-    }
   }
 
   .show-more {
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -96,7 +96,11 @@ div.show-more {
   }
 
   #host-meta-info {
-    margin-bottom: 0
+    margin-bottom: 0;
+
+    .host-meta-info {
+      margin-bottom: .5em;
+    }
   }
 
   &.overdue,
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -152,6 +152,10 @@ div.show-more {
     display: flex;
     justify-content: flex-end;
 
+	.control-button {
+      margin-right: 0;
+    }
+
     // Icinga Web 2 override
     label {
       margin-right: 0;
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -289,3 +289,9 @@ div.show-more {
     }
   }
 }
+
+.hostgroup-list > col.object-statistics-graph,
+.servicegropup-list > col.object-statistics-graph {
+  padding-right: .25em;
+}
+
--- a/public/css/list/comment-list.less
+++ b/public/css/list/comment-list.less
@@ -47,3 +47,9 @@
     white-space: normal;
   }
 }
+
+.controls .comment-list .list-item {
+  .visual {
+    width: auto;
+  }
+}
--- a/public/css/list/comment-list.less
+++ b/public/css/list/comment-list.less
@@ -24,10 +24,6 @@
 }
 
 .comment-list.minimal .list-item {
-  .visual {
-    width: 2em;
-  }
-
   .user-ball {
     font-size: .857em;
     height: 1.75em;
@@ -45,11 +41,5 @@
   .caption {
     max-height: 4.5em;
     white-space: normal;
-  }
-}
-
-.controls .comment-list .list-item {
-  .visual {
-    width: auto;
   }
 }
--- a/public/css/list/downtime-list.less
+++ b/public/css/list/downtime-list.less
@@ -74,7 +74,7 @@
   }
 }
 
-.downtime-list.minimal .list-item {
+.item-list.downtime-list.minimal .list-item {
   .visual {
     display: block;
     line-height: 1.5;
--- a/public/css/list/downtime-list.less
+++ b/public/css/list/downtime-list.less
@@ -86,17 +86,6 @@
       font-size: 1em;
     }
   }
-
-  .caption {
-    .text-ellipsis()
-  }
-}
-
-.downtime-list.detailed .list-item {
-  .caption {
-    height: auto;
-    max-height: 4.5em;
-  }
 }
 
 .controls .downtime-list.minimal .list-item .visual {
--- a/public/css/list/downtime-list.less
+++ b/public/css/list/downtime-list.less
@@ -62,7 +62,6 @@
     flex-shrink: 0;
     line-height: 1em;
     margin-right: .5em;
-    margin-bottom: 1px;
     padding: .5em .25em;
     text-align: center;
     width: 6em;
@@ -86,8 +85,4 @@
       font-size: 1em;
     }
   }
-}
-
-.controls .downtime-list.minimal .list-item .visual {
-  margin-bottom: 0;
 }
--- a/public/css/list/downtime-list.less
+++ b/public/css/list/downtime-list.less
@@ -57,7 +57,6 @@
     height: ~"calc(100% + 1px)"; // +1px due to the border added exclusively for the first item
   }
 
-
   .visual {
     justify-content: center;
     flex-shrink: 0;
@@ -86,5 +85,22 @@
       display: inline-block;
       font-size: 1em;
     }
+  }
+
+  .caption {
+    .text-ellipsis()
+  }
+}
+
+.downtime-list.detailed .list-item {
+  .caption {
+    height: auto;
+    max-height: 4.5em;
+  }
+}
+
+.controls .downtime-list.minimal .list-item {
+  &:only-child .visual {
+    margin-bottom: 0;
   }
 }
--- a/public/css/list/downtime-list.less
+++ b/public/css/list/downtime-list.less
@@ -99,8 +99,6 @@
   }
 }
 
-.controls .downtime-list.minimal .list-item {
-  &:only-child .visual {
-    margin-bottom: 0;
-  }
+.controls .downtime-list.minimal .list-item .visual {
+  margin-bottom: 0;
 }
--- a/public/css/list/item-list.less
+++ b/public/css/list/item-list.less
@@ -66,7 +66,7 @@
 
     header {
       display: flex;
-      align-items: baseline;
+      align-items: flex-start;
       justify-content: space-between;
     }
 
--- a/public/css/list/item-list.less
+++ b/public/css/list/item-list.less
@@ -167,3 +167,13 @@
 .controls .list-item:not(:only-child):not(:last-child) {
   margin-bottom: .5em;
 }
+
+.controls .item-list:not(.detailed):not(.minimal) .list-item {
+  .plugin-output {
+    line-height: 1.5
+  }
+
+  .caption {
+    height: 2.5em;
+  }
+}
--- a/public/css/list/item-list.less
+++ b/public/css/list/item-list.less
@@ -163,3 +163,7 @@
     .line-clamp(4)
   }
 }
+
+.controls .list-item:not(:only-child):not(:last-child) {
+  margin-bottom: .5em;
+}
--- a/public/css/list/item-list.less
+++ b/public/css/list/item-list.less
@@ -89,6 +89,7 @@
   }
 
   .visual {
+    width: auto;
     margin-right: .5em;
   }
 
@@ -164,7 +165,7 @@
   }
 }
 
-.controls .list-item:not(:only-child):not(:last-child) {
+.controls .list-item:not(:last-child) {
   margin-bottom: .5em;
 }
 
--- a/public/css/list/item-table.less
+++ b/public/css/list/item-table.less
@@ -62,3 +62,27 @@
     width: 1em;
   }
 }
+
+.controls .user-list,
+.controls .usergroup-list {
+  .usergroup-ball,
+  .user-ball {
+    height: 1.5em;
+    width: 1.5em;
+    line-height: ~"calc(1.5em - 4px)";
+    display: block;
+  }
+
+  .title br {
+    display: none;
+  }
+
+  .list-item > .col {
+    padding: .25em 0;
+  }
+}
+
+.hostgroup-list > col.object-statistics-graph,
+.servicegropup-list > col.object-statistics-graph {
+  padding-right: .25em;
+}
--- a/public/css/list/item-table.less
+++ b/public/css/list/item-table.less
@@ -62,9 +62,3 @@
     width: 1em;
   }
 }
-
-
-.hostgroup-list > col.object-statistics-graph,
-.servicegropup-list > col.object-statistics-graph {
-  padding-right: .25em;
-}
--- a/public/css/list/item-table.less
+++ b/public/css/list/item-table.less
@@ -63,24 +63,6 @@
   }
 }
 
-.controls .user-list,
-.controls .usergroup-list {
-  .usergroup-ball,
-  .user-ball {
-    height: 1.5em;
-    width: 1.5em;
-    line-height: ~"calc(1.5em - 4px)";
-    display: block;
-  }
-
-  .title br {
-    display: none;
-  }
-
-  .list-item > .col {
-    padding: .25em 0;
-  }
-}
 
 .hostgroup-list > col.object-statistics-graph,
 .servicegropup-list > col.object-statistics-graph {
--- a/public/css/list/user-list.less
+++ b/public/css/list/user-list.less
@@ -1,0 +1,20 @@
+// Layout
+
+.controls .user-list,
+.controls .usergroup-list {
+  .usergroup-ball,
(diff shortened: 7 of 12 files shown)

### CODE REVIEW

File: public/css/widget/quick-actions.less [hunk 1]
```diff
@@ -2,27 +2,35 @@
   display: flex;
   flex-wrap: wrap;
   list-style-type: none;
-  // TODO: Eh? Minus .25em on the left and right, maybe, because padding and responsiveness.
-  //       But what is the additional 1em on the left for???
-  margin: 0 -1.25em 0 -.25em;
-  margin: 0;
+  margin: 0 -.5em;
```
nilmerg, line 5: What does the decrease by .25em accomplish in this regard? It's horizontal, why is it necessary for the controls alignment?

public/css/widget/quick-actions.less
```diff
@@ -2,27 +2,35 @@
...
+
+  li {
+    margin: 0 .25em .5em .25em;
+    vertical-align: middle;
+    white-space: nowrap;
+  }
+}
+
+
+.controls > .quick-actions:not(:only-child) {
```
nilmerg, line 32: `.controls > .quick-actions:last-child` suffices here.

public/css/list/item-table.less
```diff
@@ -62,3 +62,27 @@
...
+  }
+
+  .title br {
+    display: none;
+  }
+
+  .list-item > .col {
+    padding: .25em 0;
+  }
+}
```
nilmerg, line 83: Doesn't belong here. This has nothing to do with ordinary item tables. Put that into a separate file for user- and usergroup-lists.

public/css/list/item-table.less
```diff
@@ -62,3 +62,27 @@
...
+
+.hostgroup-list > col.object-statistics-graph,
+.servicegropup-list > col.object-statistics-graph {
+  padding-right: .25em;
+}
```
nilmerg, line 88: Also doesn't belong here. Put that into `common.less` where all the other hostgroup- and servicegroup-list rules are.

nilmerg (changes requested):

public/css/list/downtime-list.less
```diff
@@ -68,7 +68,6 @@
     flex-shrink: 0;
     line-height: 1em;
     margin-right: .5em;
-    margin-bottom: 1px;
```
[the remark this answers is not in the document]
flourish86: Fixed with [this](https://github.com/Icinga/icingadb-web/blob/bf6d9bf31ae8568a4a79effcb84baae1daa35b1d/public/css/list/downtime-list.less#L109)

I see no other way to isolate it in the downtime detail header.

public/css/widget/quick-actions.less
flourish86: I'll leave this as discussed.

public/css/widget/quick-actions.less
flourish86: ✅

public/css/list/item-table.less
flourish86: ✅

public/css/list/item-table.less
flourish86: ✅

public/css/list/user-list.less
```diff
@@ -0,0 +1,18 @@
+.controls .user-list,
```
nilmerg, line 1: Please put a `// Layout` comment above this block.

public/css/list/comment-list.less
```diff
@@ -47,3 +47,14 @@
...
   }
 }
+
+.controls .comment-list .list-item {
+  .visual {
+    width: auto;
+
+    .user-ball {
+      // Align optically to `quick-actions`
+      margin-left: -1px;
```
nilmerg, line 57: Which quick actions?? There are none in a comment's detail/details view.

Also, the negative margin feels arbitrary if applied to the user-ball. Apply it to the visual if you absolutely have to.

public/css/list/host-service-list.less
```diff
@@ -0,0 +1,11 @@
+.controls .host-list .list-item,
+.controls .service-list .list-item {
+  .visual {
+    width: auto;
+
+    .state-ball {
+      // Align optically to `quick-actions`
+      margin-left: -1px;
```
nilmerg, line 8: Again, feels arbitrary. Apply this to the visual.

nilmerg (changes requested):
Some impressions of where the lines don't lign up. In widescreen mode, didn't bother to check the other layouts at all yet.

Screenshot from 2021-05-27 11-17-32
Screenshot from 2021-05-27 11-17-21
Screenshot from 2021-05-27 11-16-46
Screenshot from 2021-05-27 11-16-17
Screenshot from 2021-05-27 11-16-11
Screenshot from 2021-05-27 11-16-00
Screenshot from 2021-05-27 11-15-41
Screenshot from 2021-05-27 11-15-29
Screenshot from 2021-05-27 11-14-33
Screenshot from 2021-05-27 10-50-29

public/css/list/item-table.less
```diff
@@ -62,3 +62,4 @@
     width: 1em;
   }
 }
+
```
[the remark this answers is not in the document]
flourish86: ✅

public/css/list/user-list.less
flourish86: ✅

public/css/list/downtime-list.less
```diff
@@ -62,14 +62,13 @@
...
 
-
   .visual {
     justify-content: center;
     flex-shrink: 0;
     line-height: 1em;
     margin-right: .5em;
-    margin-bottom: 1px;
     padding: .5em .25em;
+    margin-bottom: 1px;
```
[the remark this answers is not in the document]
flourish86: You mean above the `padding: .5em .25em;`? Then ✅

public/css/list/host-service-list.less
nilmerg: If it's only optically misaligned, I'd opt for leaving it out. Especially since we're talking about a single pixel. Otherwise it's even more arbitrary.

flourish86: I don't agree on the "only". 😉

But I'm fine with leaving it out.

flourish86: ✅

public/css/list/comment-list.less
flourish86: ✅

public/css/common.less
```diff
@@ -153,16 +157,22 @@ div.show-more {
...
+
+    .link-button {
+      min-width: 2em;
+      text-align: center;
+      .rounded-corners();
+
+      &:hover {
+        background: @gray-lighter;
+      }
+    }
```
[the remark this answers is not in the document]
flourish86: I get that.

Alternatively I would introduce a specific class, e.g. `.sort-direction-toggle`, analogous to `.search-editor-opener` and `.continue-with`.

Or what about having a separate class for those kind of buttons in controls, e.g. `.control-button`, that we could also use for the buttons related to the filter bar?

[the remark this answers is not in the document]
nilmerg: > Alternatively I would introduce a specific class, e.g. .sort-direction-toggle, analogous to .search-editor-opener and .continue-with.

Sounds better.

> Or what about having a separate class for those kind of buttons in controls, e.g. .control-button, that we could also use for the buttons related to the filter bar?

Sounds also good, though then the class would need to come from ipl-web. But only a class without a respective widget as well seems odd to me.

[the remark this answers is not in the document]
flourish86: > Or what about having a separate class for those kind of buttons in controls, e.g. .control-button, that we could also use for the buttons related to the filter bar?

Done this starting from c0cfe3f and [b8d3244](https://github.com/Icinga/ipl-web/pull/21/commits/b8d3244f847288363d66a5bb53d72c99e9782ee2).

application/controllers/CommentsController.php
```diff
@@ -161,11 +161,13 @@ public function detailsAction()
 
         $this->addControl((new CommentList($rs))->setViewMode('minimal'));
 
-        $this->addControl(new ShowMore(
-            $rs,
-            Links::comments()->setQueryString(QueryString::render($this->getFilter())),
-            sprintf(t('Show all %d comments'), $comments->count())
-        ));
+        if ($comments->count() > 3) {
```
nilmerg, line 164: Why do you need this? The `ShowMore` widget does that internally already.

application/controllers/HostController.php
```diff
@@ -88,6 +88,7 @@ public function sourceAction()
         }
 
         $this->addControl((new HostList([$this->host]))
+            ->setViewMode('common')
```
nilmerg, line 91: If you're applying this here, you also have to apply it on the main lists. Which doesn't suit I guess, so please think of another solution. Possibly a CSS only solution.

library/Icingadb/Web/Controller.php
```diff
@@ -319,6 +319,7 @@ public function createContinueWith(Url $detailsUrl, SearchBar $searchBar)
     {
         $continueWith = new ContinueWith($detailsUrl, [$searchBar, 'getFilter']);
         $continueWith->setBaseTarget('_next');
+        $continueWith->addAttributes(['class' => 'control-button']);
```
nilmerg, line 322: Please do that in the widget itself.

public/css/widget/view-mode-switcher.less
```diff
@@ -28,6 +28,11 @@
     &:not(:last-of-type) {
       border-right: 1px solid @low-sat-blue-dark;
     }
+
+    i {
+      // fix height for Chrome
+      display: block;
```
nilmerg, line 34: That's not really valid. The parent is an inline element.

public/css/widget/host-meta-info.less
```diff
@@ -64,7 +64,11 @@
 }
 
 .controls > .host-meta-info {
-  margin-top: -.5em;
+  padding: .5em 0;
+}
+
+.controls > #host-meta-info.collapsed {
```
nilmerg, line 70: There is already a suitable rule above, why not use it? (Instead of relying on an id which isn't for CSS rules (if there's a class) to begin with.)

public/css/common.less
```diff
@@ -111,6 +111,10 @@ div.show-more {
     .search-editor-opener,
     .continue-with {
       margin-left: .5em;
+
+      &:last-child {
+        margin-right: -.5em;
```
nilmerg, line 116: Isn't that the same as https://github.com/Icinga/ipl-web/pull/21/files#diff-89766a82815789bc3a264cb00929ae46f9cc699215193e13e3763ab8eb69ea54R59 ?

public/css/common.less
```diff
@@ -147,22 +151,12 @@ div.show-more {
     label {
       margin-right: 0;
     }
-
-    // Icinga Web 2 override
-    .spinner {
-      height: 100%;
-      padding-right: 0;
-    }
```
nilmerg, line 155: I keep seeing this (but can't find the other occurrence now), but you don't re-add it somewhere else. What makes you think this isn't required anymore?

public/css/list/comment-list.less
```diff
@@ -47,3 +47,9 @@
     white-space: normal;
   }
 }
+
+.controls .comment-list .list-item {
+  .visual {
+    width: auto;
```
nilmerg, line 53: Can't we make this the default for minimal list items now? It's now added in three places, comment, host and service list items. The width of the minimal list items already differs in lists now (compared to common and detailed) and I don't see a reason why we should it make the default. The narrower visual on all minimal list items looks okay to me.

public/css/list/item-list.less
```diff
@@ -170,3 +170,17 @@
     position: relative;
   }
 }
+
+.controls .list-item:not(:only-child):not(:last-child) {
```
nilmerg, line 174: You can remove the `:not(:only-child)` selector. If it's the only child, it's also the last.

nilmerg (changes requested):

application/controllers/CommentsController.php
flourish86: Without this, an empty show-more container will be rendered, which takes up unecessary space. 

### Current Icinga DB Web Master

Screen Shot 2021-06-02 at 09 57 08

public/css/widget/view-mode-switcher.less
flourish86: Block elements in inline elements aren't valid in HTML for structural reasons, afaik. But I don't see why this shouldn't be valid done with CSS.

Besides that, `inline` and `inline-block` result in crazily inconsistent and incomprehensible value rounding of the rendered height across browsers.

In the end `display: block` is the only thing I found, that works.

application/controllers/CommentsController.php
nilmerg: Then please play with `:empty` or so.

nilmerg: Or override `renderUnwrapped()` in the `ShowMore` widget and perform the same check as in `assemble` to return an empty string if successful.

flourish86: ✅

library/Icingadb/Web/Controller.php
flourish86: ✅

application/controllers/HostController.php
flourish86: ✅

public/css/widget/host-meta-info.less
flourish86: What do you think of 186fd62a.

Also associated af4cf52c.

(Please take `@ToDo` comments into account)

public/css/widget/view-mode-switcher.less
nilmerg: nvm, seems to be valid inside inline **-block**.

public/css/list/comment-list.less
flourish86: Fine with that. See 03a2315f / f88dc369

✅

public/css/common.less
flourish86: I put it back.

flourish86: ✅

public/css/list/item-list.less
flourish86: Smart. 💡

flourish86: ✅

public/css/common.less
flourish86: I think the link target doesn't consider the changes that happened since then.

But how do you like this, then: 0c912205

public/css/common.less
```diff
@@ -95,23 +95,23 @@ div.show-more {
     margin-bottom: .5em;
   }
 
+  // ToDo: .5em spacing (`margin-bottom`) should be applied to `.collapsible` content?
+  > .collapsible {
+    margin-bottom: 0
+  }
```
nilmerg, line 101: Too generic for me. The only reason why this is needed is due to the host meta info widget. This widget already got the bottom margin, so it should be removed in the same place due to the same circumstances. (Then possibly by targeting the id again, my previous comment against the id was flawed)

public/css/common.less
```diff
@@ -139,6 +139,14 @@ div.show-more {
     padding-top: .25em;
   }
 
+  .control-button {
+    margin-left: .5em;
+
+    &:last-child {
```
nilmerg, line 145: Please target the `.continue-with` here only. This then would also save us the `margin: 0;` for the sort control.

Besides, another reason why I don't like this is that `:last-child` is not strictly correct here. This only describes the last child in whatever container the control button is a member of. This works currently because the `.filter` container uses the entire width of the column. If it (or any other container with a control button) doesn't require the entire width, this may lead to unexpected results. (As shown with the sort control) By targeting the continue with widget only, we achieve the same result and make it less ambiguous.

public/css/common.less
```diff
@@ -151,17 +159,25 @@ div.show-more {
...
     .spinner {
       height: 100%;
-      padding-right: 0;
+    }
+
+    .control-button {
+      margin: 0;
+    }
+
+    .link-button {
```
nilmerg, line 168: Isn't that the `.control-button` now? Or is this used somewhere else?

public/css/common.less
```diff
@@ -151,17 +159,25 @@ div.show-more {
...
+
+      &:hover {
+        background: @gray-lighter;
+      }
     }
   }
 
   .show-more {
     margin-top: .25em;
-    margin-left: 6em;
```
nilmerg, line 160: This leads to less differentiation between the quick actions and the show all link:
Screenshot from 2021-06-09 17-39-07

Some gap is of advantage here IMHO. It also looks weird because below there's empty space anyway.

nilmerg (changes requested):
I also have in some circumstances a difference of a single pixel it seems: Screenshot from 2021-06-09 17-43-50

public/css/list/downtime-list.less
```diff
@@ -87,4 +86,19 @@
...
+    .text-ellipsis()
+  }
+}
+
+.downtime-list.detailed .list-item {
+  .caption {
+    height: auto;
+    max-height: 4.5em;
+  }
+}
```
nilmerg, line 100: These changes are obsolete since #202

library/Icingadb/Widget/ShowMore.php
```diff
@@ -40,10 +40,50 @@ public function getLabel()
...
-            $this->add(new ActionLink($this->getLabel(), $this->url));
+            return sprintf(
+                '<%s%s>%s</%s>',
+                $tag,
+                $attributes,
+                $content,
+                $tag
+            );
+        } else {
+            return '';
```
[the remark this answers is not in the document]
flourish86: So, this solves [this](https://github.com/Icinga/icingadb-web/pull/206#discussion_r648316520), right?

public/css/widget/host-meta-info.less
```diff
@@ -59,14 +59,20 @@
   .vertical-key-value .value,
   .vertical-key-value .key {
     font-size: 16/18em;
-    line-height: 1.5;
+    line-height: 18/16; // compensate smaller font-size (1/font-size)
   }
 }
 
+
+// @ToDo: Isn't this obsolete, since `.host-meta-info` is wrapped in `#host-meta-info`
```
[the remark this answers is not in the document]
flourish86: ✅

public/css/common.less
flourish86: Can you check, what version you checked out? On the screenshot it doesn't look like it was intended. This is what the controls are supposed to look.

Screen Shot 2021-06-30 at 14 57 59

flourish86: This way it's also tidier and clearer, which elements belong where. Heights match the row grid and I'd say it's the same height in 80% of cases as the right floating "Show More" link.

public/css/common.less
flourish86: ✅

library/Icingadb/Widget/ShowMore.php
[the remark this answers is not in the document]
nilmerg: No, there are more circumstances :wink:

public/css/common.less
nilmerg: I'm on de6e6d65 and it's still the case. Looking at your screen, please remember the wide screen layout. I don't want a useless break there on my wide screen :wink: 
Screenshot from 2021-07-01 09-48-49

public/css/list/downtime-list.less
flourish86: ✅

public/css/common.less
flourish86: You're right. Dropped that block completely, since it's not used anywhere else.

flourish86: ✅

public/css/list/downtime-list.less
```diff
@@ -105,3 +104,6 @@
   }
 }
 
+.controls .downtime-list.minimal .list-item .visual {
+  margin-bottom: 0;
```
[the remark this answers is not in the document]
flourish86: As discussed. ✅

public/css/common.less
flourish86: As discussed, I used the id.

flourish86: ✅

public/css/common.less
flourish86: Smart. I'll make specific style for `.wide-layout`.

library/Icingadb/Widget/ShowMore.php
```diff
@@ -40,10 +40,50 @@ public function getLabel()
...
         }
     }
+
+    protected function assemble()
+    {
+        $this->add(new ActionLink($this->getLabel(), $this->url));
```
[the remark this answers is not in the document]
flourish86: Readded it.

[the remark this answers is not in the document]
flourish86: ✅

library/Icingadb/Widget/ShowMore.php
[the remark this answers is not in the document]
flourish86: ✅

public/css/common.less
nilmerg: In non-wide layout it still looks this way:
Screenshot from 2021-07-13 12-31-28

public/css/widget/quick-actions.less
```diff
@@ -2,31 +2,44 @@
...
+    .rounded-corners();
+    display: inline-block;
 
     &:hover {
       background: @gray-lighter;
     }
   }
+
+  li {
+    margin: 0 .25em .5em .25em;
```
nilmerg, line 25: This leads to a strange alignment for me: (notice that the icons in default layout are indented differently than in wide layout)

### default layout
Screenshot from 2021-07-13 12-40-25

### wide layout
Screenshot from 2021-07-13 12-40-41

nilmerg (changes requested):

public/css/widget/quick-actions.less
flourish86: This is interesting. Because in Safari this looks completely as intended (see screenshots). I'll have a look at the other browsers.

flourish86: Should be fixed with  8f763c7

✅

public/css/widget/host-meta-info.less
```diff
@@ -59,12 +59,12 @@
...
     font-size: 16/18em;
-    line-height: 1.5;
+    line-height: 18/16; // compensate smaller font-size (1/font-size)
   }
 }
 
-.controls > .host-meta-info {
-  margin-top: -.5em;
+.controls .host-meta-info {
+  margin-bottom: .5em;
```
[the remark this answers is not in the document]
flourish86: ✅

public/css/common.less
```diff
@@ -95,23 +95,26 @@ div.show-more {
     margin-bottom: .5em;
   }
 
+  #host-meta-info {
+    margin-bottom: 0;
+  }
+
+  .host-meta-info {
+    margin-bottom: .5em;
+  }
```
nilmerg, line 104: Please put it inside, so that the bottom margin is only applied if it's inside the `#host-meta-info` div.

```suggestion
  #host-meta-info {
    margin-bottom: 0;

    .host-meta-info {
      margin-bottom: .5em;
    }
  }
```

nilmerg (changes requested):

public/css/common.less
flourish86: ✅

nilmerg (approved):
Looks fine now. At least code wise. Will check UI related things next week to be 100% sure. :wink:

PR #207, opened 2021-05-27
Bugfix/icons for multiselect detail

flourish86:


nilmerg:
Please use master as base for this branch, otherwise I can't review this without having to look at all the changes from #206 again.

flourish86:
> Please use master as base for this branch, otherwise I can't review this without having to look at all the changes from #206 again.

Sorry, my fault. fixed it and also the CodeSniffer error.

library/Icingadb/Widget/Detail/MultiselectQuickActions.php
```diff
@@ -158,7 +159,7 @@ protected function assemble()
             $this->assembleAction(
                 'processCheckresult',
                 t('Process check result'),
-                'icon-edit',
+                'random',
```
nilmerg, line 162: Isn't that the icon that's also being used for indicating a flapping host/service? (If not exactly this icon, it's at least a very similar one)

Personally, I'd leave this the same as before, but change the icon for the acknowledge action. (Maybe the check icon or so)

flourish86: Sounds good. I'll change it.

flourish86: ✅

nilmerg (approved):

Issue #144, opened 2020-10-21
Add tool tips for controls


flourish86:
## Describe the bug
It's hard to guess what not labeled control elements will do.

## Expected behavior
Unlabeled control elements should have a title attribute, so that there's an explanation of what they will do.

View Switcher, something like
* Show compact list items
* Show default list item information
* Show detailed list item information

Sort direction button:
* Sort list items in an ascending order
* Sort list items in an descending order

Since the other list item control elements are labeled, title attributes might only be necessary for accessibility reasons.

nilmerg:
Thanks Ravi!

PR #211, opened 2021-06-14
Add tool tip to ViewModeSwitcher.

raviks789:
Tool tip to ViewModeSwitcher has been added here.

#144

### DIFF
--- a/library/Icingadb/Widget/ViewModeSwitcher.php
+++ b/library/Icingadb/Widget/ViewModeSwitcher.php
@@ -152,8 +152,25 @@ class ViewModeSwitcher extends Form
                 new IcingaIcon($icon)
             );
 
+            switch ($viewMode) {
+                case 'minimal':
+                    $active = t('Minimal view active');
+                    $inactive = t('Switch to minimal view');
+                    break;
+                case 'common':
+                    $active = t('Common view active');
+                    $inactive = t('Switch to common view');
+                    break;
+                case 'detailed':
+                    $active = t('Detailed view active');
+                    $inactive = t('Switch to detailed view');
+            }
+
             if ($viewMode === $currentViewMode) {
                 $input->getAttributes()->add('checked', true);
+                $label->getAttributes()->add('title', $active);
+            } else {
+                $label->getAttributes()->add('title', $inactive);
             }
 
             $this->add([$input, $label]);


library/Icingadb/Widget/ViewModeSwitcher.php
```diff
@@ -154,6 +154,9 @@ protected function assemble()
 
             if ($viewMode === $currentViewMode) {
                 $input->getAttributes()->add('checked', true);
+                $label->getAttributes()->add('title', sprintf('%s view active', $viewMode));
```
flourish86, line 157: Please start tool tip sentence with Uppercase letter.

flourish86 (changes requested):

library/Icingadb/Widget/ViewModeSwitcher.php
```diff
@@ -154,6 +154,9 @@ protected function assemble()
 
             if ($viewMode === $currentViewMode) {
                 $input->getAttributes()->add('checked', true);
+                $label->getAttributes()->add('title', t(sprintf('%s view active', ucfirst($viewMode))));
```
nilmerg, line 157: Three issues regarding localisation:

a) gettext will parse the source code to extract messages, so translatable strings need to be directly wrapped by the translation function (i.e. `t('%s view active')`)
b) Passing a computed result to any translation function isn't possible. gettext extracts the arguments for such translation functions as `t()` and hence computed values won't be.. computed :smile: 
c) Translatable messages may of course contain placeholders for sprintf, but then it's usually a value that is localized differently (such as a date) or not localized at all (such as a number). This however is a name, and names are localizable, they're adjectives after all. So here we'd need three different messages, one for each mode.

library/Icingadb/Widget/ViewModeSwitcher.php
```diff
@@ -154,6 +154,9 @@ protected function assemble()
...
+            } else {
+                $label->getAttributes()->add('title', t(sprintf('Switch to %s view', $viewMode)));
```
nilmerg, line 159: See above.

nilmerg (changes requested):

flourish86 (approved):

library/Icingadb/Widget/ViewModeSwitcher.php
nilmerg: This is resolved? You've only applied a) and b), what about c)?

nilmerg (approved):

PR #218, opened 2021-06-23
Feature: Increase limit on minimal view mode

### LINKED ISSUE #100: ViewSwitcher: Increase limit on minimal view mode
List items in minimal need less vertical space by a lot in comparison to the others view modes. When switching a list view to minimal mode the additional up space is unused.

Increase (maybe double) the limit when switching to minimal view.

Screen Shot 2020-05-27 at 15 59 39

Yonas-net:
resolves #100

### DIFF
--- a/library/Icingadb/Web/Controller.php
+++ b/library/Icingadb/Web/Controller.php
@@ -338,9 +338,9 @@ class Controller extends CompatController
      *
      * @return ViewModeSwitcher
      */
-    public function createViewModeSwitcher()
+    public function createViewModeSwitcher(PaginationControl $paginationControl, LimitControl $limitControl)
     {
-        $viewModeSwitcher = new ViewModeSwitcher(Url::fromRequest());
+        $viewModeSwitcher = new ViewModeSwitcher();
         $viewModeSwitcher->setIdProtector([$this->getRequest(), 'protectId']);
 
         $prefs = $this->Auth()->getUser()->getPreferences();
@@ -349,19 +349,78 @@ class Controller extends CompatController
             $viewModeSwitcher->setDefaultViewMode($viewMode);
         }
 
-        $viewModeSwitcher->populate(['view' => $this->params->shift($viewModeSwitcher->getViewModeParam())]);
+        $viewModeSwitcher->populate([
+            'view' => $this->params->shift($viewModeSwitcher->getViewModeParam())
+        ]);
 
-        $viewModeSwitcher->on(ViewModeSwitcher::ON_SUCCESS, function (ViewModeSwitcher $viewModeSwitcher) use ($prefs) {
-            $viewMode = $viewModeSwitcher->getValue($viewModeSwitcher->getViewModeParam());
+        $session = $this->Window()->getSessionNamespace(
+            'icingadb-viewmode-' . $this->Window()->getContainerId()
+        );
 
-            $icingadbPrefs = $prefs->icingadb ?: [];
-            $icingadbPrefs['view_mode'] = $viewMode;
-            $prefs->icingadb = $icingadbPrefs;
+        $viewModeSwitcher->on(
+            ViewModeSwitcher::ON_SUCCESS,
+            function (ViewModeSwitcher $viewModeSwitcher) use ($prefs, $paginationControl, &$session) {
+                $viewMode = $viewModeSwitcher->getValue($viewModeSwitcher->getViewModeParam());
+                $icingadbPrefs = $prefs->icingadb ?: [];
+                $icingadbPrefs['view_mode'] = $viewMode;
+                $prefs->icingadb = $icingadbPrefs;
 
-            $this->redirectNow(Url::fromRequest()->setParam($viewModeSwitcher->getViewModeParam(), $viewMode));
-        });
+                $currentPage = $paginationControl->getCurrentPageNumber();
 
-        $viewModeSwitcher->handleRequest(ServerRequest::fromGlobals());
+                $requestUrl = Url::fromRequest();
+                $pageParam = $paginationControl->getPageParam();
+                $limitParam = LimitControl::DEFAULT_LIMIT_PARAM;
+                $urlParams[$viewModeSwitcher->getViewModeParam()] = $viewMode;
+
+                if (! $requestUrl->hasParam($limitParam)) {
+                    if ($viewMode === 'minimal') {
+                        $session->set('previous_page', $currentPage);
+                        $session->set('request_path', $requestUrl->getPath());
+
+                        $limit = $paginationControl->getLimit();
+                        $currentPage = (int) (floor((($currentPage * $limit) - $limit) / ($limit * 2)) + 1);
+
+                        $session->set('current_page', $currentPage);
+                    } elseif ($viewModeSwitcher->getDefaultViewMode() === 'minimal') {
+                        if ($currentPage === $session->get('current_page')) {
+                            // No other page numbers have been selected, i.e the user only
+                            // switches back and forth without changing the page numbers
+                            $currentPage =  $session->get('previous_page');
+                        } else {
+                            $limit = $paginationControl->getLimit();
+                            $currentPage = (int) (floor((($currentPage * $limit) - $limit) / ($limit / 2)) + 1);
+                        }
+
+                        $session->clear();
+                    }
+
+                    if (($requestUrl->hasParam($pageParam) && $currentPage > 1) || $currentPage > 1) {
+                        $urlParams[$pageParam] = $currentPage;
+                    }
+                } else {
+                    $urlParams[$limitParam] = $requestUrl->getParam($limitParam);
+                }
+
+                $this->redirectNow($requestUrl->setParams($urlParams));
+            }
+        )->handleRequest(ServerRequest::fromGlobals());
+
+        if ($viewModeSwitcher->getViewMode() === 'minimal') {
+            $hasLimitParam = Url::fromRequest()->hasParam($limitControl->getLimitParam());
+
+            if ($paginationControl->getDefaultPageSize() <= LimitControl::DEFAULT_LIMIT && ! $hasLimitParam) {
+                $paginationControl->setDefaultPageSize($paginationControl->getDefaultPageSize() * 2);
+                $limitControl->setDefaultLimit($limitControl->getDefaultLimit() * 2);
+
+                $paginationControl->apply();
+            }
+        }
+
+        $requestPath =  $session->get('request_path');
+
+        if ($requestPath && $requestPath !== Url::fromRequest()->getPath()) {
+            $session->clear();
+        }
 
         return $viewModeSwitcher;
     }
--- a/application/controllers/CommentsController.php
+++ b/application/controllers/CommentsController.php
@@ -50,7 +50,7 @@ class CommentsController extends Controller
                 'comment.expire_time desc'                => t('Expire Time')
             ]
         );
-        $viewModeSwitcher = $this->createViewModeSwitcher();
+        $viewModeSwitcher = $this->createViewModeSwitcher($paginationControl, $limitControl);
         $searchBar = $this->createSearchBar($comments, [
             $limitControl->getLimitParam(),
             $sortControl->getSortParam(),
--- a/application/controllers/DowntimesController.php
+++ b/application/controllers/DowntimesController.php
@@ -55,7 +55,7 @@ class DowntimesController extends Controller
                 'downtime.duration desc'                          => t('Duration')
             ]
         );
-        $viewModeSwitcher = $this->createViewModeSwitcher();
+        $viewModeSwitcher = $this->createViewModeSwitcher($paginationControl, $limitControl);
         $searchBar = $this->createSearchBar($downtimes, [
             $limitControl->getLimitParam(),
             $sortControl->getSortParam(),
--- a/application/controllers/HostgroupController.php
+++ b/application/controllers/HostgroupController.php
@@ -52,7 +52,7 @@ class HostgroupController extends Controller
 
         $limitControl = $this->createLimitControl();
         $paginationControl = $this->createPaginationControl($hosts);
-        $viewModeSwitcher = $this->createViewModeSwitcher();
+        $viewModeSwitcher = $this->createViewModeSwitcher($paginationControl, $limitControl);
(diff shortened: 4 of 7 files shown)


Yonas-net:
Please Help!

I don't know exactly how the indentation for multi-line if conditions are. No matter how I do it, the CI/CD fall through.

nilmerg:
That's described in the [PSR12 coding style](https://www.php-fig.org/psr/psr-12/):

```php
<?php

if (
    $expr1
    && $expr2
) {
    // if body
} elseif (
    $expr3
    && $expr4
) {
    // elseif body
}
```

And regarding your current implementation, you should process everything upon the view mode switcher is submitted. That's not the case in the methods `createLimitControl()` or `createPagination()` but in method `createViewModeSwitcher()` as part of the success handler there. It should not be required to save anything to the preferences to apply a new limit or page this way.

library/Icingadb/Web/Controller.php
```diff
@@ -349,19 +349,47 @@ public function createViewModeSwitcher()
...
+                $icingadbPrefs = $prefs->icingadb ?: [];
+                $icingadbPrefs['view_mode'] = $viewMode;
+                $prefs->icingadb = $icingadbPrefs;
+                $currentPage = $paginationControl->getCurrentPageNumber();
+
+                if ($viewMode === 'minimal') {
+                    $limit = $paginationControl->getLimit();
+                    $currentPage = ((($paginationControl->getCurrentPageNumber() * $limit) - $limit) / ($limit * 2) + 1);
+
+                    $paginationControl->setDefaultPageSize($paginationControl->getDefaultPageSize() * 2);
```
nilmerg, line 377: This and the same call below isn't necessary, remove it.

library/Icingadb/Web/Controller.php
```diff
@@ -349,19 +349,47 @@ public function createViewModeSwitcher()
...
+            function (ViewModeSwitcher $viewModeSwitcher) use ($prefs, &$paginationControl) {
+                $viewMode = $viewModeSwitcher->getValue($viewModeSwitcher->getViewModeParam());
+                $icingadbPrefs = $prefs->icingadb ?: [];
+                $icingadbPrefs['view_mode'] = $viewMode;
+                $prefs->icingadb = $icingadbPrefs;
+                $currentPage = $paginationControl->getCurrentPageNumber();
+
+                if ($viewMode === 'minimal') {
+                    $limit = $paginationControl->getLimit();
+                    $currentPage = ((($paginationControl->getCurrentPageNumber() * $limit) - $limit) / ($limit * 2) + 1);
```
nilmerg, line 375: Please use `$currentPage` here and below, instead of calling `$paginationControl->getCurrentPageNumber()` twice.

library/Icingadb/Web/Controller.php
```diff
@@ -349,19 +349,47 @@ public function createViewModeSwitcher()
...
+                } elseif ($viewModeSwitcher->getDefaultViewMode() === 'minimal') {
+                    $limit = $paginationControl->getLimit();
+                    $currentPage = ((($paginationControl->getCurrentPageNumber() * $limit) - $limit) / ($limit / 2) + 1);
+
+                    $paginationControl->setDefaultPageSize($paginationControl->getDefaultPageSize() / 2);
+                }
+
+                $this->redirectNow(Url::fromRequest()->setParams([
+                    $viewModeSwitcher->getViewModeParam()   => $viewMode,
+                    $paginationControl->getPageParam()      => $currentPage
```
nilmerg, line 387: Please don't add the `page` param if it's value is `1`. That's the default anyway and I'd like to avoid adding the `page` param although it has not been there previously.

nilmerg (changes requested):

nilmerg:
And I forgot to mention that if I choose `100` as limit manually while in the common view mode, switching to the minimal view mode resets it to `50`. The other way round as well. This must not happen.

nilmerg (changes requested):
* Switching from `?view=common&page=2` to minimal view results in `?view=minimal&page=1.5`
* Visiting `?view=minimal` directly (by not using the switcher) doesn't apply the increased limit

library/Icingadb/Web/Controller.php
```diff
@@ -71,6 +71,14 @@ public function createLimitControl()
...
         $limitControl->setDefaultLimit($this->getPageSize(null));
 
+        if (Url::fromRequest()->getParam(ViewModeSwitcher::DEFAULT_VIEW_MODE_PARAM) === 'minimal') {
+            $hasLimitParam = Url::fromRequest()->hasParam($limitControl->getLimitParam());
+
+            if ($limitControl->getDefaultLimit() <= LimitControl::DEFAULT_LIMIT && ! $hasLimitParam) {
+                $limitControl->setDefaultLimit($limitControl->getDefaultLimit() * 2);
+            }
+        }
+
```
nilmerg, line 81: Not sure why you moved this into `createLimitControl`. But in any case, to me this is the incorrect place. It's the view mode switcher why we'd change the default limit. So it's `createViewModeSwitcher`'s duty to do that.

Think of a view that has no view mode switcher but a limit control. Why should the limit be doubled just because there's a `view=minimal` parameter in the url?

library/Icingadb/Web/Controller.php
```diff
@@ -89,6 +97,39 @@ public function createPaginationControl(Paginatable $paginatable)
         $paginationControl->setDefaultPageSize($this->getPageSize(null));
         $paginationControl->setAttribute('id', $this->getRequest()->protectId('pagination-control'));
 
+        if (Url::fromRequest()->getParam(ViewModeSwitcher::DEFAULT_VIEW_MODE_PARAM) === 'minimal') {
+            $hasLimitParam = Url::fromRequest()->hasParam(LimitControl::DEFAULT_LIMIT_PARAM);
+
+            if ($paginationControl->getDefaultPageSize() <= LimitControl::DEFAULT_LIMIT && ! $hasLimitParam) {
+                $paginationControl->setDefaultPageSize($paginationControl->getDefaultPageSize() * 2);
+            }
+        }
```
nilmerg, line 106: Same as above in `createLimitControl`. Why should the page size be doubled just because there's a `view=minimal` parameter in the url?

library/Icingadb/Web/Controller.php
```diff
@@ -89,6 +97,39 @@ public function createPaginationControl(Paginatable $paginatable)
...
+
+        $session = $this->Window()->getSessionNamespace(
+            $this->Window()->getId() . Url::fromRequest()->getPath()
```
nilmerg, line 109: You should use `getContainerId()` here rather than just `getId()`. The latter is already baked into the session namespace identifier, but is also not of much interest because it's the container id which is the unique identifier we should rely on.

Oh and please add another hardcoded prefix into the mix, `3lkj2w3l4322l/icingaweb2/icingadb/hosts` is still too generic. What if someone else thinks of combining the container id with the url path? So please add a hardcoded name as well to make it less generic.

library/Icingadb/Web/Controller.php
```diff
@@ -89,6 +97,39 @@ public function createPaginationControl(Paginatable $paginatable)
...
+                if (
+                    $paginationControl->getCurrentPageNumber() !== $currentPage
+                    && ! Url::fromRequest()->getParams()->isEmpty()
+                ) {
+                    $this->redirectNow(Url::fromRequest()->setParams([
+                        ViewModeSwitcher::DEFAULT_VIEW_MODE_PARAM => $viewMode,
+                        $paginationControl->getPageParam()        => $currentPage
+                    ]));
+                }
+            }
```
nilmerg, line 130: You don't need to re-calculate the position here. If a user switches in minimal mode to another page, the previous page in common mode is invalidated.

If I'm on page 4 in common mode and switch to minimal mode, I'm on page 2. If I switch back to the common/detailed mode, I want to be on page 4 again, since that's what I've seen moments ago. But if I switch in minimal mode to page 4 and then back to common mode, I'd be surprised if I'm on page 7 after that since the list starts at another entry. I want to be on page 6 then.

So if a user changes the page in minimal mode, the previous page need to be cleared. But that I'd still want to see in `createViewModeSwitcher`. For this you just have to include the resulting page number in minimal mode (4 -> 2) in your cached data. If the page number in minimal mode changes (e.g. to 4) `4 -> 4` doesn't match.

nilmerg (changes requested):

library/Icingadb/Web/Controller.php
Yonas-net: > Visiting ?view=minimal directly (by not using the switcher) doesn't apply the increased limit

This is why it is necessary here.

library/Icingadb/Web/Controller.php
Yonas-net: Same.

library/Icingadb/Web/Controller.php
Yonas-net: You're absolutely right, but those use cases you just mentioned are being computed in ``createViewModeSwitcher``. Well, that's rather for another use case, namely: let's say the user is on ``?view=common&page=4`` and switches to minimal view mode and ends up on the page ``?view=minimal&page=2``. What if the user now switches to a completely different page, e.g. from ``/icingadb/hosts -> /icingadb/services``? If he switches back to this page without touching the view mode switcher, I would expect that he ends up on the page where he was before ``/icingadb/hosts?view=minimal&page=2``.

Yonas-net: However, the condition is wrong. This is the correct variant.

```php
if (
         $paginationControl->getCurrentPageNumber() !== $currentPage
         && Url::fromRequest()->getParams()->isEmpty()
     ) {...}
```

library/Icingadb/Web/Controller.php
Yonas-net: ✅

library/Icingadb/Web/Controller.php
nilmerg: `createViewModeSwitcher` also runs always. (It doesn't need to be in the success handler.) And has access to the limit control.

library/Icingadb/Web/Controller.php
nilmerg: > If he switches back to this page without touching the view mode switcher, I would expect that he ends up on the page where he was before /icingadb/hosts?view=minimal&page=2.

No. Leaving the view must clear the page setting. If one wants the previous view, there's the navigation buttons to go *back*.

library/Icingadb/Web/Controller.php
Yonas-net: ✅

library/Icingadb/Web/Controller.php
Yonas-net: ✅

library/Icingadb/Web/Controller.php
Yonas-net: ✅

library/Icingadb/Web/Controller.php
```diff
@@ -349,19 +360,82 @@ public function createViewModeSwitcher()
...
-        $viewModeSwitcher->populate(['view' => $this->params->shift($viewModeSwitcher->getViewModeParam())]);
+        $viewModeSwitcher->populate([
+            'view' => $this->params->shift($viewModeSwitcher->getViewModeParam())
+        ]);
 
-        $viewModeSwitcher->on(ViewModeSwitcher::ON_SUCCESS, function (ViewModeSwitcher $viewModeSwitcher) use ($prefs) {
-            $viewMode = $viewModeSwitcher->getValue($viewModeSwitcher->getViewModeParam());
+        if (
+            $viewModeSwitcher->getViewMode() === 'minimal'
+            || Url::fromRequest()->getParam(ViewModeSwitcher::DEFAULT_VIEW_MODE_PARAM) === 'minimal'
```
nilmerg, line 369: Above, the call to populate already populates the request's view mode so that `getViewMode()` *can* return it.

So if it does, the second condition is redundant and won't ever be evaluated. (Because if `minimal` is indeed the request's view mode, it's also the switcher's current view mode.)

Yes, it currently does not return the request's view mode, because the view mode switcher isn't assembled yet. `getViewMode()` needs to call `ensureAssembled()` internally before the call to `getValue()`. This way it does return the request's view mode. This should play fine with other usages, as every other call to `getViewMode()` already does so when the view mode switcher has been assembled. (`handleRequest()` assembles)

library/Icingadb/Web/Controller.php
```diff
@@ -349,19 +360,82 @@ public function createViewModeSwitcher()
...
+                $limitControl->setDefaultLimit($limitControl->getDefaultLimit() * 2);
+            }
 
-            $this->redirectNow(Url::fromRequest()->setParam($viewModeSwitcher->getViewModeParam(), $viewMode));
-        });
+            if ($paginationControl->getDefaultPageSize() <= LimitControl::DEFAULT_LIMIT && ! $hasLimitParam) {
+                $paginationControl->setDefaultPageSize($paginationControl->getDefaultPageSize() * 2);
+
+                $paginationControl->apply();
+            }
```
nilmerg, line 381: If the paginator uses a limit of `50`, the limiter does the same. So one of these checks should suffice. I'd say the paginator check should be kept, as it's the pagination what this is all about. So please merge the `if` above with the one below but drop its condition.

library/Icingadb/Web/Controller.php
```diff
@@ -349,19 +360,82 @@ public function createViewModeSwitcher()
...
+        }
+
+        $session = $this->Window()->getSessionNamespace(
+            self::SESSION_PREFIX . $this->Window()->getContainerId() . Url::fromRequest()->getPath()
+        );
 
-        $viewModeSwitcher->handleRequest(ServerRequest::fromGlobals());
+        // Clear session when the view mode param doesn't exists
+        if (! Url::fromRequest()->hasParam(ViewModeSwitcher::DEFAULT_VIEW_MODE_PARAM)) {
```
nilmerg, line 389: We're in `createViewmodeSwitcher()` here and should be able to know exactly what the request's view mode is. And that's when we shift the view mode value to pass it to `populate()` above. Let's use the result here as well, shall we?

nilmerg (changes requested):

nilmerg (approved):
Looks and works fine! Thank you!

PR #223, opened 2021-07-07
Optimize downtime list item titles

### LINKED ISSUE #213: Unecessary wide text spacing in list item titles
## Describe the bug

Screen Shot 2021-06-15 at 15 11 09

Screen Shot 2021-06-15 at 15 11 58

flourish86:
'for' is the most natural sounding according to Andrew.

fixes #213

### DIFF
--- a/library/Icingadb/Widget/ItemList/BaseDowntimeListItem.php
+++ b/library/Icingadb/Widget/ItemList/BaseDowntimeListItem.php
@@ -19,6 +19,7 @@ use ipl\Html\Attributes;
 use ipl\Html\BaseHtmlElement;
 use ipl\Html\Html;
 use ipl\Html\HtmlElement;
+use ipl\Html\TemplateString;
 use ipl\Html\Text;
 use ipl\Stdlib\Filter;
 use ipl\Web\Widget\Icon;
@@ -131,23 +132,40 @@ abstract class BaseDowntimeListItem extends BaseListItem
             $link = $this->createServiceLink($this->item->service, $this->item->service->host, true);
         }
 
-        if ($this->getNoSubjectLink()) {
-            $title->addHtml(new HtmlElement(
-                'span',
-                Attributes::create(['class' => 'subject']),
-                Text::create($this->item->is_flexible ? t('Flexible Downtime') : t('Fixed Downtime'))
-            ));
+        if ($this->item->is_flexible) {
+            if ($link !== null) {
+                $template = t('{{#link}}Flexible Downtime{{/link}} for %s');
+            } else {
+                $template = t('Flexible Downtime');
+            }
         } else {
-            $title->addHtml(new Link(
-                $this->item->is_flexible
-                    ? t('Flexible Downtime')
-                    : t('Fixed Downtime'),
-                Links::downtime($this->item)
-            ));
+            if ($link !== null) {
+                $template = t('{{#link}}Fixed Downtime{{/link}} for %s');
+            } else {
+                $template = t('Fixed Downtime');
+            }
         }
 
-        if ($link !== null) {
-            $title->addHtml(Text::create('for '), $link);
+        if ($this->getNoSubjectLink()) {
+            if ($link === null) {
+                $title->addHtml(HtmlElement::create('span', [ 'class' => 'subject'], $template));
+            } else {
+                $title->addHtml(TemplateString::create(
+                    $template,
+                    ['link' => HtmlElement::create('span', [ 'class' => 'subject'])],
+                    $link
+                ));
+            }
+        } else {
+            if ($link === null) {
+                $title->addHtml(new Link($template, Links::downtime($this->item)));
+            } else {
+                $title->addHtml(TemplateString::create(
+                    $template,
+                    ['link' => new Link('', Links::downtime($this->item))],
+                    $link
+                ));
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


nilmerg (changes requested):
This is now definitely subject to proper localization.

Please construct the link using `TemplateString`.

```php
TemplateString::create(
    t('{{#downtime}}Fixed Downtime{{/downtime}} for %s', '..for <service> on <host>'),
    ['downtime' => $downtimeLink],
    $link
);
```

flourish86:
Not sure, if this is how `TemplateString` is supposed to be used, but it seems to work.

34ba268

nilmerg:
Sorry, Flo, but did you even compare it with my example? I showed you how it's supposed to be used, the difference to your change should be clear, as it doesn't make any difference compared to the previous solution.

Please check again.

library/Icingadb/Widget/ItemList/BaseDowntimeListItem.php
```diff
@@ -131,24 +132,25 @@ protected function assembleTitle(BaseHtmlElement $title)
...
+        }
+
         if ($this->getNoSubjectLink()) {
-            $title->addHtml(new HtmlElement(
-                'span',
-                Attributes::create(['class' => 'subject']),
-                Text::create($this->item->is_flexible ? t('Flexible Downtime') : t('Fixed Downtime'))
+            $title->addHtml(TemplateString::create(
+                $template,
+                ['downtime' => HtmlElement::create('span')],
```
nilmerg, line 144: The span is missing the `subject` class, please re-add that.

nilmerg (changes requested):

nilmerg:
And while looking at the final result now, I'd say we should rather use `link` as the template placeholder name instead of `downtime`. This way a translator knows that this is a link.

library/Icingadb/Widget/ItemList/BaseDowntimeListItem.php
flourish86: Oh, thanks. Missed that.
✅

library/Icingadb/Widget/ItemList/BaseDowntimeListItem.php
```diff
@@ -131,24 +132,30 @@ protected function assembleTitle(BaseHtmlElement $title)
...
-                    : t('Fixed Downtime'),
-                Links::downtime($this->item)
+            $title->addHtml(TemplateString::create(
+                $template,
+                ['link' => new Link('', Links::downtime($this->item))],
+                $link
             ));
         }
-
-        if ($link !== null) {
```
nilmerg, line 149: You have no alternative for this. If the object link is disabled it looks this way in an object's detail view:
Screenshot from 2021-07-15 10-25-48

nilmerg (changes requested):

library/Icingadb/Widget/ItemList/BaseDowntimeListItem.php
flourish86: ✅

library/Icingadb/Widget/ItemList/BaseDowntimeListItem.php
```diff
@@ -131,23 +132,28 @@ protected function assembleTitle(BaseHtmlElement $title)
...
         }
 
         if ($link !== null) {
-            $title->addHtml(Text::create(': '), $link);
+            $template .= ' for %s';
+        }
+
+        if ($this->getNoSubjectLink()) {
+            $title->addHtml(TemplateString::create(
+                t($template),
```
nilmerg, line 147: Localizable strings **must** be complete. They **must not** be variables, only strings are supported.

`t('foo')` is fine, `t($foo)` not. `t('abcd' . 'efgh')` is fine, `t('abcd' . $efgh)` not. Oh, and `t('abcd' . $bla ? 'blub' : 'blubber')` is also not supported.

nilmerg (changes requested):

library/Icingadb/Widget/ItemList/BaseDowntimeListItem.php
flourish86: ✅

library/Icingadb/Widget/ItemList/BaseDowntimeListItem.php
```diff
@@ -131,24 +132,33 @@ protected function assembleTitle(BaseHtmlElement $title)
             $link = $this->createServiceLink($this->item->service, $this->item->service->host, true);
         }
 
+        if ($this->item->is_flexible) {
+            if ($link !== null) {
+                $template = t('{{#link}}Flexible Downtime{{/link}} for %s');
+            } else {
+                $template = t('{{#link}}Flexible Downtime{{/link}}');
```
nilmerg, line 139: If we don't require a template string, we shouldn't use one. It only makes the translation more difficult. Every translatable string that is not used anywhere else must be explicitly translated. Even a single difference like a comma or so leads to this. Here, the `{{#link}}` and `{{/link}}` placeholders are redundant because there's no need for them. They are only required due to the addition of the `for`, which is missing here.

So what I would like to see instead, is that in case of no object link, the subject is constructed the same as previously. Which leads to e.g. `t('Fixed Downtime')` again and we have less redundancy in the translation message catalogs.

nilmerg (changes requested):

nilmerg (approved):

PR #227, opened 2021-07-28
display user and usergroups customvars

### LINKED ISSUE #19: User custom vars should be displayed
https://github.com/Icinga/icingaweb2/issues/4070

@N-o-X mentioned that the new schema makes it possible to display the custom vars, so we might want to do that! :)

sukhwinder33445:
fix #19

### DIFF
--- a/library/Icingadb/Widget/Detail/UserDetail.php
+++ b/library/Icingadb/Widget/Detail/UserDetail.php
@@ -1,0 +1,85 @@
+<?php
+
+/* Icinga DB Web | (c) 2021 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Icingadb\Widget\Detail;
+
+use Icinga\Module\Icingadb\Common\Auth;
+use Icinga\Module\Icingadb\Common\Links;
+use Icinga\Module\Icingadb\Model\User;
+use Icinga\Module\Icingadb\Widget\EmptyState;
+use Icinga\Module\Icingadb\Widget\HorizontalKeyValue;
+use Icinga\Module\Icingadb\Widget\ItemList\UsergroupList;
+use Icinga\Module\Icingadb\Widget\ShowMore;
+use ipl\Html\BaseHtmlElement;
+use ipl\Html\HtmlElement;
+use ipl\Html\Text;
+
+class UserDetail extends BaseHtmlElement
+{
+    use Auth;
+
+    /** @var User The given user */
+    protected $user;
+
+    protected $defaultAttributes = ['class' => 'user-detail'];
+
+    protected $tag = 'div';
+
+    public function __construct(User $user)
+    {
+        $this->user = $user;
+    }
+
+    protected function createCustomVars()
+    {
+        $content = [new HtmlElement('h2', null, Text::create(t('Custom Variables')))];
+        $flattenedVars = $this->user->customvar_flat;
+        $this->applyRestrictions($flattenedVars);
+
+        $vars = $this->user->customvar_flat->getModel()->unflattenVars($flattenedVars);
+        if (! empty($vars)) {
+            $customvarTable = new CustomVarTable($vars);
+            $customvarTable->setAttribute('id', 'user-customvars');
+            $content[] = $customvarTable;
+        } else {
+            $content[] = new EmptyState(t('No custom variables configured.'));
+        }
+
+        return $content;
+    }
+
+    protected function createUserDetail()
+    {
+        return [
+            new HtmlElement('h2', null, Text::create(t('Details'))),
+            new HorizontalKeyValue(t('E-Mail'), $this->user->email),
+            new HorizontalKeyValue(t('Pager'), $this->user->pager)
+        ];
+    }
+
+    protected function createUsergroupList()
+    {
+        $userGroups = $this->user->usergroup->limit(6)->peekAhead()->execute();
+
+        $showMoreLink = new ShowMore(
+            $userGroups,
+            Links::usergroups()->addParams(['user.name' => $this->user->name])
+        );
+
+        return [
+            new HtmlElement('h2', null, Text::create(t('Groups'))),
+            new UsergroupList($userGroups),
+            $showMoreLink
+        ];
+    }
+
+    protected function assemble()
+    {
+        $this->add([
+            $this->createUserDetail(),
+            $this->createUsergroupList(),
+            $this->createCustomVars()
+        ]);
+    }
+}
--- a/library/Icingadb/Widget/Detail/UserDetail.php
+++ b/library/Icingadb/Widget/Detail/UserDetail.php
@@ -5,6 +5,7 @@
 namespace Icinga\Module\Icingadb\Widget\Detail;
 
 use Icinga\Module\Icingadb\Common\Auth;
+use Icinga\Module\Icingadb\Common\Database;
 use Icinga\Module\Icingadb\Common\Links;
 use Icinga\Module\Icingadb\Model\User;
 use Icinga\Module\Icingadb\Widget\EmptyState;
@@ -18,6 +19,7 @@ use ipl\Html\Text;
 class UserDetail extends BaseHtmlElement
 {
     use Auth;
+    use Database;
 
     /** @var User The given user */
     protected $user;
--- a/library/Icingadb/Widget/Detail/UserDetail.php
+++ b/library/Icingadb/Widget/Detail/UserDetail.php
@@ -62,10 +62,10 @@ class UserDetail extends BaseHtmlElement
     {
         $userGroups = $this->user->usergroup->limit(6)->peekAhead()->execute();
 
-        $showMoreLink = new ShowMore(
+        $showMoreLink = (new ShowMore(
             $userGroups,
             Links::usergroups()->addParams(['user.name' => $this->user->name])
-        );
+        ))->setBaseTarget('_next');
 
         return [
             new HtmlElement('h2', null, Text::create(t('Groups'))),
--- a/library/Icingadb/Widget/Detail/UsergroupDetail.php
+++ b/library/Icingadb/Widget/Detail/UsergroupDetail.php
@@ -52,10 +52,10 @@ class UsergroupDetail extends BaseHtmlElement
     {
         $users = $this->usergroup->user->limit(6)->peekAhead()->execute();
 
-        $showMoreLink = new ShowMore(
+        $showMoreLink = (new ShowMore(
             $users,
             Links::users()->addParams(['usergroup.name' => $this->usergroup->name])
-        );
+        ))->setBaseTarget('_next');
 
         return [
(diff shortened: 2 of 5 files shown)


sukhwinder33445:
In the view "User groups" the container with the "custom variables" has no left margin/padding as in "Users".
This happens because of the class "full-width". Why do we need this class? And why not in the view "Users".

library/Icingadb/Widget/Detail/CustomVarsDetail.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+/* Icinga DB Web | (c) 2021 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Icingadb\Widget\Detail;
+
+use Icinga\Module\Icingadb\Model\User;
+
+class CustomVarsDetail extends ObjectDetail
```
nilmerg, line 9: Not a fan of the name to be honest. This is either a UserDetail or UserGroupDetail.

That it extends ObjectDetail is also not fitting. This implies that all inherited methods should work, which they don't.

Please use two different detail implementations (`UserDetail` and `UserGroupDetail`) and implement `createCustomVars()` for both and `createUserList()` for the latter.

nilmerg (changes requested):

nilmerg (changes requested):
* Place constants before properties
* Add doc blocks to new constants and properties
* Reconsider the use of the `objectType` property, seems redundant to me
* Use `HtmlElement` instead of `Html`, the latter shouldn't be used inside concrete widgets
* In `UsergroupDetail` you've used `$this->assemble()` correct, though in `UserDetail` it's rather redundant as `$this->createUserDetail()` does its job. Please use `$this->assemble()` correct there as well.

sukhwinder33445:
> * In `UsergroupDetail` you've used `$this->assemble()` correct, though in `UserDetail` it's rather redundant as `$this->createUserDetail()` does its job. Please use `$this->assemble()` correct there as well.

In `UserDetail` the custom variables are also part of the user detail, so I added them to `$this->createUserDetail()`. Can you please explain why this is not correct?

nilmerg:
`assemble` is responsible for the setup of a widget. Though, it *usually* does very little as it delegates this job to specific other methods. These other methods then to a specific thing, like creating a single widget and its headline. Composing multiple widgets is still `assemble`s job. `createUserDetail` composes, which it shouldn't.

library/Icingadb/Widget/Detail/UserDetail.php
```diff
@@ -0,0 +1,110 @@
...
+            ))
+        ];
+    }
+
+    protected function assemble()
+    {
+        $this->add([
+            $this->createUserDetail(),
+            (new HtmlElement('h2', null, Text::create(t('User Groups')))),
+            new UsergroupList($this->user->usergroup),
```
nilmerg, line 106: This is more of a `createUsergroupList()`. And while you're at it, also add the `ShowMore` widget to it.

library/Icingadb/Widget/Detail/UserDetail.php
```diff
@@ -0,0 +1,110 @@
...
+            $customvarTable->setAttribute('id', 'user-customvars');
+            $content[] = $customvarTable;
+        } else {
+            $content[] = new EmptyState(t('No custom variables configured.'));
+        }
+
+        return $content;
+    }
+
+    protected function createUserDetail()
```
nilmerg, line 50: * Why the parentheses around the constructors?
* Please use a `HorizontalKeyValue` widget for each key-value-pair, the `.key-value-list` isn't used anywhere else now and actually obsolete.

library/Icingadb/Widget/Detail/UsergroupDetail.php
```diff
@@ -0,0 +1,90 @@
...
+                        $url->setParam('fullList')
+                    ));
+
+                    break;
+                }
+                    $list[$key] = $user;
+            }
+
+            $userList = $list;
+        }
```
nilmerg, line 74: Have you taken a look at how the *show more* link is setup elsewhere? (Hint: There's a `ShowMore` widget) What you're doing here is .. interesting :joy:

library/Icingadb/Widget/Detail/UsergroupDetail.php
```diff
@@ -0,0 +1,90 @@
...
+
+        return [
+            new HtmlElement('h2', null, Text::create(t('Hosts'))),
```
nilmerg, line 77: It's `Users`, isn't it?

nilmerg (changes requested):

library/Icingadb/Widget/Detail/UserDetail.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        $userGroups = $this->user->usergroup->limit(6)->peekAhead()->execute();
+
+        $showMoreLink = new ShowMore(
+            $userGroups,
+            Links::usergroups()->addParams(['user.name' => $this->user->name])
+        );
+
+        return [
+            new HtmlElement('h2', null, Text::create(t('User Groups'))),
```
nilmerg, line 71: We're using just `Groups` in other detail views already. Do so here too.

library/Icingadb/Widget/Detail/UserDetail.php
```diff
@@ -0,0 +1,85 @@
...
+    use Auth;
+
+    /** @var User The given user */
+    protected $user;
+
+    protected $defaultAttributes = ['class' => 'user-detail'];
+
+    protected $tag = 'div';
+
+    public function __construct($user)
```
nilmerg, line 29: The docblock for the `$user` property hints at `User`, so please also require the `$user` param to be an instance of that.

library/Icingadb/Widget/Detail/UsergroupDetail.php
```diff
@@ -0,0 +1,74 @@
...
+    use Auth;
+
+    /** @var Usergroup The given user group */
+    protected $usergroup;
+
+    protected $defaultAttributes = ['class' => 'usergroup-detail'];
+
+    protected $tag = 'div';
+
+    public function __construct($usergroup)
```
nilmerg, line 28: The docblock for the `$usergroup` property hints at `Usergroup`, so please also require the `$usergroup` param to be an instance of that.

nilmerg (changes requested):
Mostly finished now :smiley:

nilmerg (changes requested):
One simple change remaining: Please let the user/-group list open in a new column when clicking `ShowMore`.

nilmerg (approved):

nilmerg (changes requested):
Sorry, but I've got an error just now. When configuring customvar protection rules this shows up in the user and usergroup detail:

```
Uncaught Error: Call to undefined method Icinga\Module\Icingadb\Widget\Detail\UserDetail::getDb() in /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Common/Auth.php:214
Stack trace:
#0 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Widget/Detail/UserDetail.php(38): Icinga\Module\Icingadb\Widget\Detail\UserDetail->applyRestrictions(Object(ipl\Orm\Query))
#1 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Widget/Detail/UserDetail.php(82): Icinga\Module\Icingadb\Widget\Detail\UserDetail->createCustomVars()
#2 /usr/local/src/ipl-html/src/HtmlDocument.php(295): Icinga\Module\Icingadb\Widget\Detail\UserDetail->assemble()
#3 /usr/local/src/ipl-html/src/HtmlDocument.php(496): ipl\Html\HtmlDocument->ensureAssembled()
#4 /usr/local/src/ipl-html/src/HtmlDocument.php(339): ipl\Html\HtmlDocument->render()
#5 /usr/local/src/ipl-html/src/BaseHtmlElement.php(255): ipl\Html\HtmlDocument->renderUnwrapped()
#6 /usr/local/src/ipl-html/src/BaseHtmlElement.php(324): ipl\Html\BaseHtmlElement->renderContent()
#7 /usr/local/src/ipl-html/src/HtmlDocument.php(498): ipl\Html\BaseHtmlElement->renderUnwrapped()
#8 /usr/local/src/ipl-html/src/HtmlDocument.php(339): ipl\Html\HtmlDocument->render()
#9 /usr/local/src/ipl-html/src/HtmlDocument.php(498): ipl\Html\HtmlDocument->renderUnwrapped()
#10 /usr/local/src/ipl-web/src/Compat/ViewRenderer.php(56): ipl\Html\HtmlDocument->render()
#11 /icingaweb2/library/vendor/Zend/Controller/Action/Helper/ViewRenderer.php(964): ipl\Web\Compat\ViewRenderer->render()
#12 /icingaweb2/library/vendor/Zend/Controller/Action/HelperBroker.php(272): Zend_Controller_Action_Helper_ViewRenderer->postDispatch()
#13 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Web/Controller.php(481): Zend_Controller_Action_HelperBroker->notifyPostDispatch()
#14 /icingaweb2/library/Icinga/Web/Controller/Dispatcher.php(76): Icinga\Module\Icingadb\Web\Controller->dispatch('indexAction')
#15 /icingaweb2/library/vendor/Zend/Controller/Front.php(937): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Module\Partnerportal\Web\Response))
#16 /icingaweb2/library/Icinga/Application/Web.php(304): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Module\Partnerportal\Web\Response))
#17 /icingaweb2/library/Icinga/Application/webrouter.php(107): Icinga\Application\Web->dispatch()
#18 /icingaweb2/public/index.php(4): require_once('/icingaweb2/lib...')
#19 {main}
```

sukhwinder33445:
Can you please confirm if this fixes the problem? If not, please tell me how I can reproduce it. I added a given custom variable to `Configuration->Modules->Monitoring->Security` and logged in again. Unfortunately, nothing changes. Did I miss something?

nilmerg (approved):

nilmerg:
This is Icinga DB Web, not the monitoring module, hence you'd have to add a custom variable to the protection restriction in a role of yours.

flourish86:
I got asked by @nilmerg, if the custom vars should be put before the list (user(-group) list).

We have the lists limited to 6 items max. and the custom vars will be collapsible at some point. So, we don't have the problem that one of the sections might be far out of reach.

Rather, I'd decide, from what's more important to most users.

I see two points, why I would have the custom vars below the list:

1) Especially in the user group detail I'd assume that most users would typically expect the contents of the group, i.e. the user list.

2) The contents of the custom vars are not determined. It's usually better to have the element, that has more expectable content be on top, i.e. a more predictable location to spot. (The farther down an element is in the detail view, the less expectable is its location).

PR #228, opened 2021-08-02
Remove redudant titles

### LINKED ISSUE #104: Redundant titles
We should avoid redundant titles wherever possible. e.g. in the host detail view the `Services` and `Host Group` titles are such a case.

Screenshot from 2020-06-15 08-53-19

sukhwinder33445:
fix #104

### DIFF
--- a/library/Icingadb/Widget/Detail/HostDetail.php
+++ b/library/Icingadb/Widget/Detail/HostDetail.php
@@ -29,9 +29,7 @@ class HostDetail extends ObjectDetail
             $services = new EmptyState(t('This host has no services'));
         }
 
-        $stats = [Html::tag('h2', t('Services'))];
-        $stats[] = new HorizontalKeyValue(t('Services'), $services);
-        return $stats;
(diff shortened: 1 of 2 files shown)


nilmerg (changes requested):
A `HorizontalKeyValue` widget has only one purpose: Showing a key/value pair horizontally.

If there's no key/value pair to show, don't use it. Use the value as is.

library/Icingadb/Widget/Detail/HostDetail.php
```diff
@@ -30,7 +30,7 @@ protected function createServiceStatistics()
         }
 
         $stats = [Html::tag('h2', t('Services'))];
-        $stats[] = new HorizontalKeyValue(t('Services'), $services);
+        $stats[] = $services;
         return $stats;
```
nilmerg, line 34: Please combine this to a single instruction.

nilmerg (changes requested):

nilmerg (approved):

PR #234, opened 2021-08-11
Host/Service comment hover bubble

### LINKED ISSUE #231: Show last host/service comment in a popup in detailed mode
Currently we only indicate in detailed list mode that a host or service has a comment. But who authored it and the comment itself is not shown.

Upon hovering over the comment icon in the list item's footer, a popup should open showing the comment's author and the comment's content.

## Additional context

* https://github.com/Icinga/icingadb-web/blob/master/library/Icingadb/Widget/ServiceListItemDetailed.php#L34-L37
* https://github.com/Icinga/icingadb-web/blob/master/library/Icingadb/Widget/HostListItemDetailed.php#L34-L37
* https://github.com/Icinga/icingadb/issues/317

sukhwinder33445:
resolves #231

### DIFF
--- a/public/css/widget/comment-popup.less
+++ b/public/css/widget/comment-popup.less
@@ -40,7 +40,8 @@ ul.item-list li:last-child .comment-wrapper {
   }
 
   .comment-popup:before {
-    bottom: -0.86em;
+    bottom: ~"calc(-0.75em - 1px)";
+    top: unset;
     transform: rotate(225deg);
   }
 }


library/Icingadb/Widget/HostListItemDetailed.php
```diff
@@ -32,8 +33,17 @@ protected function assembleFooter(BaseHtmlElement $footer)
         $statusIcons = new HtmlElement('div', Attributes::create(['class' => 'status-icons']));
 
         // ToDo(fs): Get `has_comments` from database
-        if ($this->item->comment->limit(1)->execute()->hasResult()) {
-            $statusIcons->addHtml(new Icon('comments', ['title' => t('This item has been commented')]));
+        if (! empty($this->item->state->last_comment->text)) {
```
nilmerg, line 36: Why `->text`? Empty comments can't be a thing anyway.

library/Icingadb/Widget/HostListItemDetailed.php
```diff
@@ -32,8 +33,17 @@ protected function assembleFooter(BaseHtmlElement $footer)
...
+            $comment = $this->item->state->last_comment;
+            $comment->host = $this->item;
+            $comment = (new CommentList([$comment]))
+                ->setNoSubjectLink()
+                ->setDetailActionsDisabled();
+
+            $statusIcons->addHtml(
+                (new Icon('comments', ['class' => 'popup']))
+                ->add(new HtmlElement('div', Attributes::create(['class' => 'comment-popup']), $comment))
```
nilmerg, line 45: You're adding the popup (block element) to the icon (inline element). That's not valid HTML.

nilmerg (changes requested):

library/Icingadb/Widget/HostListItemDetailed.php
sukhwinder33445: Because `..state->last-comment` is a `model/comment` instance with null values, but not null.

nilmerg: Ah, I see. Then please check for `last_comment_id !== null`. A comment's text is nowhere the deciding factor if a comment is considered existent.

nilmerg:
> fix `#231`

Please use `fix(es) #id` for bugs only. Use `resolves #id` for enhancements.

nilmerg (changes requested):
The object link in the popup doesn't work. I have no idea why, but now I think it's better to show the popup below by default and only for the *last* item above. You could also play with showing the popup on the right, maybe that works always?

nilmerg:
> The object link in the popup doesn't work. I have no idea why, but now I think it's better to show the popup below by default and only for the last item above. You could also play with showing the popup on the right, maybe that works always?

Ah sorry, and disabling the object link of course.

public/css/common.less
```diff
@@ -36,7 +36,7 @@
   .item-list .list-item:last-of-type .caption {
     min-height: 1.5em;
     max-height: 3em;
-    height: auto;
+    //height: auto;  can we remove this? because the picture in the last comment does not fit properly
```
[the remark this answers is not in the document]
sukhwinder33445: max-height of 3em is already given one line above, but unfortunately has no effect. 
max-height should overwrite the height, but this does not happen here.

public/css/common.less
```diff
@@ -328,3 +328,96 @@ div.show-more {
...
+  left: -1.6em;
+  z-index: 1;
+  font-size: 1.2em;
+  /*max-width: 35em;
+  width: 35em;
+  display: inline-block;*/
+  height: 6em;
+  border-radius: 0.25em;
+  border: #ccc 1px solid;
+  box-shadow: rgba(17, 17, 26, 0.1) 0px 1px 0px, rgba(17, 17, 26, 0.1) 0px 8px 24px, rgba(17, 17, 26, 0.1) 0px 16px 48px;
```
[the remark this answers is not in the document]
sukhwinder33445: I removed the shadow, but I think it looks (as a popup) much better with shadow. 🤔

public/css/common.less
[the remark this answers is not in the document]
nilmerg: With a set height, min-height and max-height have no effect. That's why height:auto is required. max-height never overrides height.

> I added max-height to image.

What I suggested :)

public/css/common.less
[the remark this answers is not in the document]
nilmerg: After I saw the instruction, I really had to turn it off / on to even see that there's a shadow :joy:

public/css/common.less
```diff
@@ -328,3 +328,96 @@ div.show-more {
...
+  }
+}
+
+
+ul.item-list.detailed li:first-child .comment-wrapper{
+  .comment-popup {
+    top: 2.5em;
+
+    .user-ball {
+      z-index: 1;
```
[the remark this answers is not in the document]
sukhwinder33445: To avoid comment-popup:before element overlap.

public/css/common.less
```diff
@@ -328,3 +328,96 @@ div.show-more {
...
+    }
+  }
+
+  .comment-popup:before {
+    bottom: 5.2em;
+    transform: rotate(45deg);
+  }
+}
+
+@media only screen and (max-width: 800px) {
```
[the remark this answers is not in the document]
sukhwinder33445: I didn't know we had a variety of layouts 😆  🙈 . That's great.

sukhwinder33445:
> > The object link in the popup doesn't work. I have no idea why, but now I think it's better to show the popup below by default and only for the last item above. You could also play with showing the popup on the right, maybe that works always?
> 
> Ah sorry, and disabling the object link of course.

Since we display the pop-up below by default, we no longer need the host/service name in the pop-up. Am I right? Because otherwise it is superfluous.

nilmerg:
> Am I right? Because otherwise it is superfluous.

Yes, that's what I've forgot in the review to mention and what I did in the comment afterwards.

public/css/common.less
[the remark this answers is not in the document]
sukhwinder33445: Should I add shadow, but make it shorter? 😄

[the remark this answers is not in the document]
nilmerg: Nah, the popup has a border, that should be enough? I'm not against the shadow itself, I rather think it's hard to do for little effect, as you'd need to make it variable just like the border and that's really difficult.

public/css/common.less
```diff
@@ -328,3 +332,93 @@ div.show-more {
 .servicegropup-list > col.object-statistics-graph {
   padding-right: .25em;
 }
+
+.comment-popup {
+  font-size: 1.2em;
```
nilmerg, line 337: This results for me in 12.3408px. Make it so that it's exactly 12px.

public/css/common.less
```diff
@@ -328,3 +332,93 @@ div.show-more {
...
+    overflow: auto;
+
+    // This is necessary to limit the visible comment lines
+    // and to fit the image into the comment pop-up in the host/service detail view.
+    .caption {
+      height: 3em;
+
+      img {
+        max-height: 1.7em;
+      }
```
nilmerg, line 355: * Why 1.7em? (Because it fits best isn't an explanation :wink: )
* Also, if you expect images not extending over one line of text in the popup, why not in the standard lists?

nilmerg (changes requested):

public/css/common.less
sukhwinder33445: I did it manually, it took a little while, but I fixed it. Is there an easy way to do this? The result of the Google converter was not correct.

nilmerg: `1em` are `12px` at the default font size in your scope. For the footer the font-size is reduced to `0.857em`, that are 85,7% of `1em` or `12px`. To get the font-size back to `12px` in relation to `0.857em`, the font-size has to be `1.166861144em` (1 / 0.857).

Since browsers tend to optimize fractions in numbers differently, we usually place the calculation inside the stylesheet to also emphasize what is calculated there: `font-size: 1/.857em; // default font size / footer font size = 12px`

public/css/common.less
sukhwinder33445: Should I make them as small as the text `(12px max-height)`? Example below: 
Screenshot 2021-08-19 at 11 04 52

And same for `CommentController`. And then just show the original size in the comment detail, as it is at the moment.

nilmerg: :+1:

public/css/common.less
sukhwinder33445: Thanks for the explanation 😊  .

public/css/common.less
```diff
@@ -328,3 +328,89 @@ div.show-more {
...
+  border-radius: 0.25em;
+  border: @gray-light 1px solid;
+  background-color: @body-bg-color;
+
+  .comment-list .main {
+    border: none;
+    width: auto;
+    overflow: auto;
+
+    // This is necessary to limit the visible comment lines
```
nilmerg, line 344: Please add: *because the popup is shown in detailed list mode only*

nilmerg (changes requested):
* The max-min-widths have no structure and don't work at all
  * With two cols you define a different max, but min is higher?
  * The widths have no effect because the popup is positioned
    * As such it's out of the flow and has no size reference (It's not contained by anything with a size). To solve that, give the `.comment-popup` a defined `width` depending on what the `max-width` would be. Also remove the `width: auto; overflow: auto` instructions from `.comment-list .main`, they're not required.
* The wrapper is a block element and pushes other icons to the next line, consider to make `.status-icons` also a flex container.
  Screenshot from 2021-08-20 12-08-23

sukhwinder33445:
> * Also remove the `width: auto; overflow: auto`

This causes the pop-up not to increase its width if the username is longer. The pop-up now works as a fixed-width element. 🤔 

> * give the `.comment-popup` a defined `width` depending on what the `max-width` would be.
This confuses me, you mean I should add the width to comment popup instead of adding it to comment list? The content is in the comment list.

nilmerg (changes requested):
I believe you can increase the widths a "bit" :thinking: There's not much difference now between 25em and 42em. I think both are too narrow, at least on my 2560px-width screen here.

sukhwinder33445:
What do you recommend? 50em is OK?

nilmerg:
50em default. 35em for `.twocols`. 25em for `.minimal-layout, .poor-layout.two-cols, .compact-layout.two-cols`.

nilmerg:
And don't forget the flex-box for `.status-icons`.

public/css/common.less
```diff
@@ -328,3 +328,93 @@ div.show-more {
...
+    // because the popup is shown in detailed list mode only
+    .caption {
+      height: 3em;
+    }
+  }
+}
+
+li.list-item .status-icons {
+  display: flex;
+  align-items: center;
```
nilmerg, line 352: Please add this where the general style/layout rules are for the `.status-icons`.

public/css/common.less
```diff
@@ -328,3 +328,93 @@ div.show-more {
...
+
+  .comment-wrapper {
+    position: relative;
+    display: inline-block;
```
nilmerg, line 356: Remove that. `.status-icons` is now using flex-box after all.

nilmerg (changes requested):
Please move all your `.coment-popup` related changes to `public/css/widgets/comment-popup.less`

sukhwinder33445:
Screenshot 2021-08-23 at 14 37 45

Why this is so is a mystery to me. Can you help please? 😕

public/css/list/list-item.less
```diff
@@ -59,6 +59,8 @@
 
   footer {
     .status-icons {
+      display: flex;
+      align-items: center;
```
nilmerg, line 63: There's a large `// Layout` area below, why not add this there? :wink:

public/css/common.less
```diff
@@ -328,3 +328,36 @@ div.show-more {
...
+    }
+  }
+
+  &.poor-layout,
+  &.minimal-layout {
+    .comment-popup {
+      width: 25em;
+    }
+  }
+}
```
nilmerg, line 363: That's also related to the popup only

nilmerg (changes requested):

nilmerg:
> Why this is so is a mystery to me. Can you help please? 

Only if you tell me how to reproduce that. Doesn't happen for me at least with two such list-items.

public/css/common.less
sukhwinder33445: It is only displayed with committed changes. I don't have any local changes. Maybe it only appears on my site and it works fine elsewhere?

nilmerg: What do you mean? That's the current pushed state, isn't it?

sukhwinder33445: Yes. The current pushed state.

sukhwinder33445:
> Only if you tell me how to reproduce that. Doesn't happen for me at least with two such list-items.

It is displayed with the current pushed state. I don't have any local changes. Maybe it only appears on my site and it works fine elsewhere?

public/css/list/list-item.less
```diff
@@ -110,6 +110,11 @@
       line-height: 1.5/.857em;
     }
 
+    .state-icons {
```
nilmerg, line 113: It's `.status-icons`

nilmerg (changes requested):

nilmerg:
> > Only if you tell me how to reproduce that. Doesn't happen for me at least with two such list-items.

> It is displayed with the current pushed state. I don't have any local changes. Maybe it only appears on my site and it works fine elsewhere?

If it's not multiple `.comment-popup` elements inside a `.comment-wrapper`, I don't see a reason why this would occur. Maybe you still have/had the `:hover` state for one of the items enabled in the dev tools?

sukhwinder33445:
> If it's not multiple `.comment-popup` elements inside a `.comment-wrapper`, I don't see a reason why this would occur. Maybe you still have/had the `:hover` state for one of the items enabled in the dev tools?

No, this is a misunderstanding. I don't mean why two pop-up windows appear (I activated them manually). Just look at the arrow I have highlighted and the box below it.

nilmerg:
Aaah, I see. Try `top: calc(-0.75em - .5px)` instead of `bottom: 5.2em`. That's the height of :before halfed + the 1 pixel border halfed.

nilmerg (approved):

PR #239, opened 2021-08-18
Use the same semi-infinite scrolling for the notification overview

sukhwinder33445:


nilmerg (changes requested):
Since this is mostly the same code as for the history overview, I'd like to see a trait being used for this functionality:

* Create a new trait `Icinga\Module\Icingadb\Common\LoadMore`
* Put the new properties of `NotificationList` and methods into this trait (except `init`)
* Use the trait inside `NotificationList` and `HistoryList`
* Expect a `ResultSet` in method `getIterator` as param (that's essentially `realData`)
* Introduce a new method `setLoadMoreUrl` into the trait
  * Use that method in the controller when creating the `NotificationList` and pass it the url with the before param
* Don't add the `ShowMore` widget in the controller, but let `getIterator` add it once the actual data is exhausted. With the url, page size and page number it should be possible to do that.

library/Icingadb/Common/LoadMore.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public function setPageNumber($number)
+    {
+        $this->pageNumber = $number;
+
+        return $this;
+    }
+
+    /**
+     * Set the page number
```
nilmerg, line 51: That's not a page number.

library/Icingadb/Common/LoadMore.php
```diff
@@ -0,0 +1,95 @@
...
+     * @return $this
+     */
+    public function setLoadMoreUrl(Url $url)
+    {
+        $this->loadMoreUrl = $url;
+
+        return $this;
+    }
+
+    protected function getIterator($result)
```
nilmerg, line 64: * Add phpdoc
* Require that `$result` is a `ResultSet`
  * Then override `__construct` in `HistoryList` and `NotificationList` to also require there that `$data` is a `ResultSet` (but then just call the parent constructor, it's just about the type declaration)

nilmerg (changes requested):
Nice! :+1: Just a few small changes left :)

library/Icingadb/Common/LoadMore.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public function setPageNumber($number)
+    {
+        $this->pageNumber = $number;
+
+        return $this;
+    }
+
+    /**
+     * Set the url for LoadMore widget
```
nilmerg, line 52: `LoadMore` doesn't exist as widget, at least I read this as if there's a `Icingadb\Widget\LoadMore`. Just describe what the url is used for: to fetch *more* items

library/Icingadb/Common/LoadMore.php
nilmerg: Also add the type declaration for the parameter in the method interface.

library/Icingadb/Common/LoadMore.php
```diff
@@ -0,0 +1,101 @@
...
+     */
+    public function setLoadMoreUrl(Url $url)
+    {
+        $this->loadMoreUrl = $url;
+
+        return $this;
+    }
+
+    /**
+     * Add the page separator and the "LoadMore" button at the desired position
```
nilmerg, line 66: That's the description of what it additionally does (if it begins with *Adds*), not the main functionality. Keep that with *Adds* as extended description and add a main description what the method returns. And on that note, also add a `@returns Generator` annotation.

nilmerg (changes requested):

nilmerg (changes requested):
Please check all other usages of `HistoryList` and `NotificationList`. It's not just the EventController which is affected.

sukhwinder33445:
I have already checked. NotificationList is not used anywhere else and HistoryList already gets a Resultset parameter (used in the HostController, ServiceController and in the EventController (fixed)).

nilmerg:
Aye, but it's not just the constructor you've changed :wink:

sukhwinder33445:
Yes, I also changed the getIterator method in the LoadMore trait, but it has no effect. Am I missing something?

nilmerg:
Just an example: https://github.com/Icinga/icingadb-web/blob/feature/add-loadmore-to-notificationsController/application/controllers/HostController.php#L163-L164

sukhwinder33445:
Oh, I see. It is no longer necessary to use ShowMore in the controller. I will change it soon.

nilmerg (approved):

PR #240, opened 2021-08-18
HostMetaInfo: Add dots instead of underscores

sukhwinder33445:
To avoid that users misinterpret the host detail labels as filter columns.

### DIFF
--- a/library/Icingadb/Widget/Detail/HostMetaInfo.php
+++ b/library/Icingadb/Widget/Detail/HostMetaInfo.php
@@ -31,12 +31,12 @@ class HostMetaInfo extends BaseHtmlElement
     protected function assemble()
     {
         $this->addHtml(
-            new VerticalKeyValue('host_name', $this->host->name),
+            new VerticalKeyValue('host.name', $this->host->name),
             new HtmlElement(
                 'div',
                 null,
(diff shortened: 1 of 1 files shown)


library/Icingadb/Widget/Detail/HostMetaInfo.php
```diff
@@ -31,15 +31,15 @@ public function __construct(Host $host)
...
                 'div',
                 null,
-                new HorizontalKeyValue('host_address', $this->host->address ?: '-'),
-                new HorizontalKeyValue('host_address_v6', $this->host->address6 ?: '-')
+                new HorizontalKeyValue('host.address', $this->host->address ?: '-'),
+                new HorizontalKeyValue('host.address6', $this->host->address6 ?: '-')
             ),
             new VerticalKeyValue(
-                'last_state_change',
+                'host.state.last_state_change',
```
nilmerg, line 42: We should leave the relation path out here, the column name explains enough that it's not strictly required. Otherwise the host name is cut if the v6 address is very long:

Screenshot from 2021-08-19 08-29-58
Screenshot from 2021-08-19 08-29-42

nilmerg (changes requested):

nilmerg (approved):

PR #243, opened 2021-08-19
Bugfix ObjectInspectionDetail::formatTimestamp()

### LINKED ISSUE #242: Uncaught Error: Call to a member function setTimezone() on bool
## Describe the bug
When pressing any service object in `Services` which is in the `NOT-AVAILABLE` state and navigating to the `Source` tab, you will encounter the following error message.

Bildschirmfoto 2021-08-19 um 11 24 39

sukhwinder33445:
fixes #242


sukhwinder33445:
It works well now, but I wonder how? 🤣  🙈

sukhwinder33445:
I found the reason. $ts was 0.0 (float) and therefore `0.0 === 0` was wrong.

nilmerg (approved):

PR #245, opened 2021-08-19
configuration.php: Add service grid icon

sukhwinder33445:


sukhwinder33445:
I actually wanted to add the "th-thumb-empty" icon, but it didn't work. Do you know the reason for this @nilmerg @flourish86 ?

nilmerg:
> I actually wanted to add the "th-thumb-empty" icon, but it didn't work. Do you know the reason for this

Because it [doesn't exist](https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=th-thumb-empty&m=free).

nilmerg:
I'm not sure the `th-` icons even fit for this here. They're usually related to interactable elements (such as sub-menus). Maybe we should still use the icon for services as previously intended?

sukhwinder33445:
You mean this one: 
Screenshot 2021-08-20 at 12 21 57

Should I add it?

nilmerg (approved):

PR #246, opened 2021-08-20
Check and fix licence header

sukhwinder33445:


nilmerg (changes requested):
There's still `Sni.php`, `PivotTable.php` and `ProblemToggle.php` which miss the header

sukhwinder33445:
Oh, sorry, I don't know how I missed that?😕

nilmerg (approved):

Issue #252, opened 2021-08-23
Missing `name` parameter when changing view mode in hostgroup-/servicegroup-detail


nilmerg:
Trying to switch to another viewmode in either view results in an error: Required parameter 'name' missing

PR #254, opened 2021-08-23
Controller: Fix uncaught argument count error

### LINKED ISSUE #248: Uncaught ArgumentCountError: Too few arguments
## Describe the bug
As this [PR](https://github.com/Icinga/icingadb-web/pull/218) was created before this 262a17598c520ee294d7592ab3925bd6d4c3b129 commit was being merged, unfortunately there is now problems because the signature of `Controller::createViewModeSwitcher()` method has been changed.

Bildschirmfoto 2021-08-20 um 16 36 51

Yonas-net:
resolves #248

### DIFF
--- a/application/controllers/HistoryController.php
+++ b/application/controllers/HistoryController.php
@@ -40,18 +40,16 @@ class HistoryController extends Controller
 
         $before = $this->params->shift('before', time());
         $url = Url::fromPath('icingadb/history')->setParams(clone $this->params);
-        if (! $this->params->has('page') || ($page = (int) $this->params->shift('page')) < 1) {
-            $page = 1;
-        }
 
         $limitControl = $this->createLimitControl();
+        $paginationControl = $this->createPaginationControl($history);
         $sortControl = $this->createSortControl(
             $history,
             [
                 'history.event_time desc' => t('Event Time')
             ]
         );
-        $viewModeSwitcher = $this->createViewModeSwitcher();
+        $viewModeSwitcher = $this->createViewModeSwitcher($paginationControl, $limitControl);
         $searchBar = $this->createSearchBar($history, [
             $limitControl->getLimitParam(),
             $sortControl->getSortParam(),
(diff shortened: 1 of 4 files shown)


nilmerg (changes requested):
I'd rather make `icingadb/history`, `icingadb/notifications`, `icingadb/host/history` and `icingadb/service/history` use of `PaginationControl`. Albeit only internally, they shouldn't display it, but only utilize it. It can be used to identify the current page (`$page`) and the limit is already applied on the query by `createPaginationControl()`.

This also solves the problem that the viewmode switcher currently doesn't behave the same as in the other lists. For the host/service history this of course means that they need to show the viewmode switcher. But this is something that would have turned up the next days anyway.

Yonas-net:
> `icingadb/host/history` and `icingadb/service/history` use of `PaginationControl`

This should be noted for later implementation, as these do not support view mode switchers so far.

nilmerg (changes requested):
* Please create a separate PR for the #252 changes.
* Please go a step further as I already noted:
  * Initialize `$page` with `$paginationControl->getCurrentPageNumber()`
  * Don't apply the limit a second time (PaginationControl::apply() does that first)
    * Only the limit in case of `$page > 1 && ! $compact` is still relevant
  * Don't apply the offset a second time (PaginationControl::apply() does that first)

Yonas-net:
I don't quite follow what you're talking about here.

> * Initialize `$page` with `$paginationControl->getCurrentPageNumber()`

Isn't that what I'm doing [here](https://github.com/Icinga/icingadb-web/pull/254/files#diff-6962f30e60df477292c92d47f3fe51db2ba4f5d87c37628f653c09addc1ae35fL374)? I just have a different name for the variable.

> * Don't apply the limit a second time (PaginationControl::apply() does that first)
> * Don't apply the offset a second time (PaginationControl::apply() does that first)

Sure I have to, because `limit` and the `offset` are computed based on the current `defaultPageSize` or the `limit` param. And when the `defaultPageSize` has changed, you have to apply it again.

>   * Only the limit in case of `$page > 1 && ! $compact` is still relevant

Don't know exactly what you are referring to? I don't have to do anything at all if limit param is given.

Yonas-net:
Not sure if it's due to the new implementation, however, if you have e.g. in `icingadb/notifications` just enough data for two pages with default limit (25), and you press `Load More` from the first page, after a few seconds the first page can't be seen anymore. I.e. there will be only page 2 rendering and you can't scroll up to page 1 or something.

nilmerg (approved):
That's probably due to the autorefresh. The notification overview shouldn't use autorefresh anyway anymore, please open an issue for that.

PR #299, opened 2021-08-27
Support `ViewModeSwitcher` & `sortControl` in `Host/Service` history

### LINKED ISSUE #296: Support `ViewModeSwitcher` & sort Control in `Host/Service history`
It would be nice to have these features supported in `Host/Service History` too.

Yonas-net:
resolves #296

### DIFF
--- a/application/controllers/HostController.php
+++ b/application/controllers/HostController.php
@@ -23,6 +23,7 @@ use Icinga\Module\Icingadb\Widget\Detail\QuickActions;
 use Icinga\Module\Icingadb\Widget\ItemList\HostList;
 use Icinga\Module\Icingadb\Widget\ItemList\HistoryList;
 use Icinga\Module\Icingadb\Widget\ItemList\ServiceList;
+use ipl\Web\Url;
 
 class HostController extends Controller
 {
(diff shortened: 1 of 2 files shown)


application/controllers/HostController.php
```diff
@@ -124,31 +124,38 @@ public function historyAction()
                 'history.object_type = ?' => 'host'
             ]);
 
-        $url = HostLinks::history($this->host);
-        if (! $this->params->has('page') || ($page = (int) $this->params->shift('page')) < 1) {
-            $page = 1;
-        }
+        $before = $this->params->shift('before', time());
+        $urlParams = clone $this->params;
+        $url = HostLinks::history($this->host)->overwriteParams(array_map('rawurldecode', $urlParams->toArray(false)));
```
nilmerg, line 129: Why so complicated? Why not just `setParams()`?

nilmerg (changes requested):
Also, the view mode switcher has no effect.

application/controllers/HostController.php
Yonas-net: Because it is getting the host `name` param passed to it in `HostLinks::history()` and when I do `setParams()` here, then I do override all the params but it also removes the `name` param which is required when clicking `Load More`.

nilmerg: I see. While working on #312 I noticed that when exporting an object's detail to PDF, the URL in the lower right didn't have any parameters. The URL is there to be able to identify later on where this export is coming from and in lists the used filter is perfectly preserved there. The `name` parameter however is stripped and so doesn't show up. Such params in detail views must be preserved for this reason, hence I created #313.

Please do a `Url::fromRequest()->setParams(clone $this->params)` here (and in `ServiceController::historyAction`) instead. That should work once #313 is merged.

nilmerg (changes requested):

nilmerg (approved):

PR #300, opened 2021-08-27
LoadMore: Don't reset page number on view mode change

### LINKED ISSUE #297: LoadMore resets the page number when the view mode is changed
## Describe the bug
LoadMore resets the page number when the view mode is changed In `History` and `Notification` Controller

## Expected behavior
Page counting should continue.

## Screenshots

Screenshot 2021-08-27 at 10 24 30

## Additional context
Url contains the `page` parameter twice with different values. The second parameter should not be added and the value of the first parameter should be increased correctly.

sukhwinder33445:
fix #297

### DIFF
--- a/library/Icingadb/Common/LoadMore.php
+++ b/library/Icingadb/Common/LoadMore.php
@@ -77,18 +77,8 @@ trait LoadMore
         $count = 0;
         $pageNumber = $this->pageNumber ?: 1;
 
-        if ($this->loadMoreUrl !== null) {
-            $showMore = (new ShowMore(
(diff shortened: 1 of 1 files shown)


nilmerg:
You've just moved code around. Please explain why this works and why it didn't before.

sukhwinder33445:
The for loop also increments the page number so that after creating all the elements for a page, if there are more elements, the `Load More` button will always have the page number incremented by the loop.

nilmerg (approved):
I see! Thanks.

PR #303, opened 2021-08-27
Configuration: Provide custom navigation items

### LINKED ISSUE #293: Having no access to the monitoring module breaks the detail views
## Describe the bug
Visiting a service's detail you'll get this: (a similar error occurs on the host detail)

```
Invalid navigation item type service-action provided

#0 /icingaweb2/library/Icinga/Application/Config.php(445): Icinga\Application\Config::getNavigationConfigPath(String, NULL)
#1 /icingaweb2/library/Icinga/Application/Web.php(243): Icinga\Application\Config::navigation(String)
#2 /icingaweb2/library/Icinga/Web/Navigation/Navigation.php(437): Icinga\Application\Web->getSharedNavigation(String)
#3 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Widget/Detail/ObjectDetail.php(74): Icinga\Web\Navigation\Navigation->load(String)
#4 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Widget/Detail/ServiceDetail.php(15): Icinga\Module\Icingadb\Widget\Detail\ObjectDetail->createActions()
#5 /usr/local/src/ipl-html/src/HtmlDocument.php(295): Icinga\Module\Icingadb\Widget\Detail\ServiceDetail->assemble()
#6 /usr/local/src/ipl-html/src/HtmlDocument.php(496): ipl\Html\HtmlDocument->ensureAssembled()
#7 /usr/local/src/ipl-html/src/HtmlDocument.php(339): ipl\Html\HtmlDocument->render()
#8 /usr/local/src/ipl-html/src/BaseHtmlElement.php(255): ipl\Html\HtmlDocument->renderUnwrapped()
#9 /usr/local/src/ipl-html/src/BaseHtmlElement.php(323): ipl\Html\BaseHtmlElement->renderContent()
#10 /usr/local/src/ipl-html/src/HtmlDocument.php(498): ipl\Html\BaseHtmlElement->renderUnwrapped()
#11 /usr/local/src/ipl-html/src/HtmlDocument.php(339): ipl\Html\HtmlDocument->render()
#12 /usr/local/src/ipl-html/src/HtmlDocument.php(498): ipl\Html\HtmlDocument->renderUnwrapped()
#13 /usr/local/src/ipl-web/src/Compat/ViewRenderer.php(56): ipl\Html\HtmlDocument->render()
#14 /icingaweb2/library/vendor/Zend/Controller/Action/Helper/ViewRenderer.php(964): ipl\Web\Compat\ViewRenderer->render()
#15 /icingaweb2/library/vendor/Zend/Controller/Action/HelperBroker.php(272): Zend_Controller_Action_Helper_ViewRenderer->postDispatch()
#16 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Web/Controller.php(551): Zend_Controller_Action_HelperBroker->notifyPostDispatch()
#17 /icingaweb2/library/Icinga/Web/Controller/Dispatcher.php(76): Icinga\Module\Icingadb\Web\Controller->dispatch(String)
#18 /icingaweb2/library/vendor/Zend/Controller/Front.php(937): Icinga\Web\Controller\Dispatcher->dispatch(Object(Icinga\Web\Request), Object(Icinga\Module\Partnerportal\Web\Response))
#19 /icingaweb2/library/Icinga/Application/Web.php(304): Zend_Controller_Front->dispatch(Object(Icinga\Web\Request), Object(Icinga\Module\Partnerportal\Web\Response))
#20 /icingaweb2/library/Icinga/Application/webrouter.php(107): Icinga\Application\Web->dispatch()
#21 /icingaweb2/public/index.php(4): require_once(String)
#22 {main}
```

yhabteab:
resolves #293

Haven't explicitly checked whether we don't have access to monitoring module, however it should also work without.

### DIFF
--- a/library/Icingadb/Model/Host.php
+++ b/library/Icingadb/Model/Host.php
@@ -4,17 +4,24 @@
 
 namespace Icinga\Module\Icingadb\Model;
 
+use Icinga\Module\Icingadb\Common\Auth;
 use Icinga\Module\Icingadb\Model\Behavior\BoolCast;
 use Icinga\Module\Icingadb\Model\Behavior\ReRoute;
 use ipl\Orm\Behaviors;
 use ipl\Orm\Model;
+use ipl\Orm\Query;
 use ipl\Orm\Relations;
+use ipl\Orm\ResultSet;
 
 /**
  * Host model.
  */
 class Host extends Model
 {
+    use Auth;
+
+    protected $accessorsAndMutatorsEnabled = true;
+
     public function getTableName()
     {
         return 'host';
@@ -136,6 +143,27 @@ class Host extends Model
         $behaviors->add(new ReRoute([
             'servicegroup'  => 'service.servicegroup'
         ]));
+    }
+
+    /**
+     * Mutates flattened custom vars to an associative array
+     *
+     * @param  Query|ResultSet $_
+     *
+     * @return array
+     */
+    public function mutateVarsProperty($_): array
+    {
+        if (! $this->customvar_flat instanceof ResultSet) {
+            $this->applyRestrictions($this->customvar_flat);
+        }
+
+        $vars = [];
+        foreach ($this->customvar_flat as $customVar) {
+            $vars[$customVar->flatname] = $customVar->flatvalue;
+        }
+
+        return $vars;
     }
 
     public function createRelations(Relations $relations)
--- a/library/Icingadb/Model/Service.php
+++ b/library/Icingadb/Model/Service.php
@@ -4,13 +4,20 @@
 
 namespace Icinga\Module\Icingadb\Model;
 
+use Icinga\Module\Icingadb\Common\Auth;
 use Icinga\Module\Icingadb\Model\Behavior\BoolCast;
 use ipl\Orm\Behaviors;
 use ipl\Orm\Model;
+use ipl\Orm\Query;
 use ipl\Orm\Relations;
+use ipl\Orm\ResultSet;
 
 class Service extends Model
 {
+    use Auth;
+
+    protected $accessorsAndMutatorsEnabled = true;
+
     public function getTableName()
     {
         return 'service';
@@ -122,6 +129,27 @@ class Service extends Model
             'notifications_enabled',
             'flapping_enabled'
         ]));
+    }
+
+    /**
+     * Mutates flattened custom vars to an associative array
+     *
+     * @param  Query|ResultSet $_
+     *
+     * @return array
+     */
+    public function mutateVarsProperty($_): array
+    {
+        if (! $this->customvar_flat instanceof ResultSet) {
+            $this->applyRestrictions($this->customvar_flat);
+        }
+
+        $vars = [];
+        foreach ($this->customvar_flat as $customVar) {
+            $vars[$customVar->flatname] = $customVar->flatvalue;
+        }
+
+        return $vars;
     }
 
     public function createRelations(Relations $relations)


nilmerg:
That's the correct approach, but you're still missing an important part: When providing navigation items, there *may* be also dedicated config forms for them. Monitoring has such for actions, if we provide these items on our own, we also have to provide config forms.

These can be a near-copy of those from Monitoring. But please implement the column verification by using `Icinga\Module\Icingadb\Common\Auth::parseRestriction()`.

nilmerg:
I'm afraid but there's still something missing that even I forgot. :smile: There are also custom items required `Icinga\Module\Icingadb\Web\Navigation\` (`Icinga\Module\Monitoring\Web\Navigation`).

Also please rebase with master to the changes from #316. You need them to be able to fully disable the monitoring module, which you should disable in order to test your changes.

application/forms/navigation/ActionForm.php
```diff
@@ -0,0 +1,64 @@
...
+class ActionForm extends NavigationItemForm
+{
+    use Auth;
+
+    /**
+     * The name of the restriction to which the filter should be applied
+     *
+     * @var string
+     */
+    protected $restrictions;
```
nilmerg, line 21: Please use singular here.

application/forms/navigation/ActionForm.php
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public function isValid($formData)
+    {
+        if (! parent::isValid($formData)) {
+            return false;
+        }
+
+        if (($filterString = $this->getValue('filter')) !== null) {
+            $filter = Filter::all();
```
nilmerg, line 51: A filter object isn't required. Just call `parseRestriction`.

nilmerg (changes requested):
Also please rename `application/forms/navigation` to `application/forms/Navigation`, otherwise the forms are not found on case-sensitive filesystems such as mine.

Yonas-net:
refs https://github.com/Icinga/icingaweb2/pull/4526

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -494,4 +498,36 @@ protected function getUsersAndUsergroups()
...
+    protected function getNotesUrls(): array
+    {
+        if ($this->object->notes_url) {
+            //NOTE: notes_url of $this->object has a relation to the `NotesUrl` class
+            //      that's why we access it in such a nested way
+            return $this->object->notes_url->notes_url;
+        }
+
+        return [];
+    }
```
nilmerg, line 532: We require PHP 7 now for Icinga DB Web. So it's possible to use the new null coalescing operator (`??`). This allows the following:

```
foreach ($this->object->action_url->action_url ?? [] as $url) {
    ...
}
```

library/Icingadb/Web/Navigation/Action.php
```diff
@@ -0,0 +1,140 @@
...
+     * Whether this action's macros were already resolved
+     *
+     * @var bool
+     */
+    protected $resolved = false;
+
+    /**
+     * This action's object
+     *
+     * @var CompatHost|CompatService
```
nilmerg, line 29: It's actually the `Host` or `Service` model, I hope?

library/Icingadb/Web/Navigation/Action.php
```diff
@@ -0,0 +1,140 @@
...
+        } else {
+            parent::setUrl($url);
+        }
+
+        return $this;
+    }
+
+    /**
+     * @inheritDoc
+     */
```
nilmerg, line 115: Please don't use this. Only use it if you combine it with custom documentation.

This is fine:

```
/**
* Returns  my super custom type
*
* @inheritdoc
*
* @return SuperCustomType
* /
```

library/Icingadb/Web/Navigation/Action.php
```diff
@@ -0,0 +1,140 @@
...
+    public function getRender()
+    {
+        if ($this->render === null) {
+            $filter = $this->getFilter();
+            $this->render = !$filter || Filter::match(QueryString::fromString($filter)->parse(), $this->getObject());
```
nilmerg, line 135: Space after `!` ;)

library/Icingadb/Model/Behavior/ActionAndNoteUrl.php
```diff
@@ -0,0 +1,60 @@
...
+use Icinga\Module\Icingadb\Common\Database;
+use ipl\Orm\Contract\PropertyBehavior;
+
+class ActionAndNoteUrl extends PropertyBehavior
+{
+    use Database;
+
+    /**
+     * @inheritDoc
+     */
```
nilmerg, line 16: Remove that.

library/Icingadb/Model/Behavior/ActionAndNoteUrl.php
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        $value->action_url = $links;
+
+        return $value;
+    }
+
+    /**
+     * @inheritDoc
+     */
```
nilmerg, line 41: Remove that.

library/Icingadb/Model/Behavior/ActionAndNoteUrl.php
```diff
@@ -0,0 +1,60 @@
...
+        if (empty($value)) {
+            return $value;
+        }
+
+        $links = '';
+        foreach ($value->action_url as $url) {
+            if (! empty($links)) {
+                $links .= ' ';
+            }
+            $links .= $this->getDb()->quoteIdentifier($url);
```
nilmerg, line 53: You don't need to quote here. This is done automatically.

Yonas-net:
**Test**:

Add this to the host/service definition.
```
action_url = "'icingadb/host?name=$host.name$' 'icingadb/host/services?name=$HOSTNAME$'"
```

Bildschirmfoto 2021-09-02 um 10 25 53

library/Icingadb/Common/Macro.php
```diff
@@ -0,0 +1,75 @@
...
+    protected $icingaMacros = [
+        'HOSTNAME'            => 'host_name',
+        'HOSTADDRESS'         => 'host_address',
+        'HOSTADDRESS6'        => 'host_address6',
+        'SERVICEDESC'         => 'service_description',
+        'host.name'           => 'host_name',
+        'host.address'        => 'host_address',
+        'host.address6'       => 'host_address6',
+        'service.description' => 'service_description',
+        'service.name'        => 'service_description'
```
nilmerg, line 27: Our native column expressions are separated by dots. There's no `host_name`, just a `host.name`. We should not use different identifiers.

nilmerg (changes requested):
* The `ActionAndNoteUrl` behavior should be applied on the respective models instead. Otherwise any other columns of the `action_url` and `notes_url` tables are inaccessible from a host or service.

library/Icingadb/Common/Macro.php
Yonas-net: The expressions the user has to specify are still separated by dots, but the actual `LegacyColumns` are not separated in any way by dot. See [here](https://github.com/Icinga/icingadb-web/blob/master/library/Icingadb/Compat/CompatHost.php#L13).

Yonas-net: Never mind.

Yonas-net: We will then have to adjust the `CompatObject`'s `__get()` method because it only matches an e.g. `service|host_propertie`.

Yonas-net:
Just tested it, [this filter](https://github.com/Icinga/icingadb-web/pull/303/files#diff-05efd34d398e560be614df0752da1d77c9d38a364a2be7795d2230eeb53a2e7bR129) will never be `true`, as the `$rule->column` is always passed e.g. like this `host.name`, instead of trimming the object type up to the delimiter.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,105 @@
...
+    /**
+     * Known icinga macros
+     *
+     * @var string[]
+     */
+    protected $icingaMacros = [
+        'HOSTNAME'            => 'name',
+        'HOSTADDRESS'         => 'address',
+        'HOSTADDRESS6'        => 'address6',
+        'SERVICEDESC'         => 'description',
```
nilmerg, line 22: `description` is not a valid column.

nilmerg (changes requested):
Defining a filter for a custom nav item will not show the item as it cannot be evaluated.

The filter match implementation only supports first level columns without table prefix. The filter validation though only accepts prefixed columns (e.g. `host.name`) and may also accept `hostgroup.name`.

But I wouldn't restrict the filter to first level columns only. We support now the linkage between permissions and restrictions and evaluate roles on database level. I believe we can also support full restrictions for action urls the same way.

So to solve this, please perform a sub query inside `Action::getRender()` with which the filter is evaluated instead. Take a look at `ObjectAuthorization::grantsOn()` to get an idea how to do this.

library/Icingadb/Model/Behavior/ActionAndNoteUrl.php
```diff
@@ -0,0 +1,47 @@
...
+        if (empty($value)) {
+            return $value;
+        }
+
+        $links = '';
+        foreach ($value as $url) {
+            if (! empty($links)) {
+                $links .= ' ';
+            }
+            $links .= $url;
```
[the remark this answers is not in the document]
Yonas-net: See https://github.com/Icinga/icingadb-web/pull/303#discussion_r700866495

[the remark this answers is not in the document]
nilmerg: Quoting because it's identified as identifier by the database otherwise is one thing. The quoting I'm talking of above is the format used to designate multiple urls in the configuration option for `action_url` and `notes_url`. Take a look at the parsing algorithm.

library/Icingadb/Authentication/ObjectAuthorization.php
```diff
@@ -208,4 +216,37 @@ protected function checkGrants($permission, $roles)
...
+     *
+     * @return bool
+     */
+    public static function matchesOn($restriction, Model $object)
+    {
+        $self = new static();
+
+        $tableName = $object->getTableName();
+        $uniqueId = $object->{$object->getKeyName()};
+        if (! isset(self::$matchedFilters[$tableName][$uniqueId])) {
```
nilmerg, line 234: Don't use the table name, use the restriction. There may be different filters for the same object type.

library/Icingadb/Authentication/ObjectAuthorization.php
```diff
@@ -208,4 +216,37 @@ protected function checkGrants($permission, $roles)
...
+            $restriction = QueryString::fromString($restriction)->parse();
```
nilmerg, line 235: Why not `$this->parseRestriction()`? If because you don't have the restriction name, identify that based on the object type.

library/Icingadb/Common/Auth.php
```diff
@@ -85,6 +85,19 @@ public function isGrantedOnType($permission, $type, Filter\Rule $filter, $cache
...
 
+    /**
+     * Check whether the filter matches the given object
+     *
+     * @param string $restriction
+     * @param Model $object
+     *
+     * @return bool
+     */
+    public function isMatchedOn($restriction, Model $object)
```
nilmerg, line 96: `isMatchedBy` sounds better, doesn't it?

library/Icingadb/Authentication/ObjectAuthorization.php
```diff
@@ -208,4 +216,37 @@ protected function checkGrants($permission, $roles)
...
+
+    /**
+     * Check whether the given filter matches the given object
+     *
+     * @param string $restriction
+     * @param Model  $object
+     *
+     * @return bool
+     */
+    public static function matchesOn($restriction, Model $object)
```
nilmerg, line 228: Please place this below `grantsOnType`

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,97 @@
...
+                    $input = str_replace($matches[0][$key], $newValue, $input);
+                }
+            }
+        }
+
+        return $input;
+    }
+
+    /**
+     * Resolve a macro based on $this->object being rendered
```
nilmerg, line 51: `$this->object`?

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @param string       $macro The macro to resolve
+     * @param Host|Service $object The host or service used to resolve the macros
+     *
+     * @return string
+     */
+    public function resolveMacro($macro, $object)
+    {
+        if (! array_key_exists($macro, $this->icingaMacros) && ! ($macro = $this->validateMacro($macro))) {
```
nilmerg, line 60: That doesn't make sense. If the macro is unknown, why should it be validated? Why should an unknown macro only be ignored if invalid?

I don't think that entire check is required. The next one should suffice and does in the legacy implementation.

library/Icingadb/Model/Behavior/ActionAndNoteUrl.php
```diff
@@ -0,0 +1,47 @@
...
+        if (empty($value) || ! is_array($value)) {
+            return $value;
+        }
+
+        $links = '';
+        foreach ($value as $url) {
+            if (! empty($links)) {
+                $links .= ' ';
+            }
+            $links .= sprintf("'%s'", $url);
```
nilmerg, line 42: You might also use `"'$url'"` here.

nilmerg (changes requested):

library/Icingadb/Authentication/ObjectAuthorization.php
Yonas-net: Because they are validated anyway when creating the navigation itms. I don't have to parse them again.

library/Icingadb/Common/Auth.php
Yonas-net: Or just simply `matches`?

library/Icingadb/Authentication/ObjectAuthorization.php
nilmerg: There may have something changed the filter besides the form. Don't trust that the user doesn't bypass the form validation somehow.

library/Icingadb/Common/Auth.php
nilmerg: I like the attempt to maintain the `is` naming scheme.

library/Icingadb/Common/Macros.php
Yonas-net: What do you mean when you say `unknown macros`? When the macro couldn’t be found in `$icingaMacros`, it doesn't necessarily mean that the macro is unknown. I am checking here first whether the passed macro exists in this Known macros variable and if not I validate it for correctness, i.e. that it is only dot delimited and then the `validateMacro() ` function returns the actual property. And if none of these conditions is true, then I just return null, because I already logged the occurred error in `validateMacro()`.

library/Icingadb/Common/Auth.php
Yonas-net: Well then, I'd say `matchesTo` would be better.

nilmerg: Oh you're phrasing the method name based on the order of the parameters? With that in mind, yes, `isMatchedBy` doesn't feel correct. But `isMatchedTo` does neither, so let's keep `isMatchedOn`. :laughing:

library/Icingadb/Common/Macros.php
nilmerg: But why the check for the dot separation? It's not even possible to resolve chained properties with the `$obj->{$dottedString}` syntax.

Yonas-net: > It's not even possible to resolve chained properties with the `$obj->{$dottedString}` syntax.

That's why I'm also splitting up these chained properties here.
https://github.com/Icinga/icingadb-web/blob/0c4c8a2c210a8023a229b88c4058828b601810e8/library/Icingadb/Common/Macros.php#L89-L91

nilmerg: I see. But that also matches `host.state.last_comment.author` and tries to resolve `state.last_comment.author`. The thing is, the legacy implementation didn't support this either, and I don't think we should open this up (yet).

This also requires from the user to either add `host.` or `service.` to the macro, which is of no use anyway because the implementation doesn't care that `host.name` is really only resolved on host objects. And if I can use `/wiki/page?host=$host.name$&service=$service.name$` I'd expect to get two different results. If that's not possible, we shouldn't require any dots.

library/Icingadb/Model/Behavior/ActionAndNoteUrl.php
```diff
@@ -0,0 +1,51 @@
...
+
+        if (count($value) === 1) {
+            return $value[0];
+        }
+
+        $links = '';
+        foreach ($value as $url) {
+            if (! empty($links)) {
+                $links .= ' ';
+            }
```
nilmerg, line 45: empty line, please.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,102 @@
...
+     * @var string[]
+     */
+    protected $icingaMacros = [
+        'HOSTNAME'      => 'name',
+        'HOSTADDRESS'   => 'address',
+        'HOSTADDRESS6'  => 'address6',
+        'host.name'     => 'name',
+        'host.address'  => 'address',
+        'host.address6' => 'address6',
+        'service.name'  => 'name'
```
nilmerg, line 25: I believe we shouldn't support the all uppercase macros still. These are from Icinga 1.x days and now it's the time to lay them to rest. Anyone running Monitoring in parallel to Icinga DB Web can still use the dot separated macros.

In addition to my comment [here](https://github.com/Icinga/icingadb-web/pull/303#discussion_r703316085) we also have to still allow the differentiation between `host.name` and `service.name`. I hadn't in mind that these were always dot separated macros and only used them as examples. But anyone currently using those together in Monitoring, gets two different results. We also have to support that. So please make this possible, but only for `service.name`. (Consider to just define a path `['service', 'name']` as macro expression and handle arrays differently then)

nilmerg (changes requested):

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,88 @@
...
+        if (
+            isset($this->icingaMacros[$macro])
+            && isset($object->{$this->icingaMacros[$macro]})
+            && $macroType === $objectClass
+        ) {
+            return $object->{$this->icingaMacros[$macro]};
+        }
+
+        try {
+            if (array_key_exists($macro, $this->icingaMacros)) {
```
nilmerg, line 73: Why not `isset($this->icingaMacros[$macro])` as above?

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,88 @@
...
+                // The macro does exist in the $this->icingaMacros though its object type is different
+                $value = array_reduce(explode('.', $macro), function ($first, $second) {
```
nilmerg, line 75: Don't use `array_reduce`, that's overkill. Use something like this `$object->$type->$column` or normalize it first `$object = $object->$type`.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Resolve a macro based on the given object
+     *
+     * @param string       $macro  The macro to resolve
+     * @param Host|Service $object The host or service used to resolve the macros
+     *
+     * @return string
+     */
+    public function resolveMacro($macro, $object)
```
nilmerg, line 55: It's still not strict enough. `service.name` will currently be evaluated also if `$object` is a host. Though services are not joined for hosts and even then, it'd be a list of services not just a single one. `service.name` is the **only** macro that is supported with this prefix and can only be evaluated on services to differentiate it from `host.name` (`$service->host` exists and is then targeted)

nilmerg (changes requested):

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,87 @@
...
+            && isset($object->{$this->icingaMacros[$macro]})
+            && $macroType === $objectType
+        ) {
+            return $object->{$this->icingaMacros[$macro]};
+        }
+
+        try {
+            if (isset($this->icingaMacros[$macro]) && $objectType === 'service') {
+                // The macro does exist in the $this->icingaMacros though its type is different
+                $column = substr($macro, strpos($macro, '.') + 1);
```
nilmerg, line 75: I guess `$this->icingaMacros[$macro]` should suffice here. This has two advantages:

* It's less work
* It's still correct even if the macro name is something entirely different than the expression

nilmerg (changes requested):

nilmerg (approved):
:+1:

nilmerg (changes requested):
Sorry, was a bit fast with approval here. There are still some integration issues:

1. If having both modules enabled (monitoring and icingadb), monitoring shows an error in detail views with actions configured
2. We don't provide different nav-item identifiers, so users with existing configs may already have custom items configured with filters that only the monitoring module understands. This also leads to an error but on our side

To solve this we need to:

* avoid providing our actions if the monitoring module is enabled
* handle both cases in our detail views (`Navigation::load()` may load either our actions or the ones from monitoring). i.e. an `instanceof` check in `ObjectDetail::createActions()`
* be able to parse also the legacy filters in our `Action`. (`UrlMigrator::transformFilter()` will be of help here)

This should lead to the following:

* If monitoring is enabled, it's actions have preference
  * icingadb-web will also understand and evaluate the legacy filters, but still won't understand all legacy macros
  * monitoring and icingadb detail views work fine and show both the same actions
* If monitoring is disabled
  * icingadb-web will also understand and evaluate the legacy filters, but still won't understand all legacy macros
  * icingadb detail views still show all actions, including those created for monitoring

configuration.php
```diff
@@ -259,6 +259,14 @@
...
             150
         );
+    } else {
+        /*
+        * Available navigation items
+        */
+        $this->provideNavigationItem('icingadb-host-action', $this->translate('Host Action (Icinga DB)'));
+        $this->provideNavigationItem('icingadb-service-action', $this->translate(
+            'Service Action (Icinga DB)'
+        ));
```
nilmerg, line 269: Please use e.g. `$this->translate('Host Action') . ' (Icinga DB)'`. The latter doesn't need localization and this avoids duplicates in the message catalogs.

nilmerg (changes requested):

configuration.php
```diff
@@ -253,6 +259,18 @@
             . '&view=minimal&limit=15&sort=host.state.severity desc',
             150
         );
+    } else {
+        /*
+        * Available navigation items
+        */
+        $this->provideNavigationItem(
+            'icingadb-host-action',
+            $this->translate('Host Action') . '(Icinga DB)'
```
nilmerg, line 268: There's a space missing.

nilmerg (changes requested):

nilmerg (changes requested):
Works fine now with monitoring and without. :+1: 

Though I noticed that the macro support is still limited compared to what monitoring provided: Monitoring also supported to resolve some relational data such as custom variables by use of custom macros. Also, custom macros were allowed to contain prefixes (`host_` or `service_`).

So please add support for both additional forms:

* Actions should allow this: `$<type>.<column>$`, `$<column>$`, `<type>.vars.<var>`, `vars.<var>`
* Previous limitations which remain: `$service.<column>` or `$service.vars.<var>` can only be resolved on services

Technical hints:
* The `ObjectDetail` already fetches customvars in `createCustomVars()`. Outsource this to an internal method such as `fetchCustomVars()` and use that there. Within `fetchCustomVars()`, populate the object's `customvar_flat` property with the `ResultSet` from the query.
* Then use property `customvar_flat` in `createCustomVars()` for the call to `unflattenVars`.
* `createActions()` also calls `fetchCustomVars()`
* In the `Host` and `Service` models, add a new method `mutateVarsProperty()`
  * This transforms `customvar_flat` to an associative array where cv names are keys and cv values the values.
  * If `customvar_flat` is not a `ResultSet`, apply restrictions to it (the query) beforehand
 * Eventually, `Macros` should be able to access `$object->vars` and `$service->host->vars` to quickly resolve custom variable macros
 
Performance goal: Only one query to the object's customvars and only one to a host's ones if it's a service. The latter is performed lazily just because `Macros` accesses `$service->host->vars` once.

Yonas-net:
> Monitoring also supported to resolve some relational data such as custom variables by use of custom macros. Also, custom macros were allowed to contain prefixes (`host_` or `service_`).

Monitoring module has never supported custom vars macro though, it's actually pretty cool to support this here as we don't have to do it exactly the way monitoring does it. Have tested the following and monitoring couldn't resolve it.

```
monitoring/host/show?host=$host.name$&$_host_vars.os$
monitoring/host/show?host=$host.name$&$host_vars.os$
monitoring/host/show?host=$host.name$&$_host_vars_os$
```

nilmerg:
`monitoring/host/show?host=$host.name$&$_host_os$` should work in monitoring.

library/Icingadb/Model/Service.php
```diff
@@ -124,6 +132,31 @@ public function createBehaviors(Behaviors $behaviors)
         ]));
     }
 
+    /**
+     * Mutates flattened custom vars to an associative array
+     *
+     * @param $factory
```
nilmerg, line 138: Please add a type hint here (`@param Closure|ResultSet $factory`)

library/Icingadb/Model/Service.php
```diff
@@ -124,6 +132,31 @@ public function createBehaviors(Behaviors $behaviors)
...
+            $query = $factory($this);
+            $this->applyRestrictions($query);
+        } else {
+            $query = $factory;
+        }
+
+        $vars = [];
+        foreach ($query as $customVar) {
+            $paths = explode('.', $customVar->flatname);
+            $this->transformVar($vars, $paths, $customVar->flatvalue);
```
nilmerg, line 154: I thought of just using `$vars[$customVar->flatname] = $customVar->flatvalue` suffices. This also makes resolving macros easier, as it's just a single key lookup then.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,143 @@
...
+use Icinga\Module\Icingadb\Model\Service;
+
+trait Macros
+{
+    /**
+     * Known icinga macros
+     *
+     * @var array
+     */
+    protected $icingaMacros = [
```
nilmerg, line 18: Is this really required at all? The supported macros are way more *flexible* any way, so to me this map is rather redundant, isn't it?

nilmerg (changes requested):
I didn't take a look at the actual macro resolver again, because once you've changed the `vars`  mutation, this will change anyway again.

library/Icingadb/Common/Macros.php
Yonas-net: I think so too. It was actually mainly intended to resolve the uppercase macros.

Yonas-net:
Mutating the vars property doesn't work quite right yet. It still fails when a host/service object contains a comment.

```bash
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'sub_comment.id' in 'where clause'

#0 /usr/local/src/ipl-sql/src/Connection.php(401): PDO->prepare(String)
#1 /usr/local/src/ipl-sql/src/Connection.php(418): ipl\Sql\Connection->prepexec(String, Array)
#2 /usr/local/src/ipl-orm/src/Query.php(639): ipl\Sql\Connection->select(Object(ipl\Sql\Select))
#3 /usr/local/src/ipl-orm/src/ResultSet.php(104): ipl\Orm\Query->yieldResults()
#4 [internal function]: ipl\Orm\ResultSet->yieldTraversable(Object(Generator))
#5 /usr/local/src/ipl-orm/src/ResultSet.php(86): Generator->valid()
#6 /usr/local/src/ipl-orm/src/ResultSet.php(78): ipl\Orm\ResultSet->advance()
#7 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Model/Host.php(164): ipl\Orm\ResultSet->rewind()
#8 /usr/local/src/ipl-stdlib/src/Properties.php(126): Icinga\Module\Icingadb\Model\Host->mutateVarsProperty(Object(Closure))
#9 /usr/local/src/ipl-stdlib/src/Properties.php(101): ipl\Orm\Model->mutateProperty(String)
#10 /usr/local/src/ipl-stdlib/src/Properties.php(54): ipl\Orm\Model->setProperty(String, Object(Closure))
#11 /usr/local/src/ipl-orm/src/Hydrator.php(141): ipl\Orm\Model->setProperties(Array)
#12 /usr/local/src/ipl-orm/src/Query.php(646): ipl\Orm\Hydrator->hydrate(Array, Object(Icinga\Module\Icingadb\Model\Comment))
#13 /usr/local/src/ipl-orm/src/ResultSet.php(104): ipl\Orm\Query->yieldResults()
#14 [internal function]: ipl\Orm\ResultSet->yieldTraversable(Object(Generator))
#15 /usr/local/src/ipl-orm/src/ResultSet.php(33): Generator->valid()
#16 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Widget/Detail/ObjectDetail.php(171): ipl\Orm\ResultSet->hasResult()
#17 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Widget/Detail/ServiceDetail.php(17): Icinga\Module\Icingadb\Widget\Detail\ObjectDetail->createComments()
#18 /usr/local/src/ipl-html/src/HtmlDocument.php(295): 
[...]
```

Yonas-net:
As a workaround you need to comment out [these lines](https://github.com/Icinga/ipl-stdlib/blob/master/src/Properties.php#L100-L102) of code.

library/Icingadb/Model/Service.php
```diff
@@ -124,6 +130,30 @@ public function createBehaviors(Behaviors $behaviors)
...
+     * Mutates flattened custom vars to an associative array
+     *
+     * @param  \Closure|ResultSet $factory
+     *
+     * @return array
+     */
+    public function mutateVarsProperty($factory)
+    {
+        if (! $factory instanceof ResultSet) {
+            $query = $factory($this);
```
nilmerg, line 143: This isn't necessary anymore with https://github.com/Icinga/ipl-stdlib/pull/24.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,120 @@
...
+        if ($object instanceof Host) {
+            $objectType = 'host';
+        } else {
+            $objectType = 'service';
+        }
+
+        $hasVarsPrefix = false;
+        if (preg_match('/^(host|service)[.](vars)[.](.+)/i', $macro, $matches)) {
+            $macroType = $matches[1];
+            $hasVarsPrefix = true;
```
nilmerg, line 61: I don't like that you access `$matches` later on again. You're already using a flag here, why don't you store the variable path here and use that as the flag?

The second group `vars` by the way isn't necessary. Don't group it here and use a hardcoded value later on to access the `vars` property.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,120 @@
...
+        } elseif (preg_match('/^(host|service)[.](.+)/i', $macro, $matches)) {
+            $macroType = $matches[1];
+            $properties = explode('.', $matches[2]);
```
nilmerg, line 64: This isn't initialized. I guess that's why the second condition below is so complicated? `($macroType !== $objectType && $objectType === 'service') || $macroType === $objectType`

Initialize it with `null` and check for this instead. Way easier to read and more efficient.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,120 @@
...
+        } else {
+            $macroType = $objectType;
+        }
+
+        try {
+            if ($hasVarsPrefix || preg_match('/(^)(vars)[.](.+)/i', $macro, $matches)) {// Custom vars
```
nilmerg, line 70: The preg_match here should not be required. Adjust your first preg_match so that the object type is optional and detect the macro type then instantly instead of later on again with a complex condition. At least I recognize the third `if` here as that.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,120 @@
...
+use Icinga\Module\Icingadb\Model\Service;
+
+trait Macros
+{
+    /**
+     * Caches already processed custom vars query for the given object
+     *
+     * @var array
+     */
+    protected static $loadedCustomVars = [];
```
nilmerg, line 18: Am I right that you've added this because I mentioned a *performance goal* above? This shouldn't be required for this. The mutation already runs only once, or is at least supposed to, and this already ensures the variables aren't queried more than once.

nilmerg (changes requested):

library/Icingadb/Common/Macros.php
Yonas-net: > This isn't initialized.

What is not initialized here? I don't quite get what you mean.

nilmerg: `$properties`. This variable is set inside the elseif branch but used in an unrelated way later on. PHPStorm should also indicate this with a hint.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,93 @@
...
+                } else {
+                    $value = $object;
+                }
+
+                if ($isCustomVar) {
+                    $value = $value->vars[$path];
+                } else {
+                    do {
+                        $column = array_shift($properties);
+                        $value = $value->$column;
```
nilmerg, line 80: This allows to access `host.hostgroup` for example. That's bad since it cannot be resolved to a useful value and might issue a query. Don't resolve such macros that result in an access to properties that have `Query`, `ResultSet` or `array` instance as value.

nilmerg (changes requested):
Remember:
> * Actions should allow this: `$<type>.<column>$`, `$<column>$`, `<type>.vars.<var>`, `vars.<var>`
> * Previous limitations which remain: `$service.<column>` or `$service.vars.<var>` can only be resolved on services

I miss `$<column>$` at the moment.

Yonas-net:
> I miss `$<column>$` at the moment.

https://github.com/Icinga/icingadb-web/blob/294565521db5c5961c67b119db03cec1d4a57263/library/Icingadb/Common/Macros.php#L83-L85

nilmerg:
Oh oops, I'm blind, obviously. :see_no_evil:

library/Icingadb/Model/Host.php
```diff
@@ -138,6 +145,27 @@ public function createBehaviors(Behaviors $behaviors)
...
+    /**
+     * Mutates flattened custom vars to an associative array
+     *
+     * @param  Query|ResultSet $queryOrResult
+     *
+     * @return array
+     */
+    public function mutateVarsProperty($queryOrResult): array
+    {
+        if (! $queryOrResult instanceof ResultSet) {
```
nilmerg, line 157: I noticed that this receives always a query. Remember this?:

> * In the Host and Service models, add a new method mutateVarsProperty()
>   * This transforms customvar_flat to an associative array where cv names are keys and cv values the values.
>   * If customvar_flat is not a ResultSet, apply restrictions to it (the query) beforehand

You have to access `$this->customvar_flat` instead of `$queryOrResult`.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,103 @@
...
+{
+    /**
+     * Get the given string with macros being resolved
+     *
+     * @param string       $input  The string in which to look for macros
+     * @param Host|Service $object The host or service used to resolve the macros
+     *
+     * @return string
+     */
+    public function expandMacros(string $input, $object): string
```
nilmerg, line 23: Add `Model` as type declaration for `$object`.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Resolve a macro based on the given object
+     *
+     * @param string       $macro  The macro to resolve
+     * @param Host|Service $object The host or service used to resolve the macros
+     *
+     * @return string
+     */
+    public function resolveMacro(string $macro, $object): string
```
nilmerg, line 45: Add `Model` as type declaration for `$object`.

library/Icingadb/Common/Macros.php
```diff
@@ -0,0 +1,103 @@
...
+    {
+        if ($object instanceof Host) {
+            $objectType = 'host';
+        } else {
+            $objectType = 'service';
+        }
+
+        $macroType = null;
+        $isCustomVar = false;
+        if (preg_match('/^((host|service)[.])?vars[.](.+)/i', $macro, $matches)) {
```
nilmerg, line 55: Why `[.]` and not `\.`?

And please don't make it case-insensitive. We process `host`, `service` or `vars` case-sensitive everywhere else. We shouldn't accept e.g. `HOST` here either.

library/Icingadb/Common/Macros.php
nilmerg: This isn't resolved yet.

nilmerg (changes requested):

nilmerg (approved):
Finally. :tada:  If there's still anything wrong, we can fix it later on, but I didn't find anything. :+1:

PR #307, opened 2021-08-30
Add `No items found` to controllers

### LINKED ISSUE #286: Search result: `No items found` should have a better design
## Describe the bug
There is an empty status div with `No items found` if there is no search result.

Screenshot 2021-08-25 at 17 22 47

But all the controllers after `Notifications` are missing it:

Screenshot 2021-08-25 at 17 21 35

The` history` controller shows nothing if no item is found.

sukhwinder33445:
fix #286

### DIFF
--- a/application/views/scripts/services/grid-flipped.phtml
+++ b/application/views/scripts/services/grid-flipped.phtml
@@ -8,7 +8,9 @@ if (! $this->compact): ?>
 <?php endif ?>
 <div class="content" data-base-target="_next">
     <?php if (empty($pivotData)): ?>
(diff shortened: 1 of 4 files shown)


sukhwinder33445:
Should I add the `empty-state` to` service grid` controller and `list/services` as well?

nilmerg:
Service grid, yes, but what is `list/services`?

sukhwinder33445:
Sorry, `list/services.phtml` is from `monitoring` module.

nilmerg (approved):

Issue #279, opened 2021-08-25
User detail empty values


sukhwinder33445:
## Describe the bug
Empty values look bad.
 
There should be a `-` for empty values and a `:` after keys.

## Screenshots
Now:

![Screenshot 2021-08-25 at 11 15 43](https://user-images.githubusercontent.com/54990055/130765129-897bdd74-f5b7-4730-ad2b-384c8a1bcf1d.png)

After:
![Screenshot 2021-08-25 at 11 18 08](https://user-images.githubusercontent.com/54990055/130765132-f15b23b0-bf46-4650-bdc5-47930c9ba741.png)

nilmerg:
> and a : after keys.

We have nowhere a colon after keys in such presentations.

sukhwinder33445:
The following issues could also be fixed for better readability:

![Screenshot 2021-08-25 at 16 43 47](https://user-images.githubusercontent.com/54990055/130813735-75a3ac36-9e6b-4cea-988d-3d06054e439b.png)

nilmerg:
Whoof, nice find!

Issue #192, opened 2021-04-30
`.item-table`s are still prone to overflow due long subjects


nilmerg:
Fixing this isn't easy. These views use `display:table` and `table-layout:fixed` is no option. `overflow:hidden` doesn't work with them. Refactoring them to not use `display:table` is IMHO the only solution.

Currently affected views:

* Users
* Usergroups
* Hostgroups
* Servicegroups

PR #309, opened 2021-08-30
UserDetail: Enhance user detail view

### LINKED ISSUE #281: Missing `Host/Service` notification `states/types`, `TimePeriods` in user datail
## Describe the bug
The `Host/Service` notification filter `states` and `types`, `TimePeriods` etc. are missing in the user details.

sukhwinder33445:
fix #281 #279

### DIFF
--- a/library/Icingadb/Widget/Detail/UserDetail.php
+++ b/library/Icingadb/Widget/Detail/UserDetail.php
@@ -54,10 +54,18 @@ class UserDetail extends BaseHtmlElement
 
     protected function createUserDetail()
     {
+        list($hostStates, $serviceStates) = $this->separateStates($this->user->states);
+        $hostStates = implode(', ', $this->localizeStates($hostStates));
+        $serviceStates = implode(', ', $this->localizeStates($serviceStates));
+        $types = implode(', ', $this->localizeTypes($this->user->types));
+
         return [
             new HtmlElement('h2', null, Text::create(t('Details'))),
-            new HorizontalKeyValue(t('E-Mail'), $this->user->email),
-            new HorizontalKeyValue(t('Pager'), $this->user->pager)
+            new HorizontalKeyValue(t('E-Mail'), $this->user->email ?: '-'),
+            new HorizontalKeyValue(t('Pager'), $this->user->pager ?: '-'),
+            new HorizontalKeyValue(t('Host States'), $hostStates ?:  '-'),
+            new HorizontalKeyValue(t('Service States'), $serviceStates ?: '-'),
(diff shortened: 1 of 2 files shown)


nilmerg (changes requested):
We should enrich the states and types somehow. How about using a `TagList` for both, a localized label and then link to a filtered notification list with the state/type and user filter applied?

nilmerg (changes requested):
Sorry I have to step back from my suggestion to use a TagList. The filters for the notification list resulting from this are waaaaaay too expensive. We cannot provide such expensive filters by default.

So, please make states and types simple comma separated text again. Also:

* Make them localizable. I'd want to see `Problem`, not `problem`. (And don't dare to use `ucfirst` for this :wink: )
* Show a dash again for no value everywhere. That's how we do it everywhere else and it's sufficient

sukhwinder33445:
> waaaaaay too expensive

You mean in terms of performance? Can you please explain why? it's just doing the same thing as we filter something with two options.

nilmerg:
I have 3.5 million notifications in my history. If I use such a filter, I'll sit there waiting for 4+ minutes to get any result. This is acceptable if the filter is my own. But a link in the UI shouldn't lead to this.

sukhwinder33445:
I see, thank you for the explanation.

library/Icingadb/Widget/Detail/UserDetail.php
```diff
@@ -90,4 +95,68 @@ protected function assemble()
...
+                    $localizedTypes[] = t('Problem');
+                    break;
+                case 'ack':
+                    $localizedTypes[] = t('Acknowledgement');
+                    break;
+                case 'recovery':
+                    $localizedTypes[] = t('Recovery');
+                    break;
+                case 'downtime_start':
+                    $localizedTypes[] = t('DowntimeStart');
```
nilmerg, line 114: It's a label/title, so please use general word boundaries/structure. (i.e. spaces)

nilmerg (changes requested):

nilmerg (dismissed):

sukhwinder33445:
I am working on separating the states for host and services.

nilmerg:
Yep, also don't forget [this](https://github.com/Icinga/icingadb-web/issues/279#issuecomment-905575411). We should remove the mail and pager from list items. The "list" being a "table" isn't good either: #192

nilmerg (changes requested):
* There is no UNKNOWN state for hosts
* You forgot the OK state in `localizeStates`
* `separateStates` cannot get the result of `localizeStates` as input

PR #310, opened 2021-08-31
HealthController: Dis/enabling Fcmds globally doesn't apply changes i…

### LINKED ISSUE #283: Dis/enabling Fcmds globally doesn't apply till manual reloading
## Describe the bug
When you disable/enable the Feature Commands globally, it doesn't have any effect until you manually reload the browser/Tab.

Yonas-net:
…mmediately

resolves #283

### DIFF
--- a/application/controllers/HealthController.php
+++ b/application/controllers/HealthController.php
@@ -101,6 +101,8 @@ class HealthController extends Controller
         ]);
         $toggleInstanceFeaturesCommandForm->setObjects([$instance]);
         $toggleInstanceFeaturesCommandForm->on(ToggleInstanceFeaturesForm::ON_SUCCESS, function () {
+            $this->getResponse()->setAutoRefreshInterval(1);
+
             $this->redirectNow(Url::fromPath('icingadb/health')->getAbsoluteUrl());
         });
         $toggleInstanceFeaturesCommandForm->handleRequest(ServerRequest::fromGlobals());


nilmerg (changes requested):
I was actually thinking to solve this the same as [along these lines](https://github.com/Icinga/icingadb-web/blob/master/library/Icingadb/Common/CommandActions.php#L142-L144). Please do that instead.

But it will still not work for some reason. I believe this is a bug in Icinga Web 2 and I will investigate this later.

application/controllers/HealthController.php
```diff
@@ -99,8 +99,11 @@ public function indexAction()
             ToggleInstanceFeatureCommand::FEATURE_PERFORMANCE_DATA =>
                 $instance->icinga2_performance_data_enabled
         ]);
+
```
nilmerg, line 102: Why that empty line here now? I keep seeing that you do this, is this a personal preference or do you think it's required?

Yonas-net: I can remove it directly if it doesn't suit you. I just wanted it to be a bit clearer. I  was advised to add a blank line after a variable declaration and before calling a function.

nilmerg: > I was advised to add a blank line after a variable declaration and before calling a function.

There's no rule in the PSR12 styleguide for this. If at all this is personal preference and you may do that if you like, but then only on code you've written personally. Not on code you don't need to touch.

I personally don't add an empty line after variable declarations that are primarily used in the control structure following it. I also write instructions to initialize objects without any empty line in between. The basic principle I follow here is that every part of the code is related to a group. A perfect example are the actions in controllers: query initialization, control initialization, searchbar handling, final query adjustments, exporting, control registration, content preparation, content registration, response handling.

nilmerg (approved):

PR #314, opened 2021-08-31
Don't show dropdown menus in `delete/remove` & `details` actions

### LINKED ISSUE #273: Dropdown menu shouldn't be rendered on remove/delete actions
## Describe the bug
The `Add to Dashboard` and `Add to Menu` dropdown menus shouldn't be rendered upon `remove/delete` actions.

Yonas-net:
resolves #273

### DIFF
--- a/application/controllers/CommentController.php
+++ b/application/controllers/CommentController.php
@@ -22,6 +22,7 @@ class CommentController extends Controller
     public function init()
     {
         $this->setTitle(t('Comment'));
+        $this->getTabs()->disableLegacyExtensions();
 
         $name = $this->params->shiftRequired('name');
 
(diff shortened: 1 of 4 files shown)


nilmerg:
When I've first looked at the report, I thought that it makes sense because if there's just a button or a form, that really doesn't need to be in the menu/dashboard or exported as e.g. PDF. As it is with the command transport configuration.

But now seeing what you've changed, I'm afraid but that's not required. The `deleteAction` is just a form action, but won't be shown to the user. The actual detail view (`init`) and the multiselect view (`detailsAction`) are not just for the remove button. In the multiselect view there is of course just the remove button, but that's maybe not always the case. And why should we disallow exporting the comment detail or downtime detail to PDF? And if someone wants to add them to the menu or dashboard, why not?

I really don't think we should do that.

PR #331, opened 2021-09-02
Link preferred view mode with the current view

### LINKED ISSUE #260: Make view mode global per view only
## Describe the bug
When switching the view mode in the `host services`, the switcher in `hosts` list is also switched, however, it has no effect on the list, as the `Paginator` and the `default limit` aren't updated.

nilmerg:
fixes #260


Yonas-net (approved):
Works like a charm.

PR #333, opened 2021-09-02
Don't allow user to delete scheduled downtimes

### LINKED ISSUE #269: Don't allow users to remove scheduled downtime objects
## Describe the bug
Trying to remove a scheduled downtime object shows an error that this isn't possible.

## Expected behavior
If it's not possible to remove the downtime, we should disable the removal button with an indication why it's disabled.

## Additional context
https://github.com/Icinga/icingadb/issues/231

sukhwinder33445:
fix #269

### DIFF
--- a/application/forms/Command/Object/DeleteDowntimeForm.php
+++ b/application/forms/Command/Object/DeleteDowntimeForm.php
@@ -25,11 +25,23 @@ class DeleteDowntimeForm extends CommandForm
 
     protected function assembleSubmitButton()
     {
+        $isDisabled = true;
+        foreach ($this->getObjects() as $downtime) {
+            if ($downtime->scheduled_by === null) {
+                $isDisabled = false;
+                break;
+            }
+        }
+
         $this->addElement(
             'submitButton',
             'btn_submit',
             [
                 'class' => ['cancel-button', 'spinner'],
+                'disabled' => $isDisabled ?: null,
+                'title' => $isDisabled
+                    ? t('Downtime cannot be removed at runtime because it is based on a configured scheduled downtime.')
+                    : null,
                 'label' => [
                     new Icon('trash'),
                     tp('Delete downtime', 'Delete downtimes', count($this->getObjects()))
@@ -40,7 +52,10 @@ class DeleteDowntimeForm extends CommandForm
 
     protected function getCommand(Model $object)
     {
-        if (! $this->isGrantedOn('icingadb/command/downtime/delete', $object->{$object->object_type})) {
+        if (
+            ! $this->isGrantedOn('icingadb/command/downtime/delete', $object->{$object->object_type})
+            || $object->scheduled_by !== null
+        ) {
             return null;
         }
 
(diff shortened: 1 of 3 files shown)


library/Icingadb/Model/Downtime.php
```diff
@@ -76,6 +77,7 @@ public function getMetaData()
             'is_in_effect'          => t('Downtime Is In Effect'),
             'start_time'            => t('Downtime Actual Start'),
             'end_time'              => t('Downtime Actual End'),
+            'scheduled_by'          => t('Downtime Scheduler Name'),
```
nilmerg, line 80: It's not a name. Use `Downtime Scheduled By`.

application/forms/Command/Object/DeleteDowntimeForm.php
```diff
@@ -7,6 +7,8 @@
 use Icinga\Module\Icingadb\Command\Object\DeleteDowntimeCommand;
 use Icinga\Module\Icingadb\Common\Auth;
 use Icinga\Module\Icingadb\Forms\Command\CommandForm;
+use ipl\Html\Attribute;
+use ipl\Html\Attributes;
```
nilmerg, line 11: What are these?

application/forms/Command/Object/DeleteDowntimeForm.php
```diff
@@ -25,11 +27,29 @@ protected function assembleElements()
...
         $this->addElement(
             'submitButton',
             'btn_submit',
             [
-                'class' => ['cancel-button', 'spinner'],
+                'class' => ['cancel-button', 'spinner', $isDisabled ? 'disabled' : null],
+                'disabled' => $isDisabled ?: null,
+                'title' => $isDisabled
+                    ? sprintf(
+                        t('Cannot remove downtime %s as it is owned by scheduled downtime object %s.'),
```
nilmerg, line 48: Don't use such a technical message. (This is directly from the api response isn't it?) Just tell the user that it's based on a configured scheduled downtime and hence cannot be removed at runtime.

application/forms/Command/Object/DeleteDowntimeForm.php
```diff
@@ -25,11 +27,29 @@ protected function assembleElements()
 
     protected function assembleSubmitButton()
     {
+        $isDisabled = false;
+        foreach ($this->getObjects() as $downtime) {
+            if ($downtime->scheduled_by) {
+                $downtimeName = $downtime->name;
+                $scheduledBy = $downtime->scheduled_by;
+                $isDisabled = true;
```
nilmerg, line 35: I don't think it's a good idea to disable the button just because there's one downtime that cannot be removed. Just disable it if *all* downtimes cannot be removed.

Then in `getCommand` only return a command if removal is possible.

nilmerg (changes requested):

nilmerg (changes requested):
Don't use a class to style the disabled button. Use an attribute selector instead.

nilmerg (approved):

PR #337, opened 2021-09-06
Controller: Reset all params on form change

### LINKED ISSUE #318: Clearing the searchbar in the tactical overview doesn't update URL
## Describe the bug
Clearing the filter using the searchbar, the url isn't updated and after the next autorefresh the filter re-appears.

## Expected behavior
The url is updated and the filter doesn't re-appear.

Yonas-net:
The problem here was all the parameters added by the search-bar could never be deleted again as
the tactical overview doesn't provide any preserved params that could reset this filter.
This PR fixes that by simply removing all parameters from the URL being redirected when the trash icon is pressed.

resolves #318

### DIFF
--- a/library/Icingadb/Web/Controller.php
+++ b/library/Icingadb/Web/Controller.php
@@ -154,7 +154,9 @@ class Controller extends CompatController
     public function createSearchBar(Query $query, array $preserveParams = null)
     {
         $requestUrl = Url::fromRequest();
-        $redirectUrl = $preserveParams !== null ? $requestUrl->onlyWith($preserveParams) : $requestUrl;
+        $redirectUrl = $preserveParams !== null
+            ? $requestUrl->onlyWith($preserveParams)
+            : (clone $requestUrl)->setParams([]);
 
         $filter = QueryString::fromString((string) $this->params)
             ->on(QueryString::ON_CONDITION, function (Filter\Condition $condition) use ($query) {


nilmerg:
But that defeats the purpose of preserving parameters.

nilmerg (changes requested):
Changing a term also doesn't clear it but re-adds it. e.g. Changing `host.name=foo` to `host.name=bar` will result in `?host.name=foo&host.name=bar`.

nilmerg (changes requested):
It works fine now. But it's still too complicated. Isn't the problem that the redirect url still contains params even if none need to be preserved? I'd try to solve that instead.

library/Icingadb/Web/Controller.php
```diff
@@ -154,7 +154,7 @@ public function createSortControl(Query $query, array $columns)
     public function createSearchBar(Query $query, array $preserveParams = null)
     {
         $requestUrl = Url::fromRequest();
-        $redirectUrl = $preserveParams !== null ? $requestUrl->onlyWith($preserveParams) : $requestUrl;
+        $redirectUrl = $preserveParams !== null ? $requestUrl->onlyWith($preserveParams) : $requestUrl->setParams([]);
```
nilmerg, line 157: Way better already :smile: But `onlyWith` clones the url first, `setParams` does not. But the cloning should happen because changing the request url implicitly isn't good.

nilmerg (changes requested):

nilmerg (approved):
LGTM :)

Issue #44, opened 2020-04-01
Now bubble overlaps scheduled start bubble of flexible downtimes


nilmerg:
![Bildschirmfoto von 2020-04-01 14-28-29](https://user-images.githubusercontent.com/16668527/78137249-45b01100-7425-11ea-9023-4ac7e884bbff.png) ![Bildschirmfoto von 2020-04-01 15-20-43](https://user-images.githubusercontent.com/16668527/78141882-6a5bb700-742c-11ea-836d-f71373c4aa1d.png)

flourish86:
I see the problem and this was also regarded when initially implemented.

The "Now" bubble is the smallest and so the most likely one to be overlapped completely, so it was put into the foreground. If this other bubbles are overlapped, they can be put in foreground by hovering on them.

I agree, that this is far from ideal, but I suspect there won't be a fix without using Javascript, which we usually try to avoid.

nilmerg:
LOL. I didn't knew that hovering shows the underlying bubble! Yes, not perfect, but acceptable if JS should be avoided. (provided that this workaround is known)

PR #338, opened 2021-09-06
Fix downtime progress bar bubble overflow

### LINKED ISSUE #264: Progress bar start/end bubble overflow
## Describe the bug

This appears in Firefox, Chrome and Safari. 

Screenshot 2021-08-24 at 15 03 35

sukhwinder33445:
fix #264

### DIFF
--- a/library/Icingadb/Widget/Detail/DowntimeCard.php
+++ b/library/Icingadb/Widget/Detail/DowntimeCard.php
@@ -110,9 +110,7 @@ class DowntimeCard extends BaseHtmlElement
                     'style' => sprintf(
                         'left: %F%%; width: %F%%;',
                         $flexStartLeft,
-                        $hPadding
-                        + $this->calcRelativeLeft(time(), null, null, -$hPadding + 3)
-                        - $flexStartLeft
+                        $hPadding + $this->calcRelativeLeft(time()) - $flexStartLeft
                     )
                 ]);
             }
--- a/library/Icingadb/Widget/Detail/DowntimeCard.php
+++ b/library/Icingadb/Widget/Detail/DowntimeCard.php
@@ -118,22 +118,23 @@ class DowntimeCard extends BaseHtmlElement
             $above->add([
                 Html::tag(
                     'li',
-                    ['class' => 'bubble start'],
-                    new VerticalKeyValue(
-                        t('Scheduled Start'),
-                        new TimeAgo($this->downtime->scheduled_start_time)
+                    ['class' => 'start positioned'],
+                    Html::tag(
+                        'div',
+                        ['class' => 'bubble'],
+                        new VerticalKeyValue(t('Scheduled Start'), new TimeAgo($this->downtime->scheduled_start_time))
                     )
                 ),
                 Html::tag(
                     'li',
                     [
-                        'class' => 'bubble end',
+                        'class' => 'end positioned',
                         'style' => sprintf(
                             'left: %F%%',
                             $hPadding + $this->calcRelativeLeft($this->downtime->scheduled_end_time)
                         )
                     ],
-                    new VerticalKeyValue(t('Scheduled End'), $endTime)
+                    Html::tag('div', ['class' => 'bubble'], new VerticalKeyValue(t('Scheduled End'), $endTime))
                 )
             ]);
 
@@ -141,18 +142,26 @@ class DowntimeCard extends BaseHtmlElement
                 Html::tag(
                     'li',
                     [
-                        'class' => 'bubble upwards start' . ($evade ? ' left' : ''),
+                        'class' => 'start positioned',
                         'style' => sprintf('left: %F%%', $flexStartLeft)
                     ],
-                    new VerticalKeyValue(t('Start'), new TimeAgo($this->downtime->start_time))
+                    Html::tag(
+                        'div',
+                        ['class' => 'bubble upwards' . ($evade ? ' left' : '')],
+                        new VerticalKeyValue(t('Start'), new TimeAgo($this->downtime->start_time))
+                    )
                 ),
                 Html::tag(
                     'li',
                     [
-                        'class' => 'bubble upwards end' . ($evade ? ' right' : ''),
+                        'class' => 'end positioned',
                         'style' => sprintf('left: %F%%', $flexEndLeft)
                     ],
-                    new VerticalKeyValue(t('End'), new TimeUntil($this->downtime->end_time))
+                    Html::tag(
+                        'div',
+                        ['class' => 'bubble upwards' . ($evade ? ' right' : '')],
+                        new VerticalKeyValue(t('End'), new TimeUntil($this->downtime->end_time))
+                    )
                 )
             ]);
         } elseif ($this->downtime->is_flexible) {
@@ -170,24 +179,32 @@ class DowntimeCard extends BaseHtmlElement
             $above->add([
                 Html::tag(
                     'li',
-                    ['class' => 'bubble start'],
-                    new VerticalKeyValue(
-                        t('Scheduled Start'),
-                        time() > $this->downtime->scheduled_start_time
-                            ? new TimeAgo($this->downtime->scheduled_start_time)
-                            : new TimeUntil($this->downtime->scheduled_start_time)
+                    ['class' => 'start positioned'],
+                    Html::tag(
+                        'div',
+                        ['class' => 'bubble'],
+                        new VerticalKeyValue(
+                            t('Scheduled Start'),
+                            time() > $this->downtime->scheduled_start_time
+                                ? new TimeAgo($this->downtime->scheduled_start_time)
+                                : new TimeUntil($this->downtime->scheduled_start_time)
+                        )
                     )
                 ),
                 Html::tag(
                     'li',
                     [
-                        'class' => 'bubble end',
+                        'class' => 'end positioned',
                         'style' => sprintf(
                             'left: %F%%',
                             $hPadding + $this->calcRelativeLeft($this->downtime->scheduled_end_time)
                         )
                     ],
-                    new VerticalKeyValue(t('Scheduled End'), $endTime)
+                    Html::tag(
+                        'div',
+                        ['class' => 'bubble'],
+                        new VerticalKeyValue(t('Scheduled End'), $endTime)
+                    )
                 )
             ]);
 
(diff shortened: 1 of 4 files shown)


nilmerg:
Though of it again. It also applies to the `Now` above the bar. I'm also not convinced that it isn't possible to still center the indicator.

Please take a look [here](https://stackoverflow.com/a/15328843) and try that, possibly with an additional wrapper element for the bubbles.

sukhwinder33445:
The changes for this also affect the Host/service statistic bubbles. They also need to be fixed.

public/css/mixin/progress-bar.less
```diff
@@ -9,23 +9,32 @@
 
     .above {
       padding: ~"calc(1em + 1px) 10%";
+
+      .now {
+        width: auto;
+        z-index: 2;
```
nilmerg, line 15: Please check again here. I don't think that's correct or the only required change. This is related to the other z-index definitions and these you didn't touch. It's also related to
#44. The li's width is too wide atm or shouldn't be related to this "hover" behavior, but it blocks the hover at the moment on the right of `Now` if it's still very left.

public/css/mixin/progress-bar.less
```diff
@@ -9,23 +9,32 @@
 
     .above {
       padding: ~"calc(1em + 1px) 10%";
+
+      .now {
+        width: auto;
```
nilmerg, line 14: That's not required I think. It doesn't get a width otherwise and `auto` is the default anyway. At least the `start` and `end` wrappers also don't have this.

public/css/mixin/progress-bar.less
```diff
@@ -9,23 +9,32 @@
...
+      position: absolute;
+    }
+
     .bubble {
       .rounded-corners(.25em);
       background-color: @body-bg-color;
       border: 1px solid @gray-light;
-      position: absolute;
+      position: relative;
       bottom: 0;
```
nilmerg, line 32: `position` is not required or is it? Didn't made a difference for me when disabled at least.
`bottom` however is definitely useless for relative positioning anyway.

public/css/mixin/progress-bar.less
```diff
@@ -9,23 +9,32 @@
...
       box-shadow: 0 0 1em 0 rgba(0, 0, 0, .1);
-      margin-left: -4em;
       padding: .25em .5em;
       text-align: center;
-      width: 8em;
+      width: auto;
+      transform: translate(-50%, 0);
```
nilmerg, line 37: Maybe add a comment what this does and what advantage it has.

public/css/mixin/progress-bar.less
```diff
@@ -117,7 +114,7 @@
 
     .start .vertical-key-value,
     .end .vertical-key-value {
-      width: 6em;
+      width: auto;
```
nilmerg, line 117: This is also not required anymore. Remove the entire block.

public/css/mixin/progress-bar.less
```diff
@@ -9,23 +9,32 @@
...
+        width: auto;
+        z-index: 2;
+      }
     }
 
     .below {
       padding: 1.25em 10%;
     }
 
+    .placement {
```
nilmerg, line 23: Please use a different name. `positioned` maybe?

public/css/widget/check-statistics.less
```diff
@@ -35,6 +35,10 @@
       .bubble {
         position: relative;
```
nilmerg, line 36: This can go. `.bubble` isn't positioned anyway anymore now.

public/css/widget/check-statistics.less
```diff
@@ -35,6 +35,10 @@
...
       }
+
+      .end .bubble {
+        transform: translate(50%, 0);
```
nilmerg, line 40: Please put that into progress-bar.less. It's too strongly related to the positioning defined there. Also, maybe couple it with the `placement` class (or rather the new name :wink: )

nilmerg (changes requested):

public/css/widget/check-statistics.less
sukhwinder33445: This is necessary here because `.above` is a flex element here (but not in downtime progress bar) with `space-between`. These changes help to move the indicator to its correct position (end of `.above`).

public/css/mixin/progress-bar.less
sukhwinder33445: `.bubble::hover{z-index}` has no effect when the position is removed.

public/css/mixin/progress-bar.less
sukhwinder33445: Disabling both in dev-tools has no effect. So I have removed the block.

public/css/mixin/progress-bar.less
```diff
@@ -15,17 +15,22 @@
...
+      position: relative;
       box-shadow: 0 0 1em 0 rgba(0, 0, 0, .1);
-      margin-left: -4em;
       padding: .25em .5em;
       text-align: center;
-      width: 8em;
+      width: auto;
+      // The wrapper of .bubble is dynamically moved to the left based on the value of the progress bar
+      // This moves the center of the bubble to the beginning of the wrapper regardless of the size of the content.
+      transform: translate(-50%, 0);
```
flourish86, line 33: This doesn't seem to work properly, yet, when the bubbles are `.right` aligned like it is e.g. applied in `DowntimeCard.php` lines 162 and 145.

This is done, when start and end bubbles are close together.

This is why the bubbles have fixed widths.

Screen Shot 2021-09-28 at 13 14 56

public/css/mixin/progress-bar.less
```diff
@@ -15,17 +15,22 @@
...
       background-color: @body-bg-color;
       border: 1px solid @gray-light;
-      position: absolute;
-      bottom: 0;
+      position: relative;
       box-shadow: 0 0 1em 0 rgba(0, 0, 0, .1);
-      margin-left: -4em;
       padding: .25em .5em;
       text-align: center;
-      width: 8em;
```
flourish86, line 28: If you find a way to make apply `width: auto` to the bubbles and still make the alignment of the carets (`.bubble:before`) work, please also add a `max-width`.

Please keep in mind, that the content width can vary, when being translated.

Screen Shot 2021-09-28 at 13 27 21

It's also important to prevent `.start` and `.end` `.bubble` being obscured completely, since the user needs to be able to hover the obscured bubble to see its content.

flourish86 (commented):
We should also ensure that the `.keys` won't break, e.g. with `white-space: nowrap`.

Screen Shot 2021-09-28 at 13 30 27

One idea for the `.bubble` caret positioning would be to have multiple fixed width modifier classes (e.g. `wide`, `narrow`, `x-wide`, `x-narrow`, …). The classes would then be applied in the widget by checking the length of the strings.

public/css/mixin/progress-bar.less
sukhwinder33445: Should work with the latest changes 😄 .

public/css/mixin/progress-bar.less
sukhwinder33445: > It's also important to prevent `.start` and `.end` `.bubble` being obscured completely

The bubble can also be displayed by moving the mouse pointer over the bubble caret. 🤔 



> please also add a `max-width`

Why is this necessary? Bubble only contains the timestamp, and it can't be that long anyway.

flourish86: > > It's also important to prevent `.start` and `.end` `.bubble` being obscured completely
> 
> The bubble can also be displayed by moving the mouse pointer over the bubble caret. 🤔

I think, it's pretty obvious on the screenshot, that when one bubble is completely obscured, it's almost impossible to spot. It's also not quite clear at first sight, which caret belongs to which bubble.

And besides spotting it in the first place, the caret alone is a much too small target and it will be pretty annoying for the user to navigate the cursor there.

> > please also add a `max-width`
> 
> Why is this necessary? Bubble only contains the timestamp, and it can't be that long anyway.

If you can ensure, that it's wide enough for all date formats in all languages, I'm fine with having no `max-width`.

public/css/mixin/progress-bar.less
flourish86: Leaving out the left and right classes isn't the solution, sorry. Please restore the classes for preventing the bubbles to overlap.

Screen Shot 2021-09-28 at 17 32 28

In this example you see that problem with the obscured bubble.

flourish86 (changes requested):

flourish86:
I discovered another inconsistency on the `. timeline-overlay` which we should fix.

Could you take care of this, @sukhwinder33445?


Screen Shot 2021-09-29 at 12 05 44


Original Copy

flourish86 (approved):
Nice, all the issues solved.

public/css/mixin/progress-bar.less
```diff
@@ -71,8 +66,9 @@
       }
 
       &.right {
-        margin-left: -1.25em;
+        margin-left: 2.75em;
         margin-right: 0;
+        width: 8em;
```
nilmerg, line 71: We need neither of those. Apply this instead:

```suggestion
        // This is (.675em (:before placement) + .5em (half :before width)) + 1px (:before border)
        transform: translate(calc(-1.175em - 1px), 0);
```

public/css/mixin/progress-bar.less
```diff
@@ -82,7 +78,8 @@
       }
 
       &.left {
-        margin-left: -6.75em;
+        margin-left: -2.75em;
+        width: 8em;
```
nilmerg, line 82: Again, neither of those is needed:

```suggestion
        // entire width (moves the right border in place of the left) + (.675em (:before placement) + .5em (half :before width)) + 1px (:before border)
        transform: translate(calc(-100% + 1.175em + 1px), 0);
```

nilmerg (changes requested):

library/Icingadb/Widget/Detail/DowntimeCard.php
```diff
@@ -109,50 +109,61 @@ protected function assemble()
                     'class' => 'timeline-overlay downtime-elapsed',
                     'style' => sprintf(
                         'left: %F%%; width: %F%%;',
-                        $hPadding + $this->calcRelativeLeft($this->downtime->start_time),
-                        $this->calcRelativeLeft(time(), $this->downtime->start_time)
+                        $flexStartLeft,
+                        $hPadding
+                        + $this->calcRelativeLeft(time(), null, null, -$hPadding + 3)
+                        - $flexStartLeft
```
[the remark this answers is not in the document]
sukhwinder33445: > I discovered another inconsistency on the `. timeline-overlay` which we should fix.



This fixes that the green background line starts at the beginning of the starting point and not in the middle.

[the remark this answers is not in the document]
nilmerg: Okay fine, but why `-$hPadding + 3` (-7) as minimum value? In my tests this has zero effect. Also, a negative percentage is also not a valid width to begin with. And why do you negate the padding and substract `3` from it? This just looks so arbitrary to me.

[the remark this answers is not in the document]
sukhwinder33445: I found that removing `-$hPadding + 3` in else (fixed downtime) causes the now pointer to be positioned at `-76%`. This moves the marker out of the progress bar.

Screenshot 2021-10-05 at 16 02 14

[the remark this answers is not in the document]
nilmerg: Got it! :man_facepalming:

nilmerg (approved):

PR #343, opened 2021-09-07
Set active state on list items after refresh

### LINKED ISSUE #341: Active state on object list item not set after page reload
## Describe the bug
Active state on object list item not set after page reload for the following controllers:

- [x] Notifications
- [x] Downtimes
- [x] Comments
- [x] Users
- [x] User/Host/Service Groups
- [x] History

sukhwinder33445:
fix #341

### DIFF
--- a/library/Icingadb/Common/BaseListItem.php
+++ b/library/Icingadb/Common/BaseListItem.php
@@ -148,20 +148,6 @@ abstract class BaseListItem extends BaseHtmlElement
     {
     }
 
-    protected function setMultiselectFilter(Rule $filter)
-    {
-        $this->addAttributes(['data-icinga-multiselect-filter' => '(' . QueryString::render($filter) . ')']);
-
-        return $this;
-    }
-
-    protected function setDetailFilter(Rule $filter)
-    {
-        $this->addAttributes(['data-icinga-detail-filter' => QueryString::render($filter)]);
-
-        return $this;
-    }
-
     protected function assemble()
     {
         $this->add([
--- a/library/Icingadb/Widget/ItemList/BaseCommentListItem.php
+++ b/library/Icingadb/Widget/ItemList/BaseCommentListItem.php
@@ -106,8 +106,8 @@ abstract class BaseCommentListItem extends BaseListItem
 
     protected function init()
     {
-        $this->setMultiselectFilter(Filter::equal('name', $this->item->name));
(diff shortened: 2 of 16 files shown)


library/Icingadb/Common/BaseTableRowItem.php
```diff
@@ -79,6 +81,20 @@ protected function createVisual()
...
+
+        return $this;
+    }
+
+    protected function setDetailFilter(Rule $filter)
+    {
+        $this->addAttributes(['data-icinga-detail-filter' => '(' . QueryString::render($filter) . ')']);
+
+        return $this;
+    }
```
nilmerg, line 96: The list already provides this with `$this->list->addDetailFilterAttribute()` etc.

library/Icingadb/Common/BaseListItem.php
```diff
@@ -157,7 +157,7 @@ protected function setMultiselectFilter(Rule $filter)
 
     protected function setDetailFilter(Rule $filter)
     {
-        $this->addAttributes(['data-icinga-detail-filter' => QueryString::render($filter)]);
+        $this->addAttributes(['data-icinga-detail-filter' => '(' . QueryString::render($filter) . ')']);
```
nilmerg, line 160: There is `Icinga\Module\Icingadb\Common\DetailActions` for this actually. Any user of this method and the one above should use the list's utility methods instead.

nilmerg (changes requested):

library/Icingadb/Common/BaseListItem.php
sukhwinder33445: Okay, I will add `DetailActions` instead. 
The parentheses were missing anyway or not?


> should use the list's utility methods instead

You mean the `DetailActions` ?

nilmerg: Take a look into e.g. `BaseHostListItem` on how this handles this. Then I guess you'll know what I mean?

nilmerg (changes requested):
Please also remove the now unused methods.

sukhwinder33445:
I have already removed the methods from `BaseTableRowItem`.  Or do you mean the method in the class `BaseListItem`? 🤔

nilmerg (approved):
> Or do you mean the method in the class BaseListItem?

Exactly those :)

PR #346, opened 2021-09-08
Fix view mode switching shows less items when using vertical pagination

### LINKED ISSUE #298: The view mode switcher shows less items when switching from the minimum to the common or detailed view.
## Describe the bug
The view mode switcher shows fewer items when switching from the minimum to the common or detailed view.
This problem occurs when there is no pagination control but the` Load more` button is present.

## To Reproduce

1. Go to History/Notification (Make sure that the `minimum view` is active in the `view mode switcher`.)
2. Load at least one more page (there should be 2 pages or more).  

> In `minimal mode`, the list is limited to 50 entries by default. The two pages therefore show `2 x 50 = 100 entries`.
> In `common` and `detailed view` mode, the default limit is set to 50. So `100 / 25 = 4 pages` are needed to load all entries that were visible in minimum mode.

3. Switch to `common/detailed view mode`. 
4. There are only `3 pages x 25 = 75 items`.

## Expected behavior

When the view mode is switched, always the same number of entries should be displayed. In this case, 4 pages are needed to list 100 entries.

Yonas-net:
resolves #298

### DIFF
--- a/library/Icingadb/Web/Controller.php
+++ b/library/Icingadb/Web/Controller.php
@@ -360,8 +360,11 @@ class Controller extends CompatController
      *
      * @return ViewModeSwitcher
      */
-    public function createViewModeSwitcher(PaginationControl $paginationControl, LimitControl $limitControl)
-    {
+    public function createViewModeSwitcher(
+        PaginationControl $paginationControl,
+        LimitControl $limitControl,
+        $verticalPagination = false
+    ) {
         $viewModeSwitcher = new ViewModeSwitcher();
         $viewModeSwitcher->setIdProtector([$this->getRequest(), 'protectId']);
 
@@ -395,7 +398,13 @@ class Controller extends CompatController
 
         $viewModeSwitcher->on(
             ViewModeSwitcher::ON_SUCCESS,
-            function (ViewModeSwitcher $viewModeSwitcher) use ($user, $preferredModes, $paginationControl, &$session) {
+            function (ViewModeSwitcher $viewModeSwitcher) use (
+                $user,
+                $preferredModes,
+                $paginationControl,
+                $verticalPagination,
+                &$session
+            ) {
                 $viewMode = $viewModeSwitcher->getValue($viewModeSwitcher->getViewModeParam());
                 $requestUrl = Url::fromRequest();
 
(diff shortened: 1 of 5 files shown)


library/Icingadb/Web/Controller.php
```diff
@@ -453,10 +468,14 @@ function (ViewModeSwitcher $viewModeSwitcher) use ($user, $preferredModes, $pagi
             }
         )->handleRequest(ServerRequest::fromGlobals());
 
-        if ($viewModeSwitcher->getViewMode() === 'minimal') {
+        $previousMode = $session->get('previous_mode');
```
nilmerg, line 471: What does this fix? Or why is it required for this fix?

Yonas-net: Because when switching from minimal to the other modes and we use the vertical pagination, the default page limit remains unchanged and we don't have to recalculate the page number.

nilmerg (changes requested):
Test:

1. Go to `Icinga DB -> Notifications`
2. Switch to minimal mode if not enabled already
3. Click `Load More` (There are now two pages with 100 items in total)
4. Switch to common mode (There are now four pages with 100 items in total)

Result: Failed
Reason: There are only two pages in common mode

nilmerg (changes requested):
Test:

1. Go to `Icinga DB -> Notifications`
2. Switch to minimal mode if not enabled already
3. Click `Load More` (There are now two pages with 100 items in total)
4. Switch to common mode (There are now four pages with 100 items in total)

Result: Success

Test:

1. Perform first test
2. Click `Load More` (There are now five pages with 125 items in total)
3. Switch to minimal mode (There are now three pages with 150 items in total)

Result: Failed
Reason: There are just two pages in minimal mode

nilmerg (approved):
Test:

1. Go to `Icinga DB -> Notifications`
2. Switch to minimal mode if not enabled already
3. Click `Load More` (There are now two pages with 100 items in total)
4. Switch to common mode (There are now four pages with 100 items in total)

Result: Success

Test:

1. Perform first test
2. Click `Load More` (There are now five pages with 125 items in total)
3. Switch to minimal mode (There are now three pages with 150 items in total)

Result: Success

Test:

1. Perform second test
2. Switch to common mode (There are now five pages with 125 items in total)
3. Click `Load More` (There are now six pages with 150 items in total)
4. Switch to minimal mode (There are now three pages with 150 items in total)
5. Switch to common mode (There are now six pages with 150 items in total)

Result: Success

PR #350, opened 2021-09-10
Add support for `json` and `cvs` format

### LINKED ISSUE #262: Not support the requested output format: cvs/json
## Describe the bug
Support for the cvs/json format is not available.

Screenshot 2021-08-24 at 13 59 37

Clicking on the two options leads to the following error message:

Screenshot 2021-08-24 at 13 59 57

sukhwinder33445:
fix #262


sukhwinder33445:
Tests failed:
Throws an error in the Host/Service Groups Controller.
Allowed memory size exhausted. (History/Notification Controller)

sukhwinder33445:
Tests failed:
Maximum execution time of 30 seconds exceeded. (History Controller)

Can you please test this on your device too, maybe my device is just very slow?

nilmerg:
Nope, it's not just your device: Allowed memory size of 134217728 bytes exhausted (tried to allocate 1249280 bytes)

I just remembered that there's still Zend's output buffering active. You need to disable that in order to really stream the data.

sukhwinder33445:
Json export for `HistoryController` works fine now.

csv still need some fixes.

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
             return true;
         }
 
+        if ($this->format === 'json' || $this->format === 'csv') {
+            $isJsonFormat = $this->format === 'json';
+            $response = $this->getResponse();
+
+            $fileName = $this->getRequest()->getControllerName();
```
nilmerg, line 557: Please use `$this->view->title` if set. It's localized and better suited.

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
...
+            if ($isJsonFormat) {
+                $response
+                    ->setHeader('Content-Type', 'application/json')
+                    ->setHeader('Cache-Control', 'no-store')
+                    ->setHeader(
+                        'Content-Disposition',
+                        'inline; filename=' . $fileName . '.json'
```
nilmerg, line 564: I think we should use `attachment` here as well. It's useless to have a list of thousands of JSON objects in the browser window.

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
...
+                $response
+                    ->setHeader('Content-Type', 'text/csv')
+                    ->setHeader('Cache-Control', 'no-store')
+                    ->setHeader(
+                        'Content-Disposition',
+                        'attachment; filename=' . $fileName . '.csv'
+                    )
+                    ->sendResponse();
+            }
+            // Disable output buffering to stream large amounts of data without a memory exhausted error.
```
nilmerg, line 577: Not sure about the usefulness of this comment to be honest, but anyway, add an empty line beforehand. There's a control structure above it after all.

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
...
+            ob_end_clean();
+            set_time_limit(0);
```
nilmerg, line 579: Use the `Environment` class here instead. And don't let it run forever, use the default timeout of 300 seconds.

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
...
+
+            $query = $queries[0];
+            $query->limit(Url::fromRequest()->getParam('limit'));
+            $query->resetOffset();
```
nilmerg, line 583: Why do you reset the offset? If I use `?limit=50&page=2`, I want the results from page 2 only.

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
...
+            $this->applyRestrictions($query);
```
nilmerg, line 584: The restrictions are already applied.

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
...
+
+            $keyName = is_array($query->getModel()->getKeyName())
+                ? $query->getModel()->getKeyName()
+                : (array) $query->getModel()->getKeyName();
+            $col = array_merge($keyName, $query->getModel()->getColumns());
```
nilmerg, line 589: Simplify this to `$col = array_merge((array) $query->getModel()->getKeyName(), ...` please.

And don't export only the base model's columns. There may be more models (tables) joined in and these can and should be exported as well. Identify them with `$query->getWith()`. Joined table columns should be prefixed with the paths that are returned by it.

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
...
+
+            if ($isJsonFormat) {
+                echo('[');
+            }
+
+            $rs = $query->execute()->disableCache();
+
+            foreach ($rs as $count => $row) {
+                if ($count) {
```
nilmerg, line 598: `$i` is the better known (for other readers) symbol here. It also makes more sense when using `if ($i > 0)`

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
...
+                    $separator = $isJsonFormat ? ',' : PHP_EOL;
```
nilmerg, line 599: Here `PHP_EOL` and below `\r\n`, why the difference?

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
...
+                    if (
+                        $alias === 'id' || substr($alias, -3) === '_id'
+                        || substr($alias, -9) === '_checksum'  || substr($alias, -4) === '_bin'
+                    ) {
+                        $result[$alias] = base64_encode($row->$alias);
+                    }
+                }
+
+                if ($count === 0 && ! $isJsonFormat) {
+                    echo('SEP=,' . "\r\n");
```
nilmerg, line 618: Where does this come from? It confuses my libre office calc:
Screenshot from 2021-09-24 14-36-32

library/Icingadb/Web/Controller.php
```diff
@@ -550,6 +550,86 @@ public function export(Query ...$queries)
...
+                    echo(implode(', ', array_keys($result)) . "\r\n");
+                }
+
+                $data = $isJsonFormat ? Json::sanitize($result) : implode(',', (array_values($result)));
```
nilmerg, line 622: Why the parentheses around `array_values`?

library/Icingadb/Web/Controller.php
```diff
@@ -713,4 +793,46 @@ protected function moduleInit()
             ->getPlugin('Zend_Controller_Plugin_ErrorHandler')
             ->setErrorHandlerModule('icingadb');
     }
+
+    private function prepareQueryResults($queries)
```
nilmerg, line 797: Please remove.

nilmerg (changes requested):

library/Icingadb/Web/Controller.php
sukhwinder33445: 300 is already default, but was not enough when streaming `json` for history (Maximum execution time of 300 seconds exceeded). 🤔

library/Icingadb/Web/Controller.php
sukhwinder33445: This tells Excel how to separate the columns. I added this because otherwise Excel inserts all the columns into the first cell.

sukhwinder33445: I have tested the `csv` in the monitoring module. My Excel also inserts the columns there in the first cell. (so this is something with `Excel on Mac`, not the export code). Therefore, we do not need this.

library/Icingadb/Web/Controller.php
nilmerg: Well. I'd say exporting the *entire* history isn't something that should be done anyway. Most users will still limit this to a reasonable size. It just shouldn't abort with a memory exhaustion error, a timeout error especially after five minutes is okay though.

sukhwinder33445: I see, but this means that the `json` stream stops after 5 minutes of execution and the trailing `]` or comma cannot be set if it is needed in between. 🤔

nilmerg: It stops *nicely*? Isn't there an error message?

sukhwinder33445: Yes, there is. Ahh, I see. So, if it takes longer than 5 minutes, then it's OK to show this error and raw json.

nilmerg: Well, the response code is 500, right? (I hope so and if so, it's fine)

nilmerg: No it's not 500, headers are sent at this stage....

nilmerg: Yeah, headers are sent, so there's no way to change the response code. Since we also have no special high level *protocol* here, other than the content type syntax, I see no other solution than running into the timeout and returning an error. It's no JSON or CSV after all, so anything processing the result should detect this and handle it accordingly.

library/Icingadb/Web/Controller.php
```diff
@@ -566,6 +567,111 @@ public function export(Query ...$queries)
...
+                if ($i > 0) {
+                    $separator = $isJsonFormat ? ',' : "\r\n";
+                    echo($separator);
+                }
+
+                foreach ($col as $alias => $name) {
+                    if (is_int($alias)) {
+                        $alias = $name;
+                    }
+                    if (0 === strpos($alias, $tableName)) {
```
Yonas-net, line 632: You are checking here whether the table name of the query models (history) is at position `0`. But might there not be such an alias? `host.state.*`

Yonas-net (commented):
I'm not sure if it's due to my laptop, but I can't export the histories. As you already mentioned above, after a while I get a network error.

Yonas-net:
Is it really necessary, e.g. for `Comments` and `Downtimes`, to export all the properties from the parent `Host/Service` object? Actually you only want to export downtimes and not all the relations too. IMO only the parent `Host/Service` name or `id` would be sufficient there.

nilmerg:
> Is it really necessary, e.g. for Comments and Downtimes, to export all the properties from the parent Host/Service object?

That's indeed not strictly necessary, but a differentiation shouldn't be done automatically in such a generic method. We may add a url parameter to limit the result to specific properties in the future.

nilmerg (changes requested):
The CSV export is still messed up.
[Hosts.csv](https://github.com/Icinga/icingadb-web/files/7275958/Hosts.csv)

sukhwinder33445:
After a long struggle 😆 , it looks like csv is now working as expected.

[Hosts (6).csv](https://github.com/Icinga/icingadb-web/files/7279202/Hosts.6.csv)

PR #352, opened 2021-09-10
Better declare empty values

### LINKED ISSUE #349: More explicitly declare ungiven values
We should more explicitly declare not given values.

Currently most values in tables, that are not given are indicated with a "-".

A simple "None" or "No value" would in most cases indicate more clearly, that there's no value set.

We also do this currently already in some places, for example in Host/Service Detail on Users/User Groups.

Screen Shot 2021-09-09 at 16 00 43

With using the same design, the user detail would look like this.

Screen Shot 2021-09-09 at 16 04 43

Places, where this applies:
* User Detail
** Pager
** Hosts States: "No Restrictions"/"All"
** Service States: "No Restrictions"/"All"
** Types: "All"

* HostService Detail
** Perfdata Values: "None"
** Custom Var Empty Strings: "empty" or "" (currently the value is empty)
Screen Shot 2021-09-09 at 16 11 01

* Host/Service Source Tab
** "null" for empty keys (if technically appropriate, please verify)

nilmerg:
### Event Detail
Screenshot from 2021-09-10 16-27-26
### Host meta info
Screenshot from 2021-09-10 16-26-14
### User Detail
Screenshot from 2021-09-10 16-25-27
### Downtime Detail
Screenshot from 2021-09-10 16-23-39
### Customvar Table
Screenshot from 2021-09-10 16-22-55
### Perfdata Table
Screenshot from 2021-09-10 16-22-36
### Source Tab
Screenshot from 2021-09-23 12-06-09

fixes #349

### DIFF
--- a/library/Icingadb/Common/ObjectInspectionDetail.php
+++ b/library/Icingadb/Common/ObjectInspectionDetail.php
@@ -206,7 +206,7 @@ abstract class ObjectInspectionDetail extends BaseHtmlElement
     private function formatTimestamp($ts)
     {
         if (empty($ts)) {
-            return '-';
+            return new EmptyState(t('n. a.'));
         }
 
         if (is_float($ts)) {
@@ -252,7 +252,7 @@ abstract class ObjectInspectionDetail extends BaseHtmlElement
         $table->addAttributes(['class' => 'name-value-table']);
         foreach ($data as $name => $value) {
             if (empty($value) && ($value === null || is_string($value) || is_array($value))) {
-                $value = '-';
+                $value = new EmptyState(t('n. a.'));
             } elseif (isset($formatters[$name])) {
                 $value = call_user_func($formatters[$name], $value);
             } else {
--- a/library/Icingadb/Widget/Detail/CustomVarTable.php
+++ b/library/Icingadb/Widget/Detail/CustomVarTable.php
@@ -4,6 +4,7 @@
 
 namespace Icinga\Module\Icingadb\Widget\Detail;
 
+use Icinga\Module\Icingadb\Widget\EmptyState;
 use ipl\Html\BaseHtmlElement;
 use ipl\Html\Html;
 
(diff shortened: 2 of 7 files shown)


nilmerg:
The `Source` tab is built way too generic to be able to provide meaningful empty values there. A simple `None` doesn't suit always. But it's a *technical* view after all, so I don't see why a `-` isn't sufficient.

flourish86:
> But it's a _technical_ view after all, so I don't see why a `-` isn't sufficient.

We already agreed in our meeting, that we should go for clearer solution than the `-` for consistency, didn't we?

> The `Source` tab is built way too generic to be able to provide meaningful empty values there. A simple `None` doesn't suit always.

I think a generic `(null)` in the proven grayed out design is not only appropriate (especially in the technical context) but also makes totally clear, that there's no value given. And that shouldn't be that hard to replace, is it?

nilmerg:
> (especially in the technical context)

I really don't want to be nitpicky, but I have to: If we're in the **technical** context, `null` is inappropriate because it may be an actual value somewhere else. The *Source* view is a generic view and only knows for few details the data type. It does not know it's meaning. An integer of zero may mean there is *nothing* (a count maybe) or that there is no date set (something doesn't expire, hasn't started..). It currently shows the dash in such cases, because that's a symbol a *technician* should know. The zero (`0`) is the other alternative, but requires IMHO too much from the user, the dash is the simpler signal to emphasize that there's no meaningful value.

flourish86:
> I really don't want to be nitpicky, …

Good point, so good nitpickyness. I also see the point with `0`.

Anyway, We should apply the faded out style, so that the given values stand out visually. This would improve it already pretty much.

I'm not super happy about the `-` (as you may have noticed). I'd rather use a term, that has an explicit message and not have the user assume something.

But if you insist on the `-`, I'll go with it, when we have the distinct style for differentiation.

Here are some more suggestions:

* n.a.
* empty

nilmerg:
I don't insist on the dash, I just want a neutral term or symbol that doesn't imply a meaning. So I'm fine `n.a.`.

flourish86:
> … So I'm fine with `n.a.`.

Nice! 🙌

flourish86 (approved):
Looks perfect to me. 💎

PR #357, opened 2021-09-16
Fix failed localdateTime validation in Safari

### LINKED ISSUE #255: Downtime cannot be scheduled because of Validation Error (Safari only)
## Describe the bug
A clear and concise description of what the issue is.

## To Reproduce

1. Open Host or Service Detail in Safari
2. Click Schedule Downtime Action
3. Enter Comment
4. Submit form by clicking "Schedule" button

## Expected behavior
Validation should work and the Downtime should be scheduled (see correct behavior in Chrome/FF)

## Screenshots
Screen Shot 2021-08-23 at 14 34 09

sukhwinder33445:
This fixes the problem of failed localdateTime validation in Safari.

fix #255

### DIFF
--- a/application/forms/Command/Object/AcknowledgeProblemForm.php
+++ b/application/forms/Command/Object/AcknowledgeProblemForm.php
@@ -147,6 +147,7 @@ class AcknowledgeProblemForm extends CommandForm
                 [
                     'data-use-datetime-picker'  => true,
                     'required'                  => true,
+                    'step'                      => 1,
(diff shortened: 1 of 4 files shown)


nilmerg (commented):
I'd rather make `1` the default for the `step` attribute in `ipl\Html\FormElement\LocalDateTimeElement`

sukhwinder33445:
Fixed in https://github.com/Icinga/ipl-html/pull/54

PR #360, opened 2021-09-17
Add keyboard navigation for lists

### LINKED ISSUE #106: Keyboard Navigation for Lists
It would be a great addition be able to navigate the lists by Keyboard.

This would (not exclusively) include the following commands:

- Arrow Up/Down: Select next/previous list item. If none is selected, arrow down would select the first, arrow up the last in the current page.

- Shift + Arrow Up/Down: Extend the selection by the next item, Reduce selection by the last item.

sukhwinder33445:
resolves #106 

Merge https://github.com/Icinga/icingadb-web/pull/642 before merging this PR.

## blocked by:
- https://github.com/Icinga/icingaweb2/pull/5049


flourish86 (changes requested):
Works great so far. There are some issues, or respectively some changes to do, though:

1) After a hard reload, the keyboard selection won't work.
We should handle this as discussed:
* if there’s a list in any column, this should be the active view, where the keyboard shortcuts should be applied on, even directly after a hard reload
* if there’s lists in both columns, we should priorize col1. This is not the ultimate solution, but a practical one, since this should occur only rarely.

2) Cmd/Ctrl + A Currently should work without a filter applied. Pressing Cmd/Ctrl + A should then:
* select all visible list items in the active column (i.e. items from current page)
* open #col2 with a multiselect view of the selected items
* focus #col2
* if the active list is in #col2, the list should move to #col1 and open the multiselect view in #col2

3) As discussed, the arrow key selection shouldn’t go over the edge, which means when the last item is selected pressing the arrow down button shouldn’t have any (further) effect (same for first and arrow up).

flourish86 (changes requested):
Most things seem to work now, but there's some issues:

1) It seems that keyboard navigation only works, when the focus is in the same column as the list. This results in some problems, that should be solved:
* Still, after a hard reload I can't use keyboard navigation. I have to either tab as long as an element in col1 is focussed or re-click on the current view in the main navigation sidebar.
2) Ctrl + A works great. Cmd + A never works for me. (Safari)
3) "Select All" (Ctrl + A) results in "page not found", if there's no Multiselect action, e.g. Host Groups List, etc
* I think we should then disable "Select All", if there's no multi action available

flourish86 (changes requested):
Some issues I’ve found:
* There’s weird behaviour (at least in Safari, but probably Chrome as well), when you quickly select multiple items after another.
* if there are two columns with lists in Safari, keyboard nav won’t work. E.g. http://localhost/icingaweb2/icingadb/hosts#!/icingaweb2/icingadb/services
* There’s now two item count elements in the list footer Screenshot 2023-05-24 at 16 59 38

nilmerg:
Didn't find any obvious issues related to the navigation. Though, you allow to quickly cycle through items if the user keeps pressing up or down. This has the effect that every item is loaded, but the requests get cancelled nearly immediately. But cancelling them isn't enough, the server still processes them. So this quickly overloads the server. You should only load an item after a short delay (e.g. 200-500ms), so only the item is loaded where the user stops pressing the key.

flourish86 (approved):
It seems like the additional changes made some significant difference. It works smooth as fluff, as far as I could find.

There’s one usability issue though. It’s not obvious, if multiselect will work or not. The user has to find out via trial and error. While that’s kind of acceptable in the first place, we should consider adding a hint instead of letting him try and then requiring to remember it for every list type.

OS UIs do this by playing a sound. Since it would probably be a little over the top to introduce sound effects, maybe we can come up with a visual effect to achieve this (probably in another issue).

public/js/action-list.js
```diff
@@ -30,6 +30,14 @@
             this.on('close-column', this.onColumnClose, this);
 
             this.on('rendered', '.container', this.onRendered, this);
+            this.on('keydown','#body', this.onKeyDown, this);
```
nilmerg, line 33: It should be possible to avoid listening on body, should it? I mean, listening on `#col1, #col2, .action-list` should also suffice, right?

nilmerg (changes requested):

raviks789:
I feel the list view should scroll up/down after the selected list item is the first/last item in the view. In my opinion visually your eyes follows the first/last selected element while scrolling up/down and it would be somewhat annoying if the view is scrolled up/down before reaching the first/last item in the view.

public/js/action-list.js
```diff
@@ -76,74 +83,378 @@
...
+            }, 250);
+        }
+
+        handleLoadMoreNavigate(loadMoreElement, lastActivatedItem) {
+            let req = this.loadMore($(loadMoreElement.querySelector('a')));
+            this.isProcessingLoadMore = true;
+            req.done(() => {
+                this.isProcessingLoadMore = false;
+                // list has now new items, so select the lastActivatedItem and then move forward
+                let toActiveItem = document.querySelector('.action-list').querySelector(`[data-icinga-detail-filter="${ lastActivatedItem.dataset.icingaDetailFilter }"]`);
```
nilmerg, line 363: a) Don't use `document` as reference. Use `loadMoreElement.closest('.action-list')` instead.
b) Wrap long lines also in JS please

public/js/action-list.js
```diff
@@ -76,74 +83,378 @@
...
+                this.setActive(toActive, list.lastChild);
+            }
+        }
+
+        clearSelection(selectedItems) {
+            selectedItems.forEach(item => item.classList.remove('active'));
+        }
+
+        setActive(toActiveItem, markAsLastActive) {
+            if (toActiveItem instanceof HTMLElement) {
```
nilmerg, line 324: When can `toActiveItem` not be an `HTMLElement`?

public/js/action-list.js
```diff
@@ -76,74 +83,378 @@
...
+                toActiveItem.previousElementSibling
+                    ? toActiveItem.previousElementSibling.scrollIntoView({block: "nearest"})
+                    : null;
+            }
+
+            this.lastActivatedItemUrl = markAsLastActive
+                ? markAsLastActive.dataset.icingaDetailFilter
+                : toActiveItem.dataset.icingaDetailFilter;
+
+            let list = document.querySelector('.action-list');
```
nilmerg, line 339: There may be not just one action list. Think of an alternative way of identifying the list.

public/js/action-list.js
```diff
@@ -76,74 +83,378 @@
...
+            let selectedItems = selectionCount.querySelector('.selected-items');
+            selectedItems.innerText = activeItemCount
+                ? list.dataset.icingaMultiselectCountLabel.replace('%d', activeItemCount)
+                : list.dataset.icingaMultiselectHintLabel;
+
+            activeItemCount === 0 ? selectedItems.classList.add('hint') : selectedItems.classList.remove('hint');
+        }
+
+        onKeyDown(event) {
+            let list = document.querySelector('.action-list'); // col1 if both given, intended
```
nilmerg, line 162: Again, don't use document as reference please.

public/js/action-list.js
```diff
@@ -76,74 +83,378 @@
...
+            let _this = event.data.self;
+
+            if (
+                ! list
+                || list.querySelectorAll(':scope > [data-action-item]').length === 0
+                || _this.isProcessingLoadMore
+                || (document.querySelector('.search-suggestions')
+                    && document.querySelector('.search-suggestions').hasChildNodes()) // search suggestion list is visible
```
nilmerg, line 170: This shouldn't be needed if you don't listen on events emitted by them.

public/js/action-list.js
```diff
@@ -76,74 +83,378 @@
...
+
+            let activeItems = list.querySelectorAll(':scope > [data-action-item].active');
+            let isMultiSelectableList = list.matches('[data-icinga-multiselect-url]');
+            let pressedArrowDownKey = event.key === 'ArrowDown';
+            let pressedArrowUpKey = event.key === 'ArrowUp';
+
+            if (isMultiSelectableList && (event.ctrlKey || event.metaKey) && event.keyCode === 65) { // ctrl|cmd + A
+                    event.preventDefault();
+                    _this.selectAll(list);
+                    return;
```
nilmerg, line 183: Consistent indentation please.

public/js/action-list.js
```diff
@@ -76,74 +83,378 @@
...
+                    _this.clearSelection(activeItems);
+
+                    _this.detailUrl = continueWith.getAttribute('href');
+                    _this.icinga.loader.loadUrl(
+                        _this.detailUrl, _this.icinga.loader.getLinkTargetFor($(continueWith))
+                    );
+                }
+
+                return;
+            }
```
nilmerg, line 199: I thought we decided to drop this?

nilmerg (changes requested):
On the dashboard, only the first dashlet accepts the input. Please either disable the arrow key navigation there or make it possible to navigate through all dashlets.

public/js/action-list.js
sukhwinder33445: It can be a single `HtmlElement` OR a `NodeList` (`selectAll()` for example).

nilmerg (changes requested):
Which list is navigable and which isn't, isn't obvious anymore now. I think the following rules should meet our requirements:

--

Which lists are navigable?
- primary lists
- the list which's container has the focus or the list which's item has the focus or contains the element that has it

i.e. isPrimaryList && (document.activeElement === container || list.contains(document.activeElement))

What's a primary list?
- the container is a top level one
- the list is a direct child of said container's content

i.e. container.matches('#main > :scope') && list.matches('.content > :scope')

--

With these rules, the following should be the case:

click on hostgroups overview = hostgroup list is navigable, the host list is not
 -> click on a host = host list is navigable

click on hosts overview = host list is navigable
 -> click on Show all of a comment or downtime list = comment/downtime list is navigable
 -> click on the services or history tab = service or history list is navigable, the host list is not
 
 --

`data-icinga-navigable-list` may not be required anymore. The focus defines which list is navigable after all.

public/js/action-list.js
```diff
@@ -28,194 +28,622 @@
...
+                _this.selectAll(list);
+                return;
+            }
 
-                if ($list.is('[data-icinga-multiselect-url]')) {
-                    if (!footer.children('.selection-count').length) {
-                        footer.prepend('<div class="selection-count"></div>');
+            event.preventDefault();
+
+            list.closest('.container[id^=col]').dataset.suspendAutorefresh = '';
```
nilmerg, line 207: ~~.container[id^=col]~~ `#main > .container`

public/js/action-list.js
```diff
@@ -28,194 +28,622 @@
...
+            let isMultiSelect = isMultiSelectableList && event.shiftKey;
+            let wasAllSelected = activeItems.length === listItemsLength;
+            let lastActivatedItem = list.querySelector(
+                `[data-icinga-detail-filter="${ _this.lastActivatedItemUrl }"]`
+            );
+            let directionalNextItem = _this.getDirectionalNext(lastActivatedItem, event.key);
+            let markAsLastActive = null; // initialized only if it is different from toActiveItem
+            let toActiveItem = null;
+
+            switch (true) {
```
nilmerg, line 218: Please use an if here..

public/js/action-list.js
```diff
@@ -28,194 +28,622 @@
...
                     }
+                    break;
+                case isMultiSelect:
+                    if (activeItems.length === 1) {
+                        toActiveItem = directionalNextItem;
+                    } else if (wasAllSelected
+                        && (
+                        (lastActivatedItem !== list.firstChild && pressedArrowDownKey)
+                        || (lastActivatedItem !== list.lastChild && pressedArrowUpKey))
+                    ) {
```
nilmerg, line 234:
```suggestion
                    } else if (wasAllSelected && (
                        (lastActivatedItem !== list.firstChild && pressedArrowDownKey)
                        || (lastActivatedItem !== list.lastChild && pressedArrowUpKey)
                    )) {
```

public/js/action-list.js
```diff
@@ -28,194 +28,622 @@
...
+            event.stopPropagation();
+            event.preventDefault();
+
+            event.data.self.loadMore(event.target);
+
+            return false;
+        }
+
+        onKeyPress(event) {
+            if (event.which === 32) {
```
nilmerg, line 474: Consider using `event.key` here as well please

public/js/action-list.js
```diff
@@ -28,194 +28,622 @@
...
+            } else {
+                list = container.querySelector('.action-list');
+            }
 
             if (list && list.matches('[data-icinga-multiselect-url], [data-icinga-detail-url]')) {
-                let _this = event.data.self;
-                let detailUrl = _this.icinga.utils.parseUrl(_this.icinga.history.getCol2State().replace(/^#!/, ''));
+                let detailUrl = _this.icinga.utils.parseUrl(
+                    _this.icinga.history.getCol2State().replace(/^#!/, '')
+                );
```
nilmerg, line 624: redundant

public/js/action-list.js
```diff
@@ -28,194 +28,622 @@
...
 
-            if (event.currentTarget !== container) {
-                // Nested containers are not processed multiple times
+            if (event.currentTarget !== container || _this.lastTimeoutId !== null) {
+                // Nested containers are not processed multiple times || still processing selection/navigation
                 return;
             } else if (isTopLevelContainer && container.id !== 'col1') {
-                return;
-            }
+                list = document.querySelector('#main > #col1 .action-list');
```
nilmerg, line 595: I believe the list can be identified by identifying the item based on `detailUrl` first. Then the list can be anywhere and we're not limited to `#col1`.

public/js/action-list.js
```diff
@@ -28,8 +28,17 @@
 
             this.on('click', '.action-list [data-action-item]:not(.page-separator), .action-list [data-action-item] a[href]', this.onClick, this);
             this.on('close-column', this.onColumnClose, this);
+            this.on('column-moved', this.onColumnMoved, this);
 
             this.on('rendered', '.container', this.onRendered, this);
+            this.on('keydown', '#col1, #col2, .action-list', this.onKeyDown, this);
```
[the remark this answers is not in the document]
nilmerg: I believe this should also listen on events triggered while `body` is focused. That's the case directly after a full reload, and the navigation could then be also possible.

public/js/action-list.js
```diff
@@ -28,194 +28,622 @@
...
-                let activeItems = list.querySelectorAll('[data-action-item].active');
-
-                if (activeItems.length) {
-                    if (!container.querySelector('.footer')) {
-                        container.appendChild(notjQuery.render(
-                            '<div class="footer" data-action-list-automatically-added></div>'
-                        ));
-                    }
+                if (item) {
+                    _this.scrollItemIntoView(item);
```
nilmerg, line 642: Not sure if this is a good idea. The content could be rendered based on an autorefresh, and then the scroll position must be preserved, and not changed randomly. (The item may move it's position in the list, depending on the sort by rules)

nilmerg (changes requested):

public/js/action-list.js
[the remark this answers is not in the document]
nilmerg: What about my last comment?

public/js/action-list.js
```diff
@@ -28,194 +28,616 @@
 
             this.on('click', '.action-list [data-action-item]:not(.page-separator), .action-list [data-action-item] a[href]', this.onClick, this);
             this.on('close-column', this.onColumnClose, this);
+            this.on('column-moved', this.onColumnMoved, this);
 
             this.on('rendered', '.container', this.onRendered, this);
```
nilmerg, line 33: Should use `#main > .container` as selector. Otherwise the nested container check in the listener leads to false positives.

Issue #265, opened 2021-08-24
Performance is worse than in Monitoring in some areas


nilmerg:
## Describe the bug
In some areas, notably in the hosts/services lists, the response time is worse than in the comparable views of the Monitoring module.

## Expected behavior
Icinga DB Web **has to be** at least as fast as the Monitoring module.

nilmerg:
I just noticed that this may be due to Redis. That's a new external resource we're fetching data from and then some of the *delays* seem entirely reasonable.

Though, we should still think of a reduced timeout for redis connections. I've configured TLS for my Redis in Web without doing as well on Redis itself and this increased e.g. the host list response time to ~17 seconds! (I also smell multiple connection attempts when looking at this number, we have to ensure that only a single attempt is made)

nilmerg:
With 2.6 million entries in the history, a host's history finishes after 15s and a service's history after 30s. The host history query filters on `history.object_type` to not fetch service entries, we should add an index here. The service history query filters on `history.host.service.id` and `history.service.id` instead of just `history.service_id`, not sure why but with the latter in finishes after ~6s.

nilmerg:
> we have to ensure that only a single attempt is made

Difficult to accomplish. Elements are rendered lazily, but query results aren't evaluated in the same way.

nilmerg:
I've made a view script for the host list to see if there are any differences and how big/small they are. This revealed a difference of almost 200ms (average is rather 150ms) in response time compared to the non-view-script view and showed similar response times than those of the monitoring's host list. (Still +50ms or so, but that's to be expected since the integration of Redis as mentioned above)

The view-script is fully functional, besides the view mode switcher. The controls are all the same and are still rendered by ipl-html/ipl-web. The only difference is the host list itself. It's this which seems to have the most impact on performance atm.

nilmerg:
So, I've profiled and benchmarked a bit. A view script is the fastest, but also the most static solution. Using ipl/Html is way more dynamic after all, but comes at a cost. Though, I cannot identify a specific area where too much time is spent. It's a rather well spread usage over all involved parts. There are many small parts that cost time, but none of them too much and it's only the sum of everything which results in this *penalty* compared to a view script. So the *fix* I'd identify here is to use view scripts again. But, well, no, not going to happen.

[...]

PR #364, opened 2021-09-21
service/history: Help the query to use a good index

nilmerg:
refs #265


nilmerg:
May require https://github.com/Icinga/ipl-sql/pull/17

PR #368, opened 2021-09-23
Extend the width of the action list in the detail view

### LINKED ISSUE #134: Detail view list items hover background cropped.
## Describe the bug
The background of comment and downtime list items in detail views is cropped.

## Expected behavior
Something that results in the same as:

```
.host-detail .list-item {
  margin-left: -1em;
  margin-right: -1em;
  padding-left: 1em;
  padding-right: 1em;
}
```

## Screenshots
Screenshot 2020-10-19 at 18 01 44

Screenshot 2020-10-19 at 18 07 38

sukhwinder33445:
fix #134

### DIFF
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -186,10 +186,19 @@ div.show-more {
   padding: .5em;
 }
 
-.object-detail .item-list .list-item:last-of-type .caption {
-  min-height: 1.5em;
-  max-height: 3em;
-  height: auto;
+.object-detail .item-list {
+  &.action-list .list-item {
+    margin-right: -1em;
+    margin-left: -1em;
+    padding-right: 1em;
+    padding-left: 1em;
+  }
+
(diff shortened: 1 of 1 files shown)


public/css/common.less
```diff
@@ -186,10 +186,17 @@ div.show-more {
...
 }
 
-.object-detail .item-list .list-item:last-of-type .caption {
-  min-height: 1.5em;
-  max-height: 3em;
-  height: auto;
+.object-detail .item-list {
+  &.action-list .list-item {
+    margin: 0 -1em;
+    padding: 0 1em;
```
nilmerg, line 192: Please use `margin-left` and `margin-right` as well as `padding-left` and `padding-right`.

It's the horizontal margin/padding only that is of interest here, it shouldn't have an effect on the vertical one if not necessary.

nilmerg (changes requested):

PR #372, opened 2021-09-24
Use css variables for colors

### LINKED ISSUE #385: Adjust LESS variable usage for compatibility with Icinga Web 2.10

sukhwinder33445:
resolves #385

### DIFF
--- a/public/css/widget/state-change.less
+++ b/public/css/widget/state-change.less
@@ -2,7 +2,7 @@
   display: inline-flex;
 
   .state-ball {
-    box-shadow: 0 0 0 1px @body-bg-color;
+    .box-shadow(0, 0, 0, 1px, ~"var(--body-bg-color, @{body-bg-color})");
   }
 
   .state-ball ~ .state-ball {
@@ -35,7 +35,7 @@
   .state-ball.state-up {
     &.ball-size-l,
     &.ball-size-xl {
-      background-color: @body-bg-color;
+      .var(background-color, body-bg-color);
     }
   }
 }
--- a/public/css/widget/view-mode-switcher.less
+++ b/public/css/widget/view-mode-switcher.less
@@ -9,9 +9,9 @@
   }
 
   label {
-    color: @icinga-blue;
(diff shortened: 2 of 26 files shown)


public/css/widget/view-mode-switcher.less
```diff
@@ -9,9 +9,10 @@
   }
 
   label {
-    color: @icinga-blue;
+    display: inline-block;
```
Yonas-net, line 12: What is this for?

Yonas-net (dismissed):

public/css/widget/view-mode-switcher.less
sukhwinder33445: This is a change from @nilmerg. Can you please explain?

nilmerg: Good question...

nilmerg: Can't remember why I've added this. Doesn't seem to have an effect anymore either. Can be removed.

public/css/widget/state-change.less
```diff
@@ -2,7 +2,8 @@
   display: inline-flex;
 
   .state-ball {
-    box-shadow: 0 0 0 1px @body-bg-color;
+    box-shadow: 0 0 0 1px;
+    .var(color, body-bg-color); // box shadow color
```
Yonas-net, line 6: Apart from this change, everything looks fine. Because that's not how it's going to work. You should either define a variable for it or leave it as it was before.

sukhwinder33445: I got this idea from [SO](https://stackoverflow.com/questions/3012899/is-there-a-box-shadow-color-property). @nilmerg  What do you think about it? I will test it again on all the browsers we support.

nilmerg: In Icinga Web 2 `box-shadow` was already a subject for such changes. Take a look there how it's been solved.

nilmerg (approved):

PR #374, opened 2021-09-27
Add missing status details

### LINKED ISSUE #72: Missing status details
Some details about the status of an object are not shown everywhere or not at all. Below are some suggestions on how to solve what.

### Previous hard state only in history lists

The previous hard state should probably also be shown in the normal lists and the detail view.

**Mockups how this may look like:**
previous-hard-state-list-left
previous-hard-state-upper-left

### Flapping state not visible

There's no information that an object is currently flapping other than looking at its history. This should at least be included in the detail view, though no idea where exactly. (@flourish86 Your turn)

### Last state change only as relative date/time

A user would like to be able to copy the exact date and time of the last status change. <sup>1</sup>
To solve this, I have no other idea than exchanging what is shown in the detail view. i.e. Full date/time always, relative time in popup (upper right) (@flourish86 Your opinion please)

### Hostname and Address (ipv4 and ipv6) are not shown

We already have #55 for this, though I'm including it here as well. @flourish86 Please suggest how we'd show these.

## Additional context
¹ https://github.com/Icinga/icingaweb2/issues/4098

Yonas-net:
fixes #72

### DIFF
--- a/library/Icingadb/Widget/ItemList/HostListItemDetailed.php
+++ b/library/Icingadb/Widget/ItemList/HostListItemDetailed.php
@@ -49,6 +49,15 @@ class HostListItemDetailed extends BaseHostListItem
             );
         }
 
+        if ($this->item->state->is_flapping) {
+            $statusIcons->addHtml(new Icon(
+                'random',
+                [
+                    'title' => sprintf(t('Host "%s" is in flapping state'), $this->item->display_name),
+                ]
+            ));
+        }
+
         if (! $this->item->notifications_enabled) {
             $statusIcons->addHtml(new Icon('bell-slash', ['title' => t('Notifications disabled')]));
         }
--- a/library/Icingadb/Widget/ItemList/ServiceDetailHeader.php
+++ b/library/Icingadb/Widget/ItemList/ServiceDetailHeader.php
@@ -1,0 +1,53 @@
+<?php
+
+/* Icinga DB Web | (c) 2020 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Icingadb\Widget\ItemList;
+
+use Icinga\Module\Icingadb\Common\ServiceStates;
+use ipl\Html\Attributes;
+use ipl\Html\BaseHtmlElement;
+use ipl\Html\HtmlElement;
+use ipl\Web\Widget\StateBall;
+
+class ServiceDetailHeader extends ServiceListItemMinimal
+{
+    protected function getStateBallSize()
+    {
+        if ($this->state->state_type === 'soft') {
+            return StateBall::SIZE_MEDIUM_LARGE;
+        }
+
+        return StateBall::SIZE_BIG;
+    }
+
+    protected function assembleVisual(BaseHtmlElement $visual)
+    {
+        parent::assembleVisual($visual);
+
+        $isSoftState = false;
+        if ($this->state->state_type === 'soft') {
+            $isSoftState = true;
+        }
+
+        // When the current state type is a soft state change, then use the actual hard_state and not the prev. ones
+        $previousState = $isSoftState ? $this->state->hard_state : $this->state->previous_hard_state;
+        $previousHardState = ServiceStates::text($previousState);
+
+        $visual->getFirst('span')->addWrapper(new HtmlElement(
+            'div',
+            Attributes::create(['class' => 'state-change']),
+            new StateBall($previousHardState, StateBall::SIZE_BIG)
+        ));
+    }
+
+    protected function assemble()
+    {
+        $attributes = $this->list->getAttributes();
+        if (! in_array('minimal', $attributes->get('class')->getValue())) {
+            $attributes->add('class', 'minimal');
+        }
+
+        parent::assemble();
+    }
+}
--- a/library/Icingadb/Widget/ItemList/ServiceListItemDetailed.php
+++ b/library/Icingadb/Widget/ItemList/ServiceListItemDetailed.php
@@ -49,6 +49,19 @@ class ServiceListItemDetailed extends BaseServiceListItem
             );
         }
 
+        if ($this->item->state->is_flapping) {
+            $statusIcons->addHtml(new Icon(
+                'random',
+                [
+                    'title' => sprintf(
+                        t('Service "%s" on "%s" is in flapping state'),
+                        $this->item->display_name,
+                        $this->item->host->display_name
+                    ),
+                ]
+            ));
+        }
+
         if (! $this->item->notifications_enabled) {
             $statusIcons->addHtml(new Icon('bell-slash', ['title' => t('Notifications disabled')]));
         }
--- a/public/css/widget/state-change.less
+++ b/public/css/widget/state-change.less
@@ -24,6 +24,13 @@
     }
   }
 
+  .ball-size-l ~ .ball-size-ml {
+    &.ball-size-ml {
+      margin-top: .26em;
+      margin-left: -.375em;
+    }
+  }
+
   .state-ball.state-ok,
   .state-ball.state-up {
     &.ball-size-l,
--- a/application/controllers/HostController.php
+++ b/application/controllers/HostController.php
@@ -68,7 +68,7 @@ class HostController extends Controller
         }
 
         $this->addControl((new HostList([$this->host]))
-            ->setViewMode('minimal')
+            ->setViewMode('objectHeader')
             ->setDetailActionsDisabled()
             ->setNoSubjectLink());
         $this->addControl(new HostMetaInfo($this->host));
@@ -89,6 +89,7 @@ class HostController extends Controller
         }
 
         $this->addControl((new HostList([$this->host]))
+            ->setViewMode('objectHeader')
             ->setDetailActionsDisabled()
             ->setNoSubjectLink());
         $this->addContent(new HostInspectionDetail(
@@ -149,7 +150,7 @@ class HostController extends Controller
         yield $this->export($history);
 
         $this->addControl((new HostList([$this->host]))
-            ->setViewMode('minimal')
+            ->setViewMode('objectHeader')
             ->setDetailActionsDisabled()
             ->setNoSubjectLink());
         $this->addControl($sortControl);
@@ -213,7 +214,7 @@ class HostController extends Controller
             ->setViewMode($viewModeSwitcher->getViewMode());
 
         $this->addControl((new HostList([$this->host]))
-            ->setViewMode('minimal')
+            ->setViewMode('objectHeader')
             ->setDetailActionsDisabled()
             ->setNoSubjectLink());
         $this->addControl($paginationControl);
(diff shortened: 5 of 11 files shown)


Yonas-net:
I think using `previous_soft_state` is wrong here. @nilmerg correct me if I am wrong, but it should be the `previous_hard_state` that is shown as the state change and not the soft state. Since I now display the `previous_hard_state` in the object detail as well, there' a discrepancy between the latest object history and object detail header.

https://github.com/Icinga/icingadb-web/blob/f39fc669f41ffb949c166a2b8e92ec31c28710e7/library/Icingadb/Widget/ItemList/BaseHistoryListItem.php#L195-L196

nilmerg:
Since there are not only hard state changes but also soft state changes, I'd reckon that showing `previous_soft_state` for the latter is correct. But for hard state changes the `previous_hard_state` should be shown. I guess this was also intended, as it's a variable which is at the moment redundant since it's not *variable*.

Yonas-net:
I.e. if the current state type is soft, the `previous_hard_state` will be also shown and when the current state type is hard state, the `previous_soft_state` or do I got it wrong again? Because it doesn't make any sense if we would also show the `previous_soft_state` for soft state changes and the `previous_hard_state` for hard state changes.

nilmerg:
For soft state changes the `previous_soft_state`. For hard state changes the `previous_hard_state`.

library/Icingadb/Widget/ItemList/ServiceListItemDetailed.php
```diff
@@ -49,6 +49,16 @@ protected function assembleFooter(BaseHtmlElement $footer)
             );
         }
 
+        if ($this->item->state->is_flapping) {
+            $statusIcons->addHtml(new Icon(
+                'random',
+                [
+                    'class' => 'flapping-icon',
+                    'title' => sprintf(t('Service "%s" is in flapping state'), $this->item->display_name)
```
sukhwinder33445, line 57: Please remove the name. `Service is in flapping state` is sufficient.

library/Icingadb/Widget/ItemList/HostListItemDetailed.php
```diff
@@ -49,6 +49,16 @@ protected function assembleFooter(BaseHtmlElement $footer)
             );
         }
 
+        if ($this->item->state->is_flapping) {
+            $statusIcons->addHtml(new Icon(
+                'random',
+                [
+                    'class' => 'flapping-icon',
+                    'title' => sprintf(t('Host "%s" is in flapping state'), $this->item->display_name)
```
sukhwinder33445, line 57: Please remove the name. `Host is in flapping state` is sufficient.

public/css/widget/state-change.less
```diff
@@ -15,7 +15,7 @@
     }
 
     &.ball-size-m {
-      margin-left: -.275em;
+      margin: .3em 0 0 -.275em;
```
sukhwinder33445, line 18: Please change this to `margin-top: 0.3em;` margin-left: -0.275em;

library/Icingadb/Widget/ItemList/ServiceListItemDetailed.php
nilmerg: Not quite.. I guess Yonas added the name because of accessibility. I gave him a task to [work out something in this regard](https://github.com/Icinga/icingadb-web/blob/accessibility-tests/test/TODOS.md) a few weeks ago. Titles should not be ambiguous but should contain the relevant information themselves they're associated with. A screen reader user won't see the visual context with the service and only gets the title read out. Providing info about which host or service this is again, makes it easier.

I'd even go further and suggest to show `Service "<name>" on "<host-name>" is in flapping state`.

sukhwinder33445: I see. But what if the name of the service or host is a bit long? The main statement will be at the end of a long sentence.

nilmerg: :man_shrugging: We have no layout constraints here.

public/css/widget/state-change.less
```diff
@@ -32,3 +32,12 @@
     }
   }
 }
+
+.state-change.detail-history {
+  .state-ball.ball-size-m {
+    height: 1em;
+    width: 1em;
+    margin-top: .26em;
+    margin-left: -.375em;
```
sukhwinder33445, line 41: Why this? `ball-size-m` already has a `margin-left: -.275em`. Isn't that enough? Disabling this in the dev tools changes nothing for me.

library/Icingadb/Widget/ItemList/BaseHostListItem.php
```diff
@@ -42,6 +42,7 @@ protected function init()
 
         if ($this->list->getNoSubjectLink()) {
             $this->setNoSubjectLink();
+            $this->isDetailView = true;
```
sukhwinder33445, line 45: I am not sure if `getNoSubjectLink()` always means that it is a detail view. Maybe it would be better to add a method like isDetailView().

```
$this->addControl((new HostList([$this->host]))
            ->setViewMode('minimal')
            ->setDetailActionsDisabled()
            ->setNoSubjectLink()
            ->isDetailView());
```

nilmerg: You're correct with the assumption that `getNoSubjectLink()` is no reliable indicator if it's a detail view where this is shown.

But another modifier is not desirable in my opinion. I'm already no real fan of the fact that the detail view uses a state list in the header, that should be a dedicated widget. Then we wouldn't have this discussion or any of the existing modifiers.

Though, creating the widget here is again out of scope, but still, I dislike an additional modifier. Use a new view mode that inherits from the minimal one instead. There override `getStateBallSize()` to choose the proper size and also override `assembleVisual` where in you call the super method first and then fetch the state ball using `HtmlDocument::first()` and register the div with the previous state as wrapper for it.

public/css/widget/state-change.less
Yonas-net: You need to take a closer look at the state balls for soft state change, then it should be clear why it is needed.

Bildschirmfoto 2021-10-01 um 10 11 12 Bildschirmfoto 2021-10-01 um 10 13 25

public/css/widget/state-change.less
```diff
@@ -32,3 +32,12 @@
     }
   }
 }
+
+.state-change.detail-history {
+  .state-ball.ball-size-m {
```
nilmerg, line 37: Please don't use a special class here. Combine it with the selector `.ball-size-l ~ .ball-size-ml` (`ml` is from https://github.com/Icinga/ipl-web/pull/45)

public/css/list/list-item.less
```diff
@@ -121,6 +121,10 @@
     .status-icons {
       display: flex;
       align-items: center;
+
+      .flapping-icon:not(:first-child) {
+        margin-left: .3em;
```
nilmerg, line 126: Please remove this. It looks bad and unnecessary:
Screenshot from 2021-10-01 14-23-46

library/Icingadb/Widget/ItemList/HostListItemDetailed.php
```diff
@@ -49,6 +49,16 @@ protected function assembleFooter(BaseHtmlElement $footer)
             );
         }
 
+        if ($this->item->state->is_flapping) {
+            $statusIcons->addHtml(new Icon(
+                'random',
+                [
+                    'class' => 'flapping-icon',
+                    'title' => sprintf(t('Host "%s" is in flapping state')),
```
nilmerg, line 57: The name is missing.

library/Icingadb/Widget/ItemList/HostListItemDetailed.php
```diff
@@ -49,6 +49,16 @@ protected function assembleFooter(BaseHtmlElement $footer)
             );
         }
 
+        if ($this->item->state->is_flapping) {
+            $statusIcons->addHtml(new Icon(
+                'random',
+                [
+                    'class' => 'flapping-icon',
```
nilmerg, line 56: The class should be obsolete.

library/Icingadb/Widget/ItemList/ServiceListItemDetailed.php
```diff
@@ -49,6 +49,20 @@ protected function assembleFooter(BaseHtmlElement $footer)
             );
         }
 
+        if ($this->item->state->is_flapping) {
+            $statusIcons->addHtml(new Icon(
+                'random',
+                [
+                    'class' => 'flapping-icon',
```
nilmerg, line 56: The class should be obsolete.

nilmerg (changes requested):

library/Icingadb/Widget/ItemList/ServiceDetailHeader.php
```diff
@@ -0,0 +1,40 @@
...
+        $isSoftState = false;
+        if ($this->state->state_type === 'soft') {
+            $isSoftState = true;
+        }
+
+        // When the current state type is a soft state change, then use the actual hard_state and not the prev. ones
+        $previousState = $isSoftState ? $this->state->hard_state : $this->state->previous_hard_state;
+        $previousHardState = ServiceStates::text($previousState);
+
+        $visual->getFirst('span')->addWrapper(HtmlElement::create('div', ['class' => 'state-change'], [
```
nilmerg, line 36: Here `::create`, in the host item class `new HtmlElement`? :thinking:

library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -5,7 +5,10 @@
 namespace Icinga\Module\Icingadb\Widget\ItemList;
 
 use Icinga\Module\Icingadb\Common\BaseListItem;
+use Icinga\Module\Icingadb\Common\HostStates;
```
nilmerg, line 8: Please remove the now unused usages.

public/css/widget/state-change.less
```diff
@@ -18,6 +18,11 @@
       margin-left: -.275em;
     }
 
+    &.ball-size-ml {
```
nilmerg, line 21: The rules here all depend upon the fact that the two balls are of the same size. That's why I suggested to add a new rule that also checks the left ball size.

nilmerg (changes requested):

public/css/widget/state-change.less
Yonas-net: It doesn't matter how diverse the state ball sizes are, the rule looks like this and should be sufficient, I think.
`.state-ball ~ .state-ball`

nilmerg: Well, fine, if you think so, combine m + ml for a moment :wink:

nilmerg: or heck, even ml + ml

nilmerg (changes requested):
You'll need to also combine the css classes of the new view mode. It should not just be `objectHeader`, but also `minimal` be applied.

Screenshot from 2021-10-01 16-33-14

nilmerg (dismissed):

nilmerg (approved):

PR #377, opened 2021-09-27
css: Always use 2px height for the downtime progress bar

### LINKED ISSUE #263: Downtime progress bar has different heights
## Describe the bug
The downtimes have different heights for the progress bar and there is no space between the visuals.

## Screenshot

Screenshot 2021-08-24 at 13 22 53

## Reproduce
Add some downtimes to a host/service.

nilmerg:
fixes #263


flourish86 (dismissed):
Seems to be fixed.

Tested on Edge, Safari, Chrome/FF (Mac/Win).

public/css/list/downtime-list.less
```diff
@@ -24,9 +24,15 @@
...
-    background-color: @color-ok;
-    color: @text-color-on-icinga-blue;
+  &.active {
+    .visual {
+      background-color: @color-ok;
+      color: @text-color-on-icinga-blue;
+    }
+
+    .main {
+      padding-top: 0; // If active the progress bar represents the padding top
```
flourish86, line 34: I'm not sure, why you did this. But sadly this has the following effect.

Screen Recording 2021-09-27 at 14 27 58

The `.active` class is for selected `.list-item`s not active downtime items

nilmerg: Yeesss, it collides with the action list behavior. There `active` is also the class used to indicate which list item is active :roll_eyes:

sukhwinder33445 (commented):
It works fine now, except when zoomed to `80%` in Chrome.
Screenshot 2021-09-29 at 09 53 32

flourish86:
> It works fine now, except when zoomed to `80%` in Chrome.

From my side, I'm fine with accepting the 80% zoom Chrome inconsistency, mainly because I've run out of ideas on how to fix it. Secondly because I think Chrome will improve on this over time.

flourish86 (approved):

nilmerg:
Yep, I'm also fine with this for the same reasons. We cannot fix every single small weirdness in every single edge case. 99% of users are *probably* using 90% to 125% zoom anyway.

PR #387, opened 2021-09-30
Add default path to icon img if no relative path has been specified

### LINKED ISSUE #388: Host/Service icon image has no default path
## Describe the bug
No default path is specified yet. If no relative path is used, the host/service cannot display an icon image.

## Expected behavior
If no relative path is given, the default path should be used.

sukhwinder33445:
fix #388

### DIFF
--- a/library/Icingadb/Widget/ItemList/StateListItem.php
+++ b/library/Icingadb/Widget/ItemList/StateListItem.php
@@ -60,6 +60,11 @@ abstract class StateListItem extends BaseListItem
     {
         if (isset($this->item->icon_image->icon_image)) {
             $src = $this->item->icon_image->icon_image;
+
+            if (strpos($src, '/') === false) {
+                $src = 'img/icons/' . $src;
+            }
+
             if (getenv('ICINGAWEB_EXPORT_FORMAT') === 'pdf') {
                 $srcUrl = Url::fromPath($src);
                 $path = $srcUrl->getRelativeUrl();


library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -60,6 +63,11 @@ protected function assembleIconImage(BaseHtmlElement $iconImage)
     {
         if (isset($this->item->icon_image->icon_image)) {
             $src = $this->item->icon_image->icon_image;
+
+            if (strpos($src, '/') === false) {
+                $src = (string) Url::fromPath(static::DEFAULT_ICON_PATH . $src);
```
nilmerg, line 68: Don't use Url::fromPath, it doesn't have to be an absolute path, that's done by the browser.

nilmerg (changes requested):

library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -28,6 +28,9 @@ abstract class StateListItem extends BaseListItem
     /** @var State The state of the item */
     protected $state;
 
+    /** @var string Default path of icon image */
+    const DEFAULT_ICON_PATH = 'img/icons/';
```
nilmerg, line 32: Use constants with care. Is it a value specific to this widget? No, it is specific to Icinga Web 2.

So we have a publicly accessible constant here, that is related to the framework, not even the module.
Anyone using the constant from the outside will do so only if using a `StateListItem` anyway for some reason. (hopefully) But there's literally zero reason why such users should then require it to begin with. Icon images are already supported.

tl;dr, Don't use a constant for this. This constant would fit in the `Environment` class of Icinga Web 2 for example, but we're in the module here so a hard coded string suffices.

nilmerg (changes requested):

library/Icingadb/Widget/ItemList/StateListItem.php
sukhwinder33445: I see, thanks for explanation. :)

nilmerg (approved):

PR #393, opened 2021-10-11
Vertical list show more redesign

flourish86:


nilmerg (changes requested):
Please restore the anchor navigation. With https://github.com/Icinga/icingaweb2/pull/4581 and the `data-icinga-no-scroll-on-focus` attribute on the page separator's anchors the view shouldn't move still.

PR #408, opened 2021-10-27
Redesign migrate popup

### LINKED ISSUE #407: Redesign migrate popup
After thinking some time about it, I'm really struggling with some design issues of this popup. It looks pretty much out of place with the rest of the UI. I think we should fix that.

* There are some odd font choices, we should use 14px or even 12px as a base size as it is in the rest of the UI
* Same for line height. It is currently at 1, so it looks cramped
* It's not obvious that the items, like "Service Problems" are clickable. I see that we're using bold "links" for list item titles, but they don't look anything like those. Even if they're technically `button`s, they behave like links, so they applying the known design would make it much clearer, what they do.
* There're some odd padding choices
* ~~I know, this is a small thing, but I find "Preview *this* …" pretty casual. "Preview this *view* …" would be more explicit and clear.~~
* ~~We avoided animations and transitions ever since in the UI, it would make it more coherent if we also did for this.~~

Considering these an improved version could look like this.
Screen Shot 2021-10-21 at 14 04 54

_Originally posted by @flourish86 in https://github.com/Icinga/icingadb-web/issues/395#issuecomment-948548351_

sukhwinder33445:
resolves #407

### DIFF
--- a/public/css/widget/migrate-popup.less
+++ b/public/css/widget/migrate-popup.less
@@ -8,6 +8,7 @@
   top: 0;
   right: 4em;
   pointer-events: none;
+  line-height: 1.5em;
 
   .transform(translateY(-100%));
   .transition(transform @transitionLength ease-in);
@@ -75,9 +76,10 @@
   }
 
   .suggestion-area {
-    padding: .75em .75em .25em;
+    padding: .75em;
     flex-grow: 1;
     pointer-events: auto;
+    font-size: .75em;
 
     .var(background, body-bg-color);
     border-radius: 0 0 4px 4px;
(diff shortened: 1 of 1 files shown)


public/css/widget/migrate-popup.less
```diff
@@ -8,6 +8,8 @@
   top: 0;
   right: 4em;
   pointer-events: none;
+  line-height: 1.5em;
+  font-size: .875em;
```
nilmerg, line 12: Why is that on the entire popup? This way it has an effect not only on font-size, but on the entire layout of the popup. I thought it's only the text that's too big.

Currently I have difficulty too read the "Don't show this again" label. It's way too small.

nilmerg (changes requested):

nilmerg:
@flourish86 Please take a look now.

flourish86 (approved):
Looks fine now. Thanks!

nilmerg (approved):

PR #409, opened 2021-10-27
Bugfix/mobile issues 401

### LINKED ISSUE #401: Mobile Issues
- [x] Title attributes are not accessible on touch devices
- [x] Some lists won’t ellipsisize correctly: Downtimes
- [x] Some lists won’t ellipsisize correctly: Comments
- [x] Some lists won’t ellipsisize correctly: Notifications
- [x] Some lists won’t ellipsisize correctly: History
- [x] Health view labels should handle line break better
- [x] Tactical Overview Cards look better in full width on mobile
- [x] See [icingaweb2/#4562](https://github.com/Icinga/icingaweb2/issues/4562#issue-1030297058)

I want to collect some of the issues I found, when using Icinga DB Web from a mobile device.

### Title attributes are not accessible on touch devices
We use title attributes to reveal ellipsisized text. This is especially critical for ellipsisized list item titles.
We could have a workaround for this in form of a tool tip, that appears on taphold, like outlined [here](https://stackoverflow.com/questions/38010965/mobile-web-design-alternative-to-title-atribute-in-html-tags).

### Some lists won’t ellipsisize correctly
Downtimes
Screen Shot 2021-10-19 at 14 36 49

Comments
Screen Shot 2021-10-19 at 14 39 19

Notifications
Screen Shot 2021-10-19 at 14 40 42

History
Screen Shot 2021-10-19 at 14 59 36

### Health view labels should handle line break better

Screen Shot 2021-10-19 at 14 59 59

### Tactical Overview Cards look better in full width on mobile
Like this
Screen Shot 2021-10-19 at 15 01 16

flourish86:
fixes #401

### DIFF
--- a/library/Icingadb/Common/HostLink.php
+++ b/library/Icingadb/Common/HostLink.php
@@ -22,6 +22,6 @@ trait HostLink
 
         $content[] = $host->display_name;
 
-        return Html::tag('a', ['href' => Links::host($host)], $content);
+        return Html::tag('a', ['href' => Links::host($host), 'class' => 'subject'], $content);
     }
 }
--- a/library/Icingadb/Common/ServiceLink.php
+++ b/library/Icingadb/Common/ServiceLink.php
@@ -25,10 +25,10 @@ trait ServiceLink
 
         return Html::sprintf(
             t('%s on %s', '<service> on <host>'),
-            Html::tag('a', ['href' => Links::service($service, $host)], $content),
+            Html::tag('a', ['href' => Links::service($service, $host), 'class' => 'subject'], $content),
             Html::tag(
                 'a',
-                ['href' => Links::host($host)],
+                ['href' => Links::host($host), 'class' => 'subject'],
                 [
                     new StateBall($host->state->getStateText(), StateBall::SIZE_MEDIUM),
                     ' ',
--- a/public/css/list/comment-list.less
+++ b/public/css/list/comment-list.less
@@ -13,7 +13,6 @@
 
   .title a {
     &.subject {
-      overflow: visible;
       margin-right: 0;
     }
 
--- a/public/css/widget/donut-container.less
+++ b/public/css/widget/donut-container.less
@@ -11,5 +11,14 @@
 
   &:not(:last-of-type) {
     margin-right: 1em;
+    margin-bottom: 1em;
   }
+
+  .donut {
+    margin: 0 auto;
+  }
+}
+
+#layout.minimal-layout .donut-container {
+  width: 100%;
 }
--- a/public/css/widget/monitoring-health.less
+++ b/public/css/widget/monitoring-health.less
@@ -11,6 +11,8 @@
   }
 
   .check-summary {
+    padding: .5em 0;
+
     .col {
       padding: 0 1em .5em 1em;
     }
@@ -27,7 +29,6 @@
     border: 1px solid;
     .var(border-color, gray-light);
     display: flex;
-    padding: .5em 0;
     width: 100%;
   }
 
@@ -73,7 +74,7 @@
   .instance-features {
     .control-group {
       flex: 1 1 auto;
-      margin: 0;
+      margin: .5em 0;
       padding: .5em;
       width: 0;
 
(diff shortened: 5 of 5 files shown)


library/Icingadb/Common/HostLink.php
```diff
@@ -11,7 +11,7 @@
 
 trait HostLink
 {
-    protected function createHostLink(Host $host, bool $withStateBall = false): BaseHtmlElement
+    protected function createHostLink(Host $host, bool $withStateBall = false, array $atts = null): FormattedString
```
nilmerg, line 14: This doesn't return a `FormattedString`.

public/css/list/comment-list.less
```diff
@@ -13,7 +13,6 @@
 
   .title a {
     &.subject {
-      overflow: visible;
```
nilmerg, line 16: Not sure about this. I've added this as I thought the username is the most important thing here. See:

**with overflow visible**
Screenshot from 2021-11-04 09-40-40

**without**
Screenshot from 2021-11-04 09-40-50

Currently only the *commented* part is not visible, but with long usernames I believe they'll also get compressed too far so that they are not readable anymore. And I think those are more important than the object names, aren't they?

But yes, I know, veeeeery long usernames then cause overflow again or get cut...

public/css/widget/donut-container.less
```diff
@@ -12,4 +14,8 @@
   &:not(:last-of-type) {
     margin-right: 1em;
   }
+
+  .donut {
+    margin: 0 auto;
```
nilmerg, line 19: What effect has this? I can't see one.

public/css/widget/donut-container.less
```diff
@@ -1,6 +1,8 @@
 .donut-container {
   .card();
 
+  margin-bottom: 1em;
```
nilmerg, line 4: Please add this to the same rule that also applies the right margin.

nilmerg (changes requested):
Since there are no usages of `HostLink` or `ServiceLink` that don't apply class `subject`, I think both traits can apply this on their own. So there's no need for a new parameter there.

public/css/widget/donut-container.less
flourish86: ✅

public/css/widget/donut-container.less
flourish86: Makes sense, this was missing:
https://github.com/Icinga/icingadb-web/blob/0c2b7b9e74fadea3a96c9f76fd8e14b8802a4cf2/public/css/widget/donut-container.less#L22

See the difference:
Desktop Copy 4

flourish86: ✅

public/css/list/comment-list.less
flourish86: Your example is fine. But I wouldn’t consider this as accaptable.

Screen Shot 2021-11-04 at 13 13 18

I suspect there won’t be an elegant CSS solution because `text-ellipsis` is very limited. As discussed we should rely on the detailed view mode to handle this.

The only alternative I see is having the title break in normal view-mode as well in minimal layout, but this would make the detailed view mode almost obsolete.

nilmerg: Nah, as said, I'm not sure and don't have an alternative solution either..

library/Icingadb/Common/HostLink.php
```diff
@@ -11,7 +11,7 @@
 
 trait HostLink
 {
-    protected function createHostLink(Host $host, bool $withStateBall = false): BaseHtmlElement
```
nilmerg, line 14: That's not intended, I guess?

library/Icingadb/Common/ServiceLink.php
```diff
@@ -12,7 +12,7 @@
 
 trait ServiceLink
 {
-    protected function createServiceLink(Service $service, Host $host, bool $withStateBall = false): FormattedString
```
nilmerg, line 15: Again, not intended?

nilmerg (changes requested):

library/Icingadb/Common/ServiceLink.php
flourish86: ✅

nilmerg (approved):

PR #443, opened 2021-11-11
comment-list: Add missing space before "on" in title

### LINKED ISSUE #438: Comment list: space is missing in subject
## Screenshots
Screenshot 2021-11-10 at 11 07 17

flourish86:
fixes #438

### DIFF
--- a/library/Icingadb/Web/Navigation/IcingadbHostAction.php
+++ b/library/Icingadb/Web/Navigation/IcingadbHostAction.php
@@ -6,5 +6,4 @@ namespace Icinga\Module\Icingadb\Web\Navigation;
 
 class IcingadbHostAction extends Action
 {
-
 }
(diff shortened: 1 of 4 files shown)


flourish86:
There were also spacing issues in titles in detailed view mode, which is fixed by 7002703. I think, there’s a better solution, though.

It’s actually awkward, because the .title element is `display: block` in detailed view mode list elements. This is why there should be spaces added automatically (the child elements of `.title` are `display: inline`). Maybe there are spaces missing in the markup, so the elements “collapse”.

This is how i looked with 10a0c7f.

Screen Shot 2021-11-11 at 17 11 08

nilmerg:
It's `display:block` in detailed mode, to disable the overflow mechanics. There were these `<br>` previously, but I got rid of them due to being semantically incorrect.

The current solution though still doesn't work for all cases. See below:

Screenshot from 2021-11-12 11-18-04

The first badge has too much margin-left. For service comments, the last badge if there are more is missing margin-right.

nilmerg (approved):

nilmerg:
Please squash all but the last commit.

nilmerg (approved):

PR #452, opened 2021-11-12
Fix incorrect assignment of class handled to the service badges in service grid.

### LINKED ISSUE #451: All services in service grid have handled problem badges.
## Describe the bug
The services in service grid have handled problem badges even though they are not handled.

## Screenshots
Screenshot 2021-11-12 at 15 11 32

raviks789:
fix #451


nilmerg (approved):

nilmerg:
There's actually something else wrong: Behaviors like `BoolCast` don't run for aliased columns. But that's something for later [in a separate issue](https://github.com/Icinga/ipl-orm/issues/33).

PR #454, opened 2021-11-12
Use display names as axis headers for hosts and services in service grid.

### LINKED ISSUE #453: Services and hosts must use display name in service grid
## Describe the bug
The services and hosts in service grid are must be using their display name instead of service.name and host.name respectively.

## Screenshots
Screenshot 2021-11-12 at 15 24 27

raviks789:
The x- and y-axis column headers and sort control in service grid must use display names for hosts and services.

fix #453

### DIFF
--- a/application/controllers/ServicesController.php
+++ b/application/controllers/ServicesController.php
@@ -228,9 +228,9 @@ class ServicesController extends Controller
 
         $problemToggle = $this->createProblemToggle();
         $sortControl = $this->createSortControl($query, [
-            'service.name' => t('Service Name'),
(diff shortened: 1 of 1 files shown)


nilmerg (changes requested):
The original grid differentiates between `host_name` and `host_display_name`. It uses the latter only for labeling the headers. I don't quite know for what the first is being used or required, but that doesn't matter right know I believe, since it seems to work fine there.

You have the option as well to differentiate, please do so.

nilmerg (approved):

PR #455, opened 2021-11-15
Fix redirect url to keep the active state of the selected objects

### LINKED ISSUE #441: Multi-select highlight for hosts disappear after feature commands or enabled or disabled
## Describe the bug
Multi-select highlighting of selected hosts disappear after an action is taken on feature commands. This happens only in the case of hosts, in case of services this was not encountered.

## Screenshots

https://user-images.githubusercontent.com/33730024/141297230-43f68205-46c6-46c2-8aae-61f4e4cc8cdb.mov

sukhwinder33445:
fixes #441

### DIFF
--- a/application/controllers/HostsController.php
+++ b/application/controllers/HostsController.php
@@ -216,7 +216,17 @@ class HostsController extends Controller
 
     protected function getCommandTargetsUrl(): Url
     {
-        return Links::hostsDetails()->setQueryString(QueryString::render($this->getFilter()));
+        $queryString = '';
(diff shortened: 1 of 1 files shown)


sukhwinder33445:
$this->getFilter()` does not contain the required round brackets. This leads to the fact that after the redirection the url is without brackets and active elements cannot be identified.

nilmerg:
The thing is, I cannot identify a reason for these additional parentheses anymore. So I removed the need for them in #484, which replaces this.

PR #457, opened 2021-11-15
Use font-awesome icons if icon_image exists without extension

### LINKED ISSUE #456: Host icons for Icinga Web internal icons not showing
## Describe the bug
Icingaweb internal icons not showing for hosts (eg print) external images work fine (eg. tux.png).

## To Reproduce

1. Configure a Icingaweb2 internal image as hostimage. (eg. print)

2. Show the host in std. host view:
grafik

3. Show the host in Icinga DB view:
grafik

## Expected behavior
The internal icon should be visible in both cases.

## Screenshots
see above

## Your Environment

* Icinga DB Web version (System - About): 1.0.0-rc2 (f402e1e5dc4ffde136ab0fed109925f2a754ca86)
* Icinga Web 2 version (System - About): 2.9.4 (d208d1ccfef86606984739e6357138c459146134)
* Web browser: Google Chrome 95.0.4638.69 / Firefox 78.15.0esr
* Icinga 2 version (`icinga2 --version`): r2.13.2-1
* Icinga DB version (`icingadb --version`): v1.0.0-rc2-dev
* PHP version used (`php --version`): 7.3.31-2
* Server operating system and version: Debian 10 / Kernel 4.19.0-17-amd64

sukhwinder33445:
fixes #456

### DIFF
--- a/library/Icingadb/Widget/ItemList/StateListItem.php
+++ b/library/Icingadb/Widget/ItemList/StateListItem.php
@@ -61,6 +61,15 @@ abstract class StateListItem extends BaseListItem
         if (isset($this->item->icon_image->icon_image)) {
             $src = $this->item->icon_image->icon_image;
 
+            if (strpos($src, '.') === false) {
+                $iconImage->add(HtmlElement::create('i', [
+                    'class'     => 'fas fa-' . $src,
+                    'alt'       => $this->item->icon_image_alt
+                ]));
+
+                return;
+            }
+
             if (strpos($src, '/') === false) {
                 $src = 'img/icons/' . $src;
             }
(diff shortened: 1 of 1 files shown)


library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -61,6 +61,15 @@ protected function assembleIconImage(BaseHtmlElement $iconImage)
         if (isset($this->item->icon_image->icon_image)) {
             $src = $this->item->icon_image->icon_image;
 
+            if (strpos($src, '.') === false) {
+                $iconImage->add(HtmlElement::create('i', [
+                    'class'     => 'fas fa-' . $src,
+                    'alt'       => $this->item->icon_image_alt
+                ]));
```
nilmerg, line 68: There's the `Icon` class, can't you use that? Also, `icon_image_alt` is of no use here, since the `alt` property is exclusive to `img` elements. For icons it's of no use, because it's mainly used to show something in case the `src` attribute of an image cannot be loaded. With icons there is no *loading* involved here.

nilmerg (changes requested):

nilmerg (approved):

PR #458, opened 2021-11-15
ObjectDetail: Avoid using Filter::unequal() `*` to match a nullable c…

Yonas-net:
…olumn


Yonas-net:
## Blocked by

-  https://github.com/Icinga/ipl-sql/pull/46

PR #461, opened 2021-11-22
BaseHistorylistItem: Don't render OK -> OK as hard state change

### LINKED ISSUE #459: Command transport passwordfield always show the password (aka is not a password field)
## Describe the bug
Command transport passwordfield always show the password

## To Reproduce
Configure a command transport, the api password is always shown. Other password fields hide the password

## Expected behavior
The password should be also hidden like all other password fields in icingaweb2

## Screenshots
grafik

## Your Environment

* Icinga DB Web version (System - About): 1.0.0-rc2
* Icinga Web 2 version (System - About): 2.9.4
* Web browser: FireFox, Chrome, whatever
* Icinga 2 version (`icinga2 --version`): 2.13.2-1
* Icinga DB version (`icingadb --version`): v1.0.0-rc2-dev
* PHP version used (`php --version`): 8.0.12
* Server operating system and version: CentOS Linux 8

yhabteab:


yhabteab:
@cla-bot check

yhabteab:
In the object detail header will still be rendered that way, but we don't have a database column for it, like the history table at the moment.

yhabteab:
Bildschirmfoto 2021-12-09 um 11 03 17

nilmerg (changes requested):
An ok state is always a hard state so there's no way that a *soft state* may be ok. IMHO we should only render the previous soft state, ever. This has also been the case in the first place before ac13e104bc3b71e235116ecf30981c070b5b8085.

nilmerg:
Superseded by #492

PR #465, opened 2021-11-30
ApiTransportForm: Fix force creation of a transport doesn't work

yhabteab:


nilmerg:
Shouldn't the check on line no. 72 be adjusted to check for `y` instead `n`? If checked, the connection attempt is useless after all.

yhabteab:
> Shouldn't the check on line no. 72 be adjusted to check for `y` instead `n`? If checked, the connection attempt is useless after all.

I think this is wrong. This way the password would still be updated even if the checkbox is not enabled.

yhabteab:
Never mind!

nilmerg (approved):

PR #469, opened 2021-12-08
Provide hooks for icingadb support in icinga-cube-module

raviks789:
refs: Icinga/icingaweb2-module-cube#64


nilmerg:
Closing here for now. Branch can be restored if needed.

PR #477, opened 2022-01-13
Fix that invalid perfdata breaks detail views

### LINKED ISSUE #475: An equal sign in a perfdata label breaks an objects entire view
## Describe the bug

```
Uncaught TypeError: number_format(): Argument #1 ($num) must be of type float, string given in /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Util/PerfData.php:552
Stack trace:
#0 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Util/PerfData.php(552): number_format('"foo">FOO</span...', 2)
#1 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Util/PerfData.php(577): Icinga\Module\Icingadb\Util\PerfData->format('"foo">FOO</span...')
#2 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Widget/Detail/PerfDataTable.php(78): Icinga\Module\Icingadb\Util\PerfData->toArray()
#3 /usr/local/src/ipl-html/src/HtmlDocument.php(295): Icinga\Module\Icingadb\Widget\Detail\PerfDataTable->assemble()
```

## To Reproduce

1. Use "Process check result"
2. Define perfdata with an equal sign in the label (e.g. `'a=b'=0%;;2`)
3. Submit

## Expected behavior
The view should not break, but should ensure the perfdata is still rendered somehow. Even if wrong.

nilmerg:
fixes #475

### DIFF
--- a/library/Icingadb/Util/PerfData.php
+++ b/library/Icingadb/Util/PerfData.php
@@ -548,6 +548,8 @@ class PerfData
                 return PerfDataFormat::grams($value);
             case $this->isLiters():
                 return PerfDataFormat::liters($value);
+            case ! is_numeric($value):
+                return $value;
             default:
                 return number_format($value, 2);
         }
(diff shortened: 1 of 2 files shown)


sukhwinder33445 (approved):
Works fine.

Issue #472, opened 2021-12-11
Add support for redis password


flybyray:
Currently it is not possible to use redis password.

Some other icinga components would support that: https://github.com/Icinga/docker-icingadb/blame/master/README.md#L30

https://github.com/Icinga/icingadb-web/blob/2ffd820a74cbd70b5e385db369174647eafc2f00/library/Icingadb/Common/IcingaRedis.php#L156

nilmerg:
[Icinga DB](https://github.com/Icinga/icingadb/blob/master/doc/03-Configuration.md) itself doesn't document this. Redis v6 though, which is required anyway, supports passwords in [protected mode](https://redis.io/topics/security). @Al2Klimov Can I safely assume it is *safe* to implement support for that?

Al2Klimov:
Why.. not?

lippserd:
> Why.. not?

Yes it is safe. The only thing missing is the documentation.

PR #496, opened 2022-02-21
Feature/allow redis password

verboEse:
Add authentication against Redis with password.

This should solve #472.

Known error: After validation the password field is empty. Maybe you can push me to how I can fill it.


verboEse:
I just added the missing mail address to my account, @cla-bot should verify now correctly.

nilmerg:
@cla-bot check

bobapple:
@cla-bot check

library/Icingadb/Setup/RedisStep.php
```diff
@@ -117,6 +119,10 @@ public function getSummary()
             new HtmlElement('strong', null, Text::create(t('Port'))),
             $this->data['redis1_port'] ?: 6380
         ]));
+        $primaryOptions->addHtml(Table::row([
+            new HtmlElement('strong', null, Text::create(t('Password'))),
+            $this->data['redis1_password'] ?: 6380
```
nilmerg, line 124: Use `mt('icingadb', 'None', 'non-existence of a value')` as default here.

library/Icingadb/Setup/RedisStep.php
```diff
@@ -133,6 +139,10 @@ public function getSummary()
                 new HtmlElement('strong', null, Text::create(t('Port'))),
                 $this->data['redis2_port'] ?: 6380
             ]));
+            $secondaryOptions->addHtml(Table::row([
+                new HtmlElement('strong', null, Text::create(t('Password'))),
+                $this->data['redis2_password'] ?: 6380
```
nilmerg, line 144: Use `mt('icingadb', 'None', 'non-existence of a value')` as default here.

nilmerg (changes requested):
> Known error: After validation the password field is empty. Maybe you can push me to how I can fill it.

Define `'renderPassword' => true` as option.

library/Icingadb/Setup/RedisStep.php
verboEse: changed this, too.

verboEse:
I added the 'non existence' part for empty password values, but I still don't see how to re-fill the field after validation. `'renderPassword' => true` makes no difference?!

nilmerg:
Where have you added it? It needs to go to both password elements. It works for me this way fine.

verboEse:
You are right. I got the wrong spot for this. Now it works. Thanks for that hint, I learned ... :-)

nilmerg (approved):
Thanks!

Issue #491, opened 2021-11-03
`OK` -> `OK` is being displayed as a hard state change


yhabteab:
## Describe the bug

Icinga DB shows `OK -> OK` as hard state change for random hosts in web as well as in DB. In the database `soft_state`, `hard_state` and `previous_hard_state` have exactly the same value 0. And as `state_type` it states `hard`. Here screenshot of Icinga DB Web and monitoring module with the same host objects.

**Icinga DB Web:**
<img width="621" alt="Bildschirmfoto 2021-11-03 um 16 28 19" src="https://user-images.githubusercontent.com/57616252/140091115-3095ae8a-3f0f-4381-9211-c943b9c74e58.png">

**Monitoring:**
<img width="621" alt="Bildschirmfoto 2021-11-03 um 16 28 52" src="https://user-images.githubusercontent.com/57616252/140091211-5e2fffc3-a44f-4c5a-aa89-cd02cdf2ef98.png">

## To Reproduce

This always happens after I have Icinga 2 running for a couple of hours. It also doesn't happen with all host objects, but is just arbitrary.

PR #501, opened 2022-02-24
Utilize `previous_soft_state` column

nilmerg:
refs #492
refs #491


nilmerg:
Screenshot from 2022-02-24 17-26-28

The above looks fine, but is *probably* technically incorrect. In this case, if the previous soft state is the same as the current one, we should show the previous hard state again.

nilmerg:
Looks fine and is technically correct:
Screenshot from 2022-04-01 13-39-19

The object's header will show the same representation like the latest state change event.

PR #502, opened 2022-02-25
Fix view mode switcher shows different mode than what is used

### LINKED ISSUE #489: View mode switcher shows different mode than what is used
## Describe the bug
View mode switcher shows different mode than what is used.

## To Reproduce
Go to a view with a mode switcher, add `view=...` with a different mode to the url.

## Expected behavior
The mode switcher shows the view mode as active that is actually in use.

sukhwinder33445:
fix #489

### DIFF
--- a/library/Icingadb/Web/Control/ViewModeSwitcher.php
+++ b/library/Icingadb/Web/Control/ViewModeSwitcher.php
@@ -101,7 +101,13 @@ class ViewModeSwitcher extends Form
      */
     public function getViewMode(): string
     {
-        return $this->getValue($this->getViewModeParam(), $this->getDefaultViewMode());
+        $viewMode = $this->getPopulatedValue($this->getViewModeParam(), $this->getDefaultViewMode());
+
+        if (array_key_exists($viewMode, static::$viewModes)) {
+            return $viewMode;
+        }
+
+        return $this->getDefaultViewMode();
     }
 
     /**


library/Icingadb/Web/Controller.php
```diff
@@ -367,7 +367,7 @@ public function createViewModeSwitcher(
         LimitControl $limitControl,
         bool $verticalPagination = false
     ): ViewModeSwitcher {
-        $viewModeSwitcher = new ViewModeSwitcher();
+        $viewModeSwitcher = new ViewModeSwitcher(Url::fromRequest());
```
nilmerg, line 370: It doesn't need to parse the url to know the current view mode, it's already passed by `populate`, just not accessed internally.

nilmerg (changes requested):

library/Icingadb/Web/Control/ViewModeSwitcher.php
```diff
@@ -101,7 +101,13 @@ public function setViewModeParam(string $viewModeParam): self
      */
     public function getViewMode(): string
     {
-        return $this->getValue($this->getViewModeParam(), $this->getDefaultViewMode());
+        $viewMode =  $this->getPopulatedValue($this->getViewModeParam(), $this->getDefaultViewMode());
```
nilmerg, line 104: There's a superfluous space after the equal sign.

nilmerg (approved):

PR #517, opened 2022-04-04
Don't provide migrate.js when monitoring isn't enabled

yhabteab:


public/js/migrate.js
```diff
@@ -75,6 +75,11 @@
         var _this = event.data.self;
         var $target = $(event.target);
 
+        if (! _this.icinga.hasModule('monitoring')) {
+            // No need to render the popup if monitoring isn't loaded
```
nilmerg, line 79: I'd go even so far to say: Why loading this JS if the module isn't active? :grin:

yhabteab: However, the popup will never be rendered, even though monitoring is active! Because it can be the case that `icingadb` is loaded first, that's why I did it with `hasModule`.

yhabteab: > And that's the only thing you're interested in, right?

Exactly!!

nilmerg (approved):

PR #518, opened 2022-04-06
Auth: Force query optimization on union queries

### LINKED ISSUE #514: Servicegroups appear empty with service restrictions, even if not
## Describe the bug
A servicegroup is empty in the overview list. Clicking on it lists some of its services though.

## To Reproduce
Screenshot from 2022-03-30 09-00-12

## Screenshots
Screenshot from 2022-03-30 08-58-26

yhabteab:
Forces the given restriction/filter to be wrapped in a subquery if there is a restriction applied
to `*/services` or `*/hosts` with `hostgroup.name` or `servicegroup.name`. If `hostgroup.name` is
used for services restriction, it is automatically wrapped to a subquery, because service model
has no direct relation to `hostgroup` and it is the same with host model and `servicegroup.name`.

fixes #514

### DIFF
--- a/library/Icingadb/Common/Auth.php
+++ b/library/Icingadb/Common/Auth.php
@@ -123,6 +123,7 @@ trait Auth
             $queries = [$query];
         }
 
+        $orgQuery = $query;
         foreach ($queries as $query) {
             $relations = [$query->getModel()->getTableName()];
             foreach ($query->getWith() as $relationPath => $relation) {
@@ -169,19 +170,24 @@ trait Auth
                     }
 
                     if ($applyHostRestriction && ($restriction = $role->getRestrictions('icingadb/filter/hosts'))) {
-                        $roleFilter->add($this->parseRestriction($restriction, 'icingadb/filter/hosts'));
+                        $hostFilter = $this->parseRestriction($restriction, 'icingadb/filter/hosts');
+                        if ($orgQuery instanceof UnionQuery) {
+                            $this->forceQueryOptimization($hostFilter, 'hostgroup.name');
+                        }
+
+                        $roleFilter->add($hostFilter);
                     }
 
                     if (
                         $applyServiceRestriction
                         && ($restriction = $role->getRestrictions('icingadb/filter/services'))
                     ) {
-                        $roleFilter->add(
-                            Filter::any(
-                                Filter::unequal('service.id', '*'),
-                                $this->parseRestriction($restriction, 'icingadb/filter/services')
-                            )
-                        );
+                        $serviceFilter = $this->parseRestriction($restriction, 'icingadb/filter/services');
+                        if ($orgQuery instanceof UnionQuery) {
+                            $this->forceQueryOptimization($serviceFilter, 'servicegroup.name');
+                        }
+
+                        $roleFilter->add(Filter::any(Filter::unequal('service.id', '*'), $serviceFilter));
                     }
                 }
 
@@ -320,5 +326,28 @@ trait Auth
         }
 
         return $filter;
+    }
+
+    /**
+     * Force query optimization on the given service/host filter rule
+     *
+     * Applies forceOptimization, when the given filter rule contains the given filter column
+     *
+     * @param Filter\Rule $filterRule
+     * @param string $filterColumn
+     *
+     * @return void
+     */
+    protected function forceQueryOptimization(Filter\Rule $filterRule, string $filterColumn)
+    {
+        // TODO: This is really a very poor solution is therefore only a quick fix.
+        //  We need to somehow manage to make this more enjoyable and creative!
+        if ($filterRule instanceof Filter\Chain) {
+            foreach ($filterRule as $rule) {
+                $this->forceQueryOptimization($rule, $filterColumn);
+            }
+        } elseif ($filterRule->getColumn() === $filterColumn) {
+            $filterRule->metaData()->set('forceOptimization', true);
+        }
     }
 }


library/Icingadb/Common/Auth.php
```diff
@@ -107,20 +107,20 @@ public function isMatchedOn(string $queryString, Model $object): bool
      * fetching either hosts or services. It also applies custom variable restrictions and
      * obfuscations. (`icingadb/blacklist/variables` and `icingadb/protect/variables`)
      *
-     * @param Query $query
+     * @param Query $orgQuery
      *
      * @return void
      */
-    public function applyRestrictions(Query $query)
+    public function applyRestrictions(Query $orgQuery)
```
nilmerg, line 114: Why? You don't assign anything to $query anyway.

library/Icingadb/Common/Auth.php
```diff
@@ -169,19 +169,28 @@ public function applyRestrictions(Query $query)
                     }
 
                     if ($applyHostRestriction && ($restriction = $role->getRestrictions('icingadb/filter/hosts'))) {
-                        $roleFilter->add($this->parseRestriction($restriction, 'icingadb/filter/hosts'));
+                        $hostFilter = $this->parseRestriction($restriction, 'icingadb/filter/hosts');
+                        // TODO: This is really a very poor solution is therefore only a quick fix.
+                        //  We need to somehow manage to make this more enjoyable and creative!
+                        if ($orgQuery instanceof UnionQuery && strpos($restriction, 'hostgroup.name') !== false) {
```
nilmerg, line 175: Eeeh, no. This introduces another bug: What if the restriction is `host.name=foo|hostgroup.name=bar`? Then the `host.name=foo` filter is processed as subquery. :boom: 

You need to traverse through the filter object, I'm afraid.

nilmerg (changes requested):

library/Icingadb/Common/Auth.php
nilmerg: Or wait, no, `hostgroup.name` is processed as one, which you want to avoid, right? :thinking: Anyway, forcing optimization should be done on the individual rules, not the entire chain.

library/Icingadb/Common/Auth.php
yhabteab: Yes, but there is also something like this...
```php
foreach($queries as $query) ...
```
Then `$query` is no longer the original one.

nilmerg: I see. But then assign `$orgQuery` in the body, don't change the name of the parameter.

library/Icingadb/Common/Auth.php
```diff
@@ -321,4 +326,31 @@ protected function parseBlacklist(string $blacklist, string $column): Filter\Non
...
+    {
+        // TODO: This is really a very poor solution is therefore only a quick fix.
+        //  We need to somehow manage to make this more enjoyable and creative!
+        foreach ($filterRule as $rule) {
+            if (! $rule instanceof Filter\Condition) {
+                $this->forceQueryOptimization($rule, $model);
+                return;
+            }
+
+            $needle = $model === 'host' ? 'hostgroup.name' : 'servicegroup.name';
```
nilmerg, line 350: Can't you just pass the column name to the function instead of the model?

nilmerg (changes requested):

yhabteab:
Wait a minute just tried to restrict only with `servicegroup.name` and it doesn't work anymore.

yhabteab:
The passed `Filter\Rule` object can actually be `Filter\Condition` from initial, when you only use `servicegroup.name` or `hostgroup.name` as restriction.

library/Icingadb/Common/Auth.php
```diff
@@ -321,4 +327,38 @@ protected function parseBlacklist(string $blacklist, string $column): Filter\Non
...
+     * @param Filter\Rule $filterRule
+     * @param string $filterColumn
+     *
+     * @return void
+     */
+    protected function forceQueryOptimization(Filter\Rule $filterRule, string $filterColumn)
+    {
+        if ($filterColumn !== 'hostgroup.name' && $filterColumn !== 'servicegroup.name') {
+            return;
+        }
```
nilmerg, line 345: It's protected, and only used explicitly with hardcoded parameters. I don't think we need this.

library/Icingadb/Common/Auth.php
```diff
@@ -321,4 +327,38 @@ protected function parseBlacklist(string $blacklist, string $column): Filter\Non
 
         return $filter;
     }
+
+    /**
+     * Force query optimization on the given service/host filter rule
+     *
+     * Applies forceOptimization, when the given filter contains one of those columns(hostgroup.name,servicegroup.name)
```
nilmerg, line 334: This isn't correct anymore.

library/Icingadb/Common/Auth.php
```diff
@@ -321,4 +327,38 @@ protected function parseBlacklist(string $blacklist, string $column): Filter\Non
...
+                if (! $rule instanceof Filter\Condition) {
+                    $this->forceQueryOptimization($rule, $filterColumn);
+                    return;
+                }
+
+                if ($rule->getColumn() === $filterColumn) {
+                    $rule->metaData()->set('forceOptimization', true);
+                }
+            }
+        }
```
nilmerg, line 362: There's an easier solution for this:

```
if (rule is chain) {
    foreach (chain as subrule) {
        recursivecall(subrule, ...)
    }
} elseif (column matches) {
    forceOptimization
}

return rule;
```

nilmerg (changes requested):

nilmerg (approved):
Looks and works fine!

nilmerg:
> Should I consider to check whether the restriction contains these two filter columns at all before calling the new method? This way we could spare a bit of performance if the restriction do not contain these two filter columns.

Be careful with such considerations. I'd do that only if there's real gain. Most restrictions aren't probably that complex that we save performance. And if they're really long, a strpos also has to work more. This can only be decided with performance tests. Though, I'd classify these as equally expensive than leaving the code as-is. So.. your decision, if you have nothing else to do ;)

PR #519, opened 2022-04-07
SLA reports

lippserd:


library/Icingadb/ProvidedHook/Reporting/SlaReport.php
```diff
@@ -0,0 +1,264 @@
...
+        $average = $data->getAverages()[0];
+
+        if ($average < $threshold) {
+            $slaClass = 'nok';
+        } else {
+            $slaClass = 'ok';
+        }
+
+        $tableRows[] = Html::tag('tr', null, [
+            Html::tag('td', ['colspan' => count($data->getDimensions())], 'Total'),
```
nilmerg, line 241: `Total` should be translatable.

library/Icingadb/ProvidedHook/Reporting/SlaReport.php
```diff
@@ -0,0 +1,264 @@
...
+            'label' => t('Filter')
+        ]);
+
+        $form->addElement('select', 'breakdown', [
+            'label'   => t('Breakdown'),
+            'options' => [
+                'none'  => 'None',
+                'day'   => 'Day',
+                'week'  => 'Week',
+                'month' => 'Month'
```
nilmerg, line 170: These should be translatable. Not that for `None` you should pass a context, as otherwise it's ambiguous.

nilmerg (changes requested):
* There's [this](https://github.com/Icinga/icingaweb2-module-idoreports/issues/54) request for idoreports. Seems like a quick win to me to support this from the get-go here.
* Also [this](https://github.com/Icinga/icingaweb2-module-idoreports/issues/26)?

* The report names shouldn't include the prefix `Icinga DB` if the idoreports hooks are not available. We do it the same for actions already.

library/Icingadb/ProvidedHook/Reporting/HostSlaReport.php
```diff
@@ -0,0 +1,68 @@
...
+use ipl\Stdlib\Filter\Rule;
+
+use function ipl\I18n\t;
+
+class HostSlaReport extends SlaReport
+{
+    public function getName()
+    {
+        $name = t('Host SLA Report');
+        if (Icinga::app()->getModuleManager()->hasEnabled('monitoring')) {
```
nilmerg, line 22: I feel better knowing this checks for the `idoreports` module. One could still have it enabled despite having monitoring disabled. Then there are two `Host SLA Report` entries.

library/Icingadb/ProvidedHook/Reporting/SlaReport.php
```diff
@@ -206,6 +207,11 @@ public function getHtml(Timerange $timerange, array $config = null)
         }
 
         $tableRows = [];
+        $configs = Config::fromIni(Config::resolvePath('modules/reporting/config.ini'));
```
nilmerg, line 210: Huh? Report configuration is stored in database. That's controlled by `initConfigForm()`. Any element added there is available in `$config` afterwards.

library/Icingadb/ProvidedHook/Reporting/SlaReport.php
```diff
@@ -237,9 +243,12 @@ public function getHtml(Timerange $timerange, array $config = null)
             $slaClass = 'ok';
         }
 
+        $reportType = $this instanceof HostSlaReport ? 'Hosts' : 'Services';
```
nilmerg, line 246: Please construct the label in a single step.

e.g. `t('Total (%d Hosts)')`

nilmerg (changes requested):

nilmerg:
I just noticed, the idoreports module uses `Host SLA` (without the `Report` in addition), as it's clear anyway that it's a report, so it is enough. This should also ditch the `Report` extension.

nilmerg (approved):

PR #529, opened 2022-04-29
Controller: Don't set store key because it is no longer required

sukhwinder33445:
ref https://github.com/Icinga/icingaweb2/pull/4765

### DIFF
--- a/library/Icingadb/Web/Controller.php
+++ b/library/Icingadb/Web/Controller.php
@@ -188,6 +188,9 @@ class Controller extends CompatController
 
                 try {
                     $preferencesStore = PreferencesStore::create(new ConfigObject([
+                        //TODO: Don't set store key as it will no longer be needed once we drop support for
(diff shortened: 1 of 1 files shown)


nilmerg:
We cannot drop support with Icinga Web 2.10. Add a todo, leave it as is and make sure it isn't incompatible with v2.11 if it stays.

library/Icingadb/Web/Controller.php
```diff
@@ -188,6 +188,8 @@ function (ViewModeSwitcher $viewModeSwitcher) use (
 
                 try {
                     $preferencesStore = PreferencesStore::create(new ConfigObject([
+                        //TODO: Don't set store key as it will no longer be needed once Web v2.11 is released.
```
nilmerg, line 191: It's not about when v2.11 is out, but when we don't need to support earlier versions.

nilmerg (approved):

PR #535, opened 2022-05-18
services/grid: Use separate queries for axes and content

nilmerg:
Ensures compatibility with `<incoming-ipl-orm-pr>`. Backwards compatibility is still the case.


nilmerg:
Obsolete, for now.

Issue #513, opened 2022-03-28
PHP 8.1 bugs


nilmerg:
Collection of PHP 8.1 bugs. One issue per comment.

nilmerg:
Having no redis data for an object and visiting the source tab of it shows this:
![Screenshot from 2022-03-28 10-38-06](https://user-images.githubusercontent.com/16668527/160359939-f534d9d2-9bda-4017-acb7-2340115fdbf3.png)

yhabteab:
( ! ) Deprecated: Return type of Predis\Client::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /usr/share/icinga-php/vendor/vendor/predis/predis/src/Client.php on line 532
--

yhabteab:
<img width="1268" alt="Bildschirmfoto 2022-04-05 um 18 36 04" src="https://user-images.githubusercontent.com/57616252/161803098-87e0dc5d-09da-47e9-b7b4-4ff5df37ac15.png">

nilmerg:
Not sure how to trigger it though...

![Screenshot from 2022-04-07 16-02-18](https://user-images.githubusercontent.com/16668527/162217471-f1d6bbdc-f871-45a2-aa14-673273237945.png)

yhabteab:
<img width="1216" alt="Bildschirmfoto 2022-05-24 um 13 19 55" src="https://user-images.githubusercontent.com/57616252/170023058-7d882446-463c-4029-8b7c-c61455f43eb5.png">

yhabteab:
> ```
> ( ! ) Deprecated: Return type of Predis\Client::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /usr/share/icinga-php/vendor/vendor/predis/predis/src/Client.php on line 532
> --
> ```

This is also already fixed by the maintainers.

nilmerg:
Which version?

nilmerg:
All mentioned problems are fixed by now.

PR #543, opened 2022-05-24
Fix PHP 8.1 related bugs

yhabteab:
refs #513


library/Icingadb/Widget/ItemList/BaseDowntimeListItem.php
```diff
@@ -66,7 +66,7 @@ protected function init()
         $this->isActive = $this->item->is_in_effect
             || $this->item->is_flexible && $this->item->scheduled_start_time <= $this->currentTime;
 
-        $until = ($this->isActive ? $this->endTime : $this->startTime) - $this->currentTime;
+        $until = (int) ($this->isActive ? $this->endTime : $this->startTime) - $this->currentTime;
```
nilmerg, line 69: The error occurred on line no. 71. I suspect it's been the modulo operation that caused it. Can't you just cast `$until` to int there? Then the date formatting would still get the float, might be useful :thinking:

nilmerg (approved):

nilmerg (approved):

PR #549, opened 2022-06-07
Fix user header baseline

### LINKED ISSUE #531: User/-group header baseline incorrect
## Describe the bug
Screenshot from 2022-05-06 11-16-21

## Expected behavior
The A is on the same baseline as the title.

## Additional context
This used to look fine. Without looking the cause up, this can't be worked on.

sukhwinder33445:
fix #531


nilmerg (changes requested):
This increases the height of the header, which is not acceptable.

nilmerg (changes requested):
Doesn't work for me ;)

Screenshot from 2022-06-13 13-12-34

nilmerg (approved):

PR #558, opened 2022-06-22
Auth: Apply objects filter with `servicegroup.name` correctly

### LINKED ISSUE #545: Mixed host/service listings don't apply relation filters correctly
## Describe the bug

Consider a host and a service downtime, which are shown on `icingadb/downtimes`. Now add a service group filter `servicegroup.name=backend-services-dev-docker`. The service downtime should be visible because it's service is part of this group. The host should also be visible, because it has services that are in this group. (Though, not necessarily the same service for which the service downtime is)

But, the host is not visible.

The same can be observed when doing the same on the comment list. History is another candidate.

## Expected behavior
The host downtime should be visible.

## Additional context

This is caused by the `ReRoute` behavior that's registered by the `Downtime` model. `downtime.servicegroup` is resolved by `downtime.service.servicegroup`. Though, a host downtime has no service attached to it, thus it cannot match.

The solution would be to resolve the filter this way instead: `servicegroup.name=backend-services-dev-docker|downtime.object_type=host&host.service.servicegroup.name=backend-services-dev-docker`

yhabteab:
fixes #545

### DIFF
--- a/library/Icingadb/Model/Behavior/ReRoute.php
+++ b/library/Icingadb/Model/Behavior/ReRoute.php
@@ -12,6 +12,13 @@ class ReRoute implements RewriteFilterBehavior, RewritePathBehavior
 {
     protected $routes;
 
+    /**
+     * Tables with mixed object type entries for which servicegroup filters need to be resolved in multiple steps
+     *
+     * @var string[]
+     */
+    const MIXED_TYPE_RELATIONS = ['downtime', 'comment', 'history', 'notification_history'];
+
     public function __construct(array $routes)
     {
         $this->routes = $routes;
@@ -37,6 +44,21 @@ class ReRoute implements RewriteFilterBehavior, RewritePathBehavior
                     'forceOptimization',
                     $condition->metaData()->get('forceOptimization')
                 );
+            }
+
+            if (
+                in_array(substr($relation, 0, -1), self::MIXED_TYPE_RELATIONS)
+                && substr($remainingPath, 0, 13) === 'servicegroup.'
+            ) {
+                $applyAll = Filter::all();
+                $applyAll->add(Filter::equal($relation . 'object_type', 'host'));
+
+                $orgFilter = clone $filter;
+                $orgFilter->setColumn($relation . 'host.' . $path);
+
+                $applyAll->add($orgFilter);
+
+                $filter = Filter::any($filter, $applyAll);
             }
 
             return $filter;
--- a/library/Icingadb/Model/Comment.php
+++ b/library/Icingadb/Model/Comment.php
@@ -88,11 +88,6 @@ class Comment extends Model
             'expire_time'
         ]));
 
-        $behaviors->add(new ReRoute([
-            'hostgroup'     => 'host.hostgroup',
-            'servicegroup'  => 'service.servicegroup'
-        ]));
-
         $behaviors->add(new Binary([
             'id',
             'environment_id',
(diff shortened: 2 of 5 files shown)


nilmerg (changes requested):
This has some big flaws:

* It only supports equal comparisons, a `servicegroup.name!=` isn't
* It merges multiple servicegroup filters on different levels into one, thus changing the filter semantic (e.g. `servicegroup.name=foo|(hostgroup.name=bar&servicegroup.name=blub)` becomes `servicegroup.name=foo|(hostgroup.name=bar&servicegroup.name=blub)|(downtime.object_type=host&(host.service.servicegroup.name=foo|host.service.servicegroup.name=blub))`)

And lastly, the case described in the issue isn't solved at all, because this only applies to restrictions. Filtering using the search bar is still broken.

In the current form, this is completely wrong, I'm afraid.

Before continuing here, we should first talk about what needs to be considered and what a possible solution could be that meets all requirements.

library/Icingadb/Model/Behavior/ReRoute.php
```diff
@@ -39,6 +46,14 @@ public function rewriteCondition(Filter\Condition $condition, $relation = null)
                 );
             }
 
+            if (in_array(substr($relation, 0, -1), self::SPECIAL_RELATIONS) && $remainingPath === 'servicegroup.name') {
+                $applyAll = Filter::all();
+                $applyAll->add(Filter::like($relation . 'object_type', 'host'));
+                $applyAll->add(Filter::like('host.' . $path, $condition->getValue()));
```
nilmerg, line 52: You can't use `Filter::like` here. The condition might be of any type. It must not change! Use the `$class` variable from above here as well.

Though, maybe just clone `$filter` and change the column name. This way the meta data is preserved and you don't have to transfer it again.

nilmerg: Also, prepend `$relation` to `host.`. It must be a fully qualified path that's used by the condition.

library/Icingadb/Model/Behavior/ReRoute.php
```diff
@@ -12,6 +12,13 @@ class ReRoute implements RewriteFilterBehavior, RewritePathBehavior
 {
     protected $routes;
 
+    /**
+     * Special tables which require an additional step to apply filters with service group name
```
nilmerg, line 16: That's not very informative. The name of the constant is neither.

Suggestion:

**Description**: `Tables with mixed object type entries for which servicegroup filters need to be resolved in multiple steps`
**Constant**: `MIXED_TYPE_RELATIONS`

nilmerg (changes requested):

library/Icingadb/Model/Behavior/ReRoute.php
```diff
@@ -39,6 +46,21 @@ public function rewriteCondition(Filter\Condition $condition, $relation = null)
                 );
             }
 
+            if (
+                in_array(substr($relation, 0, -1), self::MIXED_TYPE_RELATIONS)
+                && strpos($remainingPath, 'servicegroup.') !== false
```
nilmerg, line 51: Please use substr here and make a startswith check. I don't like the fact that `servicegroup.` may appear anywhere.

nilmerg (changes requested):

nilmerg (approved):
Thanks!

PR #561, opened 2022-06-23
ApplicationState: Check the availability of `IcingaRedis` first

sukhwinder33445:
When icinga redis is down, icingadb also shut down after a few minutes.
In this case, the real cause of icingadb being down is that redis is down.
So we should check icinga redis status first.


yhabteab (approved):
LFTM!

PR #564, opened 2022-06-27
Don't sort perfData which can't be visualized

### LINKED ISSUE #562: Don't sort performance data that can't represent a pie chart
To list pie charts of performance data by severity, we sort all performance data by state and then by percentage if it provides a maximum value. This does not work properly for performance data that is not a pie chart and results in an unexpected view order.

yhabteab:
This ensures that all perfdata which can't be visualized are rendered exactly in the
same order they are stored in the database after all vizualizable perfData are rendered.

fixes #562

### DIFF
--- a/library/Icingadb/Widget/Detail/PerfDataTable.php
+++ b/library/Icingadb/Widget/Detail/PerfDataTable.php
@@ -52,6 +52,14 @@ class PerfDataTable extends Table
         uasort(
             $pieChartData,
             function ($a, $b) {
+                if ($a->isVisualizable() && ! $b->isVisualizable()) {
+                    return -1;
+                } elseif (! $a->isVisualizable() && $b->isVisualizable()) {
+                    return 1;
+                } elseif (! $a->isVisualizable() && ! $b->isVisualizable()) {
+                    return 0;
+                }
+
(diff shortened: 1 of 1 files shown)


nilmerg:
Have you considered to adjust the sort callback instead?

yhabteab:
> Have you considered to adjust the sort callback instead?

I did, but then the visualizable perfData would be rendered after the non-visualizable when there is mixed perfData.

nilmerg:
Then I guess your algorithm wasn't sophisticated enough :sweat_smile: 

Shouldn't this work?:

1) If $a is visualizable and $b is, standard comparison
2) If $a is visualizable but $b is not, return `-1` (so that pie charts are always above non-pie charts)
3) If $a is not visualizable but $b is, return `1`
4) If neither $a nor $b is visualizable, return `0` (should maintain the order of non-pie charts)

lippserd:
uasort is only stable since PHP 8.

nilmerg:
Okay, then it's only stable upon PHP 8. But it never was stable to begin with and no-one cared, so this isn't important IMHO. I just don't like the three passes here (filter, sort, merge) instead of just one (sort)

nilmerg (approved):

PR #565, opened 2022-06-27
Don't use `columns()` of ipl/sql to specify columns to be selected

yhabteab:
The `Sql::columns()` method doesn't override existing columns when calling this directly, because
at this point the ipl/orm query has already collected all selectable columns, which are not required
for this use case. A later call to the `columns()` method using the select base doesn't prevent all the
columns registered by the query from being selected.


nilmerg:
How did you notice this?

yhabteab:
> How did you notice this?

I noticed this last time when I was trying to check dashboard user using `columns('1')`. Then I remembered that I also used such a query in icingadb-web as well.

nilmerg (approved):

PR #566, opened 2022-06-27
Redis: Apply volatile state results directly to the state model

### LINKED ISSUE #563: Inconsistent volatile state
We only apply certain volatile state values from Redis, which can result in inconsistent view state. All volatile state values from Redis should be applied.

yhabteab:
Since all the inconsistent column/redis key names have been renamed, we don't have to cache redis keys
separately anymore. This PR applies all the `Service/Host` volatile state results directly to the
respective state model.

fixes #563


nilmerg (approved):

lippserd (changes requested):
Icinga 2 may not write keys to redis if there is no value. Something like the following is required in DB Web:

```
$data = array_merge(array_fill_keys($columns, null), $data);
```

nilmerg (changes requested):
Please remove class `VolatileState` entirely, it's unused (and obsolete) now.

library/Icingadb/Redis/VolatileStateResults.php
```diff
@@ -114,13 +124,14 @@ protected function applyRedisUpdates()
...
 
-    protected function fetchStates(string $key, array $ids): Generator
+    protected function fetchStates(string $key, array $ids, array $keys): Generator
     {
         $results = IcingaRedis::instance()->getConnection()->hmget($key, $ids);
         foreach ($results as $i => $json) {
             if ($json !== null) {
                 $data = json_decode($json, true);
-                $data = array_intersect_key($data, array_flip(VolatileState::$keys));
+                $data = array_merge(array_fill_keys($keys, null), $data);
```
nilmerg, line 133: You're overriding the decoded json data here, leading to a fully unset state.

nilmerg (changes requested):

library/Icingadb/Redis/VolatileStateResults.php
```diff
@@ -114,13 +124,14 @@ protected function applyRedisUpdates()
...
+                $data = array_intersect_key($data, $keys);
```
nilmerg, line 134: Where did the array_flip went? Now $data is empty and the redis result is not used??

nilmerg (changes requested):

nilmerg (approved):

PR #575, opened 2022-07-01
Sort also pie charts in detailed view mode

yhabteab:
At the moment we only render the sorted perfData in the object details, but when switching to
the detailed view mode the same perfData are rendered unsorted. This PR ensures that the perfData
are also sorted once after they have been successfully parsed.


nilmerg (commented):
There are only two ways to retrieve perf data objects from a set:

1) Iterating over it (calls `asArray()`)
2) Calling `asArray()`

This means, there is no way to retrieve the original data anyway once `asArray()` orders.

So I opt to let `parse()` perform the ordering. This ensures it runs only once.

nilmerg (approved):

PR #593, opened 2022-07-12
README.md: Fix wrong latest doc link

yhabteab:


nilmerg:
@bobapple The package isn't called `icinga-db-web` either, but `icingadb-web`, the module name here is also `icingadb`. Why are we now using `icinga-db` in the website urls? I mean, I'd accept this as well if you insist on this :man_shrugging:

bobapple:
The reason is the "Product Name":

- Icinga Web is `/icinga-web`
- Icinga 2 is `/icinga-2`
- Icinga Director is `/icinga-director`
- Icinga for Windows is `/icinga-for-windows`
- Icinga DB is `/icinga-db`
- Icinga DB Web is `/icinga-db-web`
- and so on...

nilmerg (approved):

PR #594, opened 2022-07-12
Add some extra columns to the union models to be used by the nagvis m…

yhabteab:
…odule

### DIFF
--- a/library/Icingadb/Model/Hostgroupsummary.php
+++ b/library/Icingadb/Model/Hostgroupsummary.php
@@ -143,9 +143,13 @@ class Hostgroupsummary extends UnionModel
                     'host_state'             => 'state.soft_state',
                     'host_handled'           => 'state.is_handled',
                     'host_severity'          => 'state.severity',
+                    'host_is_in_downtime'    => 'state.in_downtime',
+                    'host_is_acknowledged'   => 'state.is_acknowledged',
                     'service_id'             => new Expression('NULL'),
                     'service_state'          => new Expression('NULL'),
-                    'service_handled'        => new Expression('NULL')
+                    'service_handled'        => new Expression('NULL'),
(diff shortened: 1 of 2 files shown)


nilmerg:
I'm not convinced that the nagvis module requires a change here.

* The new columns aren't even part of the primary column set and are not aggregated, i.e. they won't and can't be queried
* It breaks with Postgres

yhabteab:
> * The new columns aren't even part of the primary column set and are not aggregated, i.e. they won't and can't be queried

Since these are only being used by the nagvis module, I thought of including these columns only in the union queries. The nagvis module can then query them with such an expression.

```php
'hosts_up_downtime' => new Expression('SUM(CASE WHEN host_state = 0 AND host_is_in_downtime = \'y\' THEN 1 ELSE 0 END)')
```

Though, if you know how I can create such expression without changing anything in DB Web, I would be eager to hear it.

nilmerg:
> Since these are only being used by the nagvis module, I thought of including these columns only in the union queries.

Nah, starting with ` only being used by ` should have alerted you that this assumption is incorrect.

What makes this even worse is this statement in the related PR:

```
$query = Hostgroupsummary::on($this->getDb());
$query->getSelectBase()
    ->columns($this->hostStateSummary)
    ->columns($this->serviceStateSummary);
```

You are bypassing the orm. Why? Because the orm doesn't know anything about these new columns, because they're not part of the model's primary columns! Another, very obvious, hint that something is wrong here.

> if you know how I can create such expression without changing anything in DB Web, I would be eager to hear it.

The first thing that comes to my mind, is to inherit from a model that provides what you want and to extend it.

But we're talking about a model that implements the `UnionModel` here. This is subject to change and I don't like to see that the nagvis module is dependent on what our `HostgroupSummary` model does.

What you should do, is copy the models over from here and extend them directly.

PR #601, opened 2022-07-19
Preserve filters after clicking `Load More`

### LINKED ISSUE #590: History load more doesn't preserve filters
## Describe the bug
Filtering the history and clicking load more, shows unfiltered results on page no. 2.

## Expected behavior
Page no. 2 and further are filtered.

## Your Environment

* Icinga DB Web version (System - About): master

raviks789:
Currently, the load more urls are fixed and hence the filters are not preserved after clicking on load more.
In this fix load more url has been changed dynamically upon applying filters for history or notifications and
hence preserving the filters.

fixes #590

### DIFF
--- a/application/controllers/HistoryController.php
+++ b/application/controllers/HistoryController.php
@@ -13,6 +13,7 @@ use Icinga\Module\Icingadb\Web\Control\ViewModeSwitcher;
 use ipl\Stdlib\Filter;
 use ipl\Web\Control\LimitControl;
 use ipl\Web\Control\SortControl;
+use ipl\Web\Filter\QueryString;
 use ipl\Web\Url;
 
 class HistoryController extends Controller
@@ -21,6 +22,12 @@ class HistoryController extends Controller
     {
         $this->addTitleTab(t('History'));
         $compact = $this->view->compact; // TODO: Find a less-legacy way..
+
+        $preserveParams = [
+            LimitControl::DEFAULT_LIMIT_PARAM,
+            SortControl::DEFAULT_SORT_PARAM,
+            ViewModeSwitcher::DEFAULT_VIEW_MODE_PARAM
+        ];
 
         $db = $this->getDb();
 
@@ -38,7 +45,6 @@ class HistoryController extends Controller
         ]);
 
         $before = $this->params->shift('before', time());
-        $url = Url::fromPath('icingadb/history')->setParams(clone $this->params);
 
         $limitControl = $this->createLimitControl();
         $paginationControl = $this->createPaginationControl($history);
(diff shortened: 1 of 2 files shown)


nilmerg (changes requested):
* Define a filter and reload the page or column. The load more url is now missing the filter. (Worked previously)
* Define a filter and click on load more once. A second click on load more returns unfiltered results again.

application/controllers/HistoryController.php
```diff
@@ -65,6 +66,7 @@ public function indexAction()
             }
         } else {
             $filter = $searchBar->getFilter();
+            $url->setQueryString(QueryString::render($filter));
```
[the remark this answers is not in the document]
nilmerg: Also, `$url` gets some params applied when it's initialized. They're overridden here. These are params like `limit` for example, which are now lost and not transferred to the load more url anymore.

I reckon that has been intentional (:wink:) and only one fact has been forgotten: When submitting the search bar, filters are not part of `$this->params`. If the search bar is not sent, they are though. So, to fix this, I guess you need one of your conditions back (if hasBeenSent) and then apply `$filter` in a way that doesn't override other params.

nilmerg (changes requested):

application/controllers/HistoryController.php
```diff
@@ -49,11 +50,15 @@ public function indexAction()
             ]
         );
         $viewModeSwitcher = $this->createViewModeSwitcher($paginationControl, $limitControl, true);
-        $searchBar = $this->createSearchBar($history, [
+        $preserveParams = [
```
nilmerg, line 53: The actions have all their specific structure:

* Init (params, auth, tabs)
* get db
* setup query
* setup controls
* handle controls
* apply filters
* export
* register controls
* register content

So initializing `$preserveParams` should be at the top and `createSearchBar` where it has been.

application/controllers/HistoryController.php
```diff
@@ -49,11 +50,15 @@ public function indexAction()
...
             $limitControl->getLimitParam(),
             $sortControl->getSortParam(),
             $viewModeSwitcher->getViewModeParam()
-        ]);
+        ];
+
+
+        $searchBar = $this->createSearchBar($history, $preserveParams);
+        $url = $url->onlyWith($preserveParams);
```
nilmerg, line 61: You should put that to the same place where the query string is handled. Well, why not even put it right above where it's required. (Around line no. 103) You might even move the initialization of `$url` there as well. To shorten everything a bit:

```
$url = Url::fromRequest()->onlyWith...
$url->setQueryString(QueryString::render(...
```

nilmerg (changes requested):

nilmerg (approved):

PR #605, opened 2022-07-26
Fix pending state balls

raviks789:
`pending` state balls will be represented as hollowed state balls instead of solid state balls (ipl-web@3bccdd0da71a178fc0b10422275bfee8c4d3f4a4). Hence, `pending` state balls must behave similarly as `ok` or `up` upon change in state of the object.

Also, in case a downtime is scheduled for `pending` objects, their state balls must contain `plug` icon.

depends on icinga/ipl-web#90

### DIFF
--- a/library/Icingadb/Widget/ItemList/StateListItem.php
+++ b/library/Icingadb/Widget/ItemList/StateListItem.php
@@ -82,7 +82,6 @@ abstract class StateListItem extends BaseListItem
             $stateBall->getAttributes()->add('class', 'handled');
         } elseif ($this->state->getStateText() === 'pending' && $this->state->in_downtime) {
             $stateBall->addHtml(new Icon($this->getHandledIcon()));
-            $stateBall->getAttributes()->add('class', 'handled');
         }
 
         $visual->addHtml($stateBall);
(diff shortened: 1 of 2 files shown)


library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -80,6 +80,9 @@ protected function assembleVisual(BaseHtmlElement $visual)
         if ($this->state->is_handled) {
             $stateBall->addHtml(new Icon($this->getHandledIcon()));
             $stateBall->getAttributes()->add('class', 'handled');
+        } elseif ($this->state->getStateText() === 'pending' && $this->state->in_downtime) {
+            $stateBall->addHtml(new Icon($this->getHandledIcon()));
+            $stateBall->getAttributes()->add('class', 'handled');
```
nilmerg, line 85: I'm not sure this class suits hollow state balls, the contrast of the light outline with the background is very low. In light mode of course.

@flourish86 What do you think?

flourish86: > @flourish86 What do you think?

I see your point. After looking at the following test view (screenshot), I find that it works better to not apply the transparency.

We don't have any other outlined state balls that could be handled, do we? If so, I think we can establish the rule of not reducing transparency on handled item state balls. The icon itself is still enough to convey the handled state.

Screen Shot 2022-08-17 at 16 43 44

nilmerg: Aye, thanks :heart:. @raviks789 Please don't add the handled class here.

PR #618, opened 2022-08-15
Introduce Total Host and Total Service SLA Report

jhoxhaa:
resolves: Icinga/icingadb-web#771

### DIFF
--- a/library/Icingadb/Hook/Common/TotalSlaReportUtils.php
+++ b/library/Icingadb/Hook/Common/TotalSlaReportUtils.php
@@ -1,0 +1,56 @@
+<?php
+
+/* Icinga DB Web | (c) 2023 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Icingadb\Hook\Common;
+
+use Icinga\Module\Icingadb\ProvidedHook\Reporting\HostSlaReport;
+use Icinga\Module\Icingadb\Widget\EmptyState;
+use Icinga\Module\Reporting\Timerange;
+use ipl\Html\Html;
+
+use function ipl\I18n\t;
+
+trait TotalSlaReportUtils
+{
+    public function getHtml(Timerange $timerange, array $config = null)
+    {
+        $data = $this->getData($timerange, $config);
+        $count = $data->count();
+
+        if (! $count) {
+            return new EmptyState(t('No data found.'));
+        }
+
+        $threshold =  (float) ($config['threshold'] ?? static::DEFAULT_THRESHOLD);
+
+        $tableRows = [];
+        $precision = $config['sla_precision'] ?? static::DEFAULT_REPORT_PRECISION;
+
+        // We only have one average
+        $average = $data->getAverages()[0];
+
+        if ($average < $threshold) {
+            $slaClass = 'nok';
+        } else {
+            $slaClass = 'ok';
+        }
+
+        $total = $this instanceof HostSlaReport
+            ? sprintf(t('Total (%d Hosts)'), $count)
+            : sprintf(t('Total (%d Services)'), $count);
+
+        $tableRows[] = Html::tag('tr', null, [
+            Html::tag('td', ['colspan' => count($data->getDimensions())], $total),
+            Html::tag('td', ['class' => "sla-column $slaClass"], round($average, $precision))
+        ]);
+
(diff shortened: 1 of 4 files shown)


yhabteab (changes requested):
There seems to be quite a lot of duplicated code, even though the two new classes are doing exactly the same thing.

**Tip:** ~~Create a new base class called `TotalSlaReport` that extends the `SlaReport` class and the other two classes can inherit from this new class and only have to implement the `getName()` method~~.

Have double checked the code and the duplicated code is only in the `getHtml()` method.  I think it is fine to implement this method in a trait instead.

```php
trait TotalSlaReportUtils
{
     public function getHtml(Timerange $timerange, array $config = null)
     {
           // Implement this here
      }
}
```

And you can just use this method in the new classes.

```php
class TotalServiceSlaReport extends ServiceSlaReport
{
    use TotalSlaReportUtils;

     public function getName()
     {
           // Implement me
      }
}

class TotalHostSlaReport extends HostSlayReport
{
     use TotalSlaReportUtils;

     public function getName()
     {
           // Implement me
      }
}
```

yhabteab:
Ah, and you don't need an `(Icinga DB)` suffix for the names. These hooks are only provided by Icinga DB, so there can't be any naming conflict.

library/Icingadb/ProvidedHook/Reporting/TotalHostSlaReport.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Icinga\Module\Icingadb\ProvidedHook\Reporting;
+
+use Icinga\Module\Icingadb\Hook\Common\TotalSlaReportUtils;
+
+use function ipl\I18n\t;
+
+class TotalHostSlaReport extends HostSlaReport
+{
+    use TotalSlaReportUtils;
```
yhabteab, line 11: Missing empty line!

library/Icingadb/ProvidedHook/Reporting/TotalServiceSlaReport.php
```diff
@@ -0,0 +1,16 @@
...
+
+namespace Icinga\Module\Icingadb\ProvidedHook\Reporting;
+
+use Icinga\Module\Icingadb\Hook\Common\TotalSlaReportUtils;
+
+use function ipl\I18n\t;
+
+class TotalServiceSlaReport extends ServiceSlaReport
+{
+    use TotalSlaReportUtils;
```
yhabteab, line 11: same here!

yhabteab (changes requested):

yhabteab (changes requested):
Missing license headers!

yhabteab (dismissed):

library/Icingadb/Hook/Common/TotalSlaReportUtils.php
```diff
@@ -0,0 +1,55 @@
...
+{
+    public function getHtml(Timerange $timerange, array $config = null)
+    {
+        $data = $this->getData($timerange, $config);
+
+        if (! count($data)) {
+            return new EmptyState(t('No data found.'));
+        }
+
+        $threshold = isset($config['threshold']) ? (float) $config['threshold'] : static::DEFAULT_THRESHOLD;
```
nilmerg, line 24: You could also use this:

```suggestion
        $threshold = (float) ($config['threshold'] ?? static::DEFAULT_THRESHOLD);
```

The parentheses around the null coalescing operation are required as otherwise the float cast is applied only to `$config['threshold']` resulting in the cast from `NULL` to `0.0`.

library/Icingadb/Hook/Common/TotalSlaReportUtils.php
```diff
@@ -0,0 +1,55 @@
...
+
+use function ipl\I18n\t;
+
+trait TotalSlaReportUtils
+{
+    public function getHtml(Timerange $timerange, array $config = null)
+    {
+        $data = $this->getData($timerange, $config);
+
+        if (! count($data)) {
```
nilmerg, line 20: You throw away the count of results here, although you need it again below anyway? Please only count once :wink:

nilmerg (changes requested):

nilmerg (approved):

Issue #629, opened 2022-07-24
Provide generic tts integration


aledr:
Hi.

Do you have plans to support [icingadb-web](/Icinga/icingadb-web)?

Thanks for your work on this plugin.

PR #630, opened 2022-08-29
Provide generic tts integration

raviks789:
ref #629

### DIFF
--- a/library/Icingadb/Common/TicketLinks.php
+++ b/library/Icingadb/Common/TicketLinks.php
@@ -1,0 +1,56 @@
+<?php
+
+/* Icinga DB Web | (c) 2022 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Icingadb\Common;
+
+use Icinga\Application\Hook;
+
+trait TicketLinks
+{
+    /** @var bool */
+    protected $ticketLinkEnabled = false;
+
+    /**
+     * Set whether list items should render host and service links
+     *
+     * @param bool $state
+     *
+     * @return $this
+     */
+    public function setTicketLinkEnabled(bool $state = true): self
+    {
+        $this->ticketLinkEnabled = $state;
+
+        return $this;
+    }
+
+    /**
+     * Get whether list items should render host and service links
+     *
+     * @return bool
+     */
+    public function getTicketLinkEnabled(): bool
+    {
+        return $this->ticketLinkEnabled;
+    }
+
+    /**
+     * Get whether list items should render host and service links
+     *
+     * @return string
+     */
+    public function createTicketLinks($text): string
+    {
+        if (Hook::has('ticket')) {
+            $tickets = Hook::first('ticket');
+        }
+
+        if ($this->getTicketLinkEnabled() && isset($tickets)) {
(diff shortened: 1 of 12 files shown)


nilmerg:
Planned for v1.1. Once working on this version, I'd like to discuss whether it makes sense to re-use some existing hooks or whether to introduce entirely new hooks. (e.g. `CommentDetailExtensionHook`) Why? Because I'd like to get rid of `TicketHook` in Icinga Web 2 and have its implementation provided by generictts. (What the missing support for it in Icinga DB Web showed, is that it's the only module making use of the hook, so why the hell not..)

lippserd:
A new hook like `CommentDetailExtensionHook` sounds good to me as well!

library/Icingadb/Common/TicketLinks.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+/* Icinga DB Web | (c) 2021 Icinga GmbH | GPLv2 */
```
nilmerg, line 3: It's 2022 ;)

library/Icingadb/Common/TicketLinks.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    /**
+     * Get whether list items should render host and service links
+     *
+     * @return string
+     */
+    public function createTicketLinks($text): string
+    {
+        $tickets = Icinga::app()->getViewRenderer()->view->tickets;
```
nilmerg, line 45: I don't quite see an advantage in this. Just retrieve the hook here directly by calling `Hook::first()`.

library/Icingadb/Common/TicketLinks.php
```diff
@@ -0,0 +1,54 @@
...
+
+        if ($this->getTicketLinkEnabled()) {
+            /** @var \Icinga\Application\Hook\TicketHook $tickets */
+            return isset($tickets) ? $tickets->createLinks($text) : $text;
```
nilmerg, line 49: Why not move the `isset()` into the `if`?

library/Icingadb/Web/Controller.php
```diff
@@ -562,5 +563,9 @@ protected function moduleInit()
         Icinga::app()->getFrontController()
             ->getPlugin('Zend_Controller_Plugin_ErrorHandler')
             ->setErrorHandlerModule('icingadb');
+
+        if (Hook::has('ticket')) {
+            $this->view->tickets = Hook::first('ticket');
+        }
```
nilmerg, line 569: Please remove this.

nilmerg (changes requested):

library/Icingadb/Web/Controller.php
```diff
@@ -8,6 +8,7 @@
 use Generator;
 use GuzzleHttp\Psr7\ServerRequest;
 use Icinga\Application\Config;
+use Icinga\Application\Hook;
```
nilmerg, line 11: There's still this now unused use statement

nilmerg (changes requested):

nilmerg (approved):
Thanks!

PR #649, opened 2022-09-19
Fix/SetApiTimeout

pdolinic:
Trying to connect to the API, there is no default timeout specified. This one sets a debatable one with 15 seconds, whereas monitoring has 30s.

### DIFF
--- a/library/Icingadb/Command/Transport/ApiCommandTransport.php
+++ b/library/Icingadb/Command/Transport/ApiCommandTransport.php
@@ -311,7 +311,7 @@ class ApiCommandTransport implements CommandTransportInterface
     public function probe()
     {
         try {
-            $response = (new Client(['timeout' => 15 ]))
+            $response = (new Client(['timeout' =>15]))
                 ->get($this->getUriFor(''), [
                     'auth'          => [$this->getUsername(), $this->getPassword()],
                     'headers'       => ['Accept' => 'application/json'],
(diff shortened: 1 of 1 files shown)


library/Icingadb/Command/Transport/ApiCommandTransport.php
```diff
@@ -311,7 +311,7 @@ public function send(IcingaCommand $command, int $now = null)
     public function probe()
     {
         try {
-            $response = (new Client())
+            $response = (new Client(['timeout' => '15']))
```
nilmerg, line 314: Please pass the seconds as integer, not as string.

nilmerg (changes requested):

nilmerg:
Please add the missing space as mentioned in the checks and take a look at your emails associated with the commits. I'd like to include this in the next minor update.

library/Icingadb/Command/Transport/ApiCommandTransport.php
```diff
@@ -311,7 +311,7 @@ public function send(IcingaCommand $command, int $now = null)
     public function probe()
     {
         try {
-            $response = (new Client())
+            $response = (new Client(['timeout' => 15 ]))
```
lippserd, line 314: Remove the space after the `15` please. Only the one before is required.

lippserd (dismissed):

nilmerg (approved):

PR #651, opened 2022-09-30
SlaReport: Breakdown into an hour view

jhoxhaa:
resolves: Icinga/icingadb-web#770

### DIFF
--- a/library/Icingadb/ProvidedHook/Reporting/SlaReport.php
+++ b/library/Icingadb/ProvidedHook/Reporting/SlaReport.php
@@ -71,6 +71,12 @@ abstract class SlaReport extends ReportHook
 
         if (isset($config['breakdown']) && $config['breakdown'] !== 'none') {
             switch ($config['breakdown']) {
+                case 'hour':
+                    $interval = new DateInterval('PT1H');
+                    $format = 'H:i:s';
+                    $boundary = '+1 hour';
+
+                    break;
                 case 'day':
                     $interval = new DateInterval('P1D');
                     $format = 'Y-m-d';
@@ -168,6 +174,7 @@ abstract class SlaReport extends ReportHook
             'label'   => t('Breakdown'),
             'options' => [
                 'none'  => t('None', 'SLA Report Breakdown'),
+                'hour'  => t('Hour'),
                 'day'   => t('Day'),
                 'week'  => t('Week'),
                 'month' => t('Month')


library/Icingadb/ProvidedHook/Reporting/SlaReport.php
```diff
@@ -234,7 +237,21 @@ public function getHtml(Timerange $timerange, array $config = null)
                 $slaClass = 'ok';
             }
 
-            $cells[] = Html::tag('td', ['class' => "sla-column $slaClass"], round($sla, $precision));
+            if ($config['breakdown'] === 'hour' && $slaClass === 'nok') {
+                foreach ($row->getDimensions() as $dimension) {
+                    $cells[] = Html::tag('td', null, $dimension);
+                }
+            } elseif ($config['breakdown'] != 'hour') {
```
yhabteab, line 244: Think about how you can handle these `if ... elseif` branches in just a single `if` block.

yhabteab (changes requested):

library/Icingadb/ProvidedHook/Reporting/SlaReport.php
```diff
@@ -71,6 +71,12 @@ protected function fetchReportData(Timerange $timerange, array $config = null)
 
         if (isset($config['breakdown']) && $config['breakdown'] !== 'none') {
             switch ($config['breakdown']) {
+                case 'hour':
+                    $interval = new DateInterval('PT1H');
+                    $format = 'h:i:s A';
```
yhabteab, line 76: Why not just use `H:i:s` as a format or is there some technical context to this that I don't get?

yhabteab (approved):

lippserd (commented):
What does "showing only downtimes" mean here?

davixd:
@lippserd 
I'm the requester of https://github.com/Icinga/icingaweb2-module-reporting/issues/146. As far as I could follow: jhoxhaa implemented my feature request on point. "showing only downtimes" should mean the following:  as the breakdown range now gets more detailed (in this case a hour view is choosable for each day), in this case its not necessary to create a report for each day with a breakdown of a hour which shows only the availability of 100% in a detailed view of a hour range. Such report would have several unnecessary pages.  My idea was to create a breakdown in a hour view, only when a downtime occurred.

Example Report for January 2023:
...
29. January - 92%
14:00 - 16:00 8% (Downtime occurrence)
30. January - 100%
...

instead of:
...
29. January - 100%
00:00 0%
01:00 0%
02:00 0%
03:00 0%
04:00 0%
and so on (24 hours)
30. January - 100%
...

As the day summary shows 100% of availability, I do not need to have a breakdown into hours. As the day shows 92% availability, our management would like to know in which hours of the day the downtime did occurred, so there is a breakdown of hours necessary (but only the affected hours). I hope it makes sense. For us this feature is really necessary to be able to use this kind of Reports for our management.

PR #655, opened 2022-10-12
Add reporting hooks to DetailsAction

### LINKED ISSUE #704: Action to create/modify report for displayed host/service
I would suggest an action (in the "Problem handling" part of the host/service detail), which will create a report  with the 'Report', 'Filter' and title fields filled in already (or perhaps even modify it, if it exists).

jhoxhaa:
fixes #704

### DIFF
--- a/library/Icingadb/ProvidedHook/CreateHostSlaReport.php
+++ b/library/Icingadb/ProvidedHook/CreateHostSlaReport.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Icinga\Module\Icingadb\ProvidedHook;
+
+use Icinga\Module\Icingadb\Hook\HostsDetailExtensionHook;
+use ipl\Html\Html;
+use ipl\Html\ValidHtml;
+use ipl\Orm\Query;
+use ipl\Web\Filter\QueryString;
+use ipl\Web\Url;
+
+class CreateHostSlaReport extends HostsDetailExtensionHook
+{
+    public function getHtmlForObjects(Query $hosts): ValidHtml
+    {
+        $url = Url::fromPath('reporting/reports/new');
+        $filter = QueryString::render($this->getBaseFilter());
+        $url->addParams(['filter' => $filter, 'report' => 'host']);
+        $wrapper = Html::tag('div');
+        $wrapper
+            ->add(Html::tag('h2', 'Reporting'))
+            ->add(Html::tag('a', ['href' => $url], 'Create Host SLA Report'));
+
+        return $wrapper;
+    }
+}
--- a/library/Icingadb/ProvidedHook/CreateHostSlaReport.php
+++ b/library/Icingadb/ProvidedHook/CreateHostSlaReport.php
@@ -2,25 +2,34 @@
 
 namespace Icinga\Module\Icingadb\ProvidedHook;
 
+use Icinga\Authentication\Auth;
 use Icinga\Module\Icingadb\Hook\HostsDetailExtensionHook;
 use ipl\Html\Html;
+use ipl\Html\HtmlDocument;
 use ipl\Html\ValidHtml;
+use ipl\I18n\Translation;
 use ipl\Orm\Query;
 use ipl\Web\Filter\QueryString;
 use ipl\Web\Url;
+use ipl\Web\Widget\Link;
 
 class CreateHostSlaReport extends HostsDetailExtensionHook
 {
+    use Translation;
+
     public function getHtmlForObjects(Query $hosts): ValidHtml
     {
-        $url = Url::fromPath('reporting/reports/new');
-        $filter = QueryString::render($this->getBaseFilter());
-        $url->addParams(['filter' => $filter, 'report' => 'host']);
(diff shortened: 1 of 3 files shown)


yhabteab (commented):
I think here is not the appropriate repository for these hooks. Also, those hooks shouldn't be object extensions, but rather `HostActions` and `ServiceActions` and have to be provided by the reporting module.

cc @nilmerg

nilmerg:
No, that's the correct location. Icinga DB Web provides the sla report functionality for reporting. Reporting has no idea what SLA means and can't provide actions for this kind of report.

But you're correct with the assumption that these should be actions. Though, I'd even question whether this should be a hook at all. `ObjectDetail::createActions` could do this instead.

lippserd:
> Though, I'd even question whether this should be a hook at all.

The reports are also hooks. I think it makes sense to keep the non-core module functionality separated.

nilmerg (changes requested):
Please let the hooks extend `Icinga\Module\Icingadb\Hook\HostActionsHook` and `Icinga\Module\Icingadb\Hook\ServiceActionsHook`.

library/Icingadb/ProvidedHook/CreateServiceSlaReport.php
```diff
@@ -0,0 +1,26 @@
...
+use ipl\Web\Url;
+
+class CreateServiceSlaReport extends ServicesDetailExtensionHook
+{
+    public function getHtmlForObjects(Query $services): ValidHtml
+    {
+        $url = Url::fromPath('reporting/reports/new');
+        $filter = QueryString::render($this->getBaseFilter());
+        $url->addParams(['filter' => $filter, 'report' => 'service']);
+        $wrapper = Html::tag('div');
```
nilmerg, line 19: Please don't use a wrapper element. The other content also doesn't use one. Use a `HtmlDocument` as wrapper instead.

library/Icingadb/ProvidedHook/CreateHostSlaReport.php
```diff
@@ -0,0 +1,26 @@
...
+use ipl\Web\Url;
+
+class CreateHostSlaReport extends HostsDetailExtensionHook
+{
+    public function getHtmlForObjects(Query $hosts): ValidHtml
+    {
+        $url = Url::fromPath('reporting/reports/new');
+        $filter = QueryString::render($this->getBaseFilter());
+        $url->addParams(['filter' => $filter, 'report' => 'host']);
+        $wrapper = Html::tag('div');
```
nilmerg, line 19: Please don't use a wrapper element. The other content also doesn't use one. Use a `HtmlDocument` as wrapper instead.

library/Icingadb/ProvidedHook/CreateServiceSlaReport.php
```diff
@@ -0,0 +1,26 @@
...
+        $wrapper
+            ->add(Html::tag('h2', 'Reporting'))
+            ->add(Html::tag('a', ['href' => $url], 'Create Service SLA Report'));
```
nilmerg, line 22: Please use the `ipl\Web\Widget\Link` class and translate both labels with `ipl\I18n\Translation::translate()`

library/Icingadb/ProvidedHook/CreateHostSlaReport.php
```diff
@@ -0,0 +1,26 @@
...
+        $wrapper
+            ->add(Html::tag('h2', 'Reporting'))
+            ->add(Html::tag('a', ['href' => $url], 'Create Host SLA Report'));
```
nilmerg, line 22: Please use the `ipl\Web\Widget\Link` class and translate both labels with `ipl\I18n\Translation::translate()`

run.php
```diff
@@ -4,13 +4,24 @@
...
+
 $this->provideHook('ApplicationState');
 $this->provideHook('X509/Sni');
 $this->provideHook('health', 'IcingaHealth');
 $this->provideHook('health', 'RedisHealth');
 $this->provideHook('Reporting/Report', 'Reporting/HostSlaReport');
 $this->provideHook('Reporting/Report', 'Reporting/ServiceSlaReport');
 
+if (
+    $this::exists('reporting') && Icinga::app()->getModuleManager()->hasEnabled('reporting')
```
nilmerg, line 18: `exists` already checks the enabled state if I'm not mistaken :thinking:

run.php
```diff
@@ -4,13 +4,24 @@
...
+    && Auth::getInstance()->hasPermission('reporting/reports')
```
nilmerg, line 19: Accessing the authentication at this stage (run.php runs very early) may cause issues, especially on routes that normally don't need authentication. Placing this here causes everything to require authentication as run.php runs always. Please check the permission in the hooks itself.

nilmerg (changes requested):

nilmerg (changes requested):
Please put always an empty line after closing a control structure.

e.g.:
```
    }

    return $result;
```

Otherwise it looks fine now.

nilmerg (approved):

PR #659, opened 2022-10-19
Avoid passing `null` to `MarkdownLine::__construct`

nilmerg:


lippserd:
@nilmerg What about something like this?

```
class MarkdownLine

...

    public static function create(string $line = null): ?self
    {
        if (empty($line)) {
            return null;
        }

        return new static($line);
    }
```

PR #669, opened 2022-10-26
migrate: Notify the user about the equal/like filter change

nilmerg:


sukhwinder33445 (approved):
Works for me.
### Tested in the following modules:

- Cube
- Graphite
- Notifications
- x509

PR #680, opened 2022-11-18
SlaReport: Add checkbox to optionally show critical SLA of hosts/serv…

jhoxhaa:
…ices reports


lippserd (commented):
Any chance to do that via SQL?

PR #686, opened 2022-11-23
Replace apt-get with apt

theFeu:
Apt-get has been deprecated for a while :)


RincewindsHat:
`apt-get` is getting deprecated?

theFeu:
Not sure what I was up to here, but I think this can be closed!

PR #699, opened 2023-01-16
Fix Error when dealing with wrong PerfData

EmTeedee:
If the perf data is not valid (e.g. `needrestart -p` with unknown values), a "non-numeric value encountered" error is thrown. This patch makes the check more specific to catch this error as well.

In our case, we run the `needrestart -p` command on raspbian with this output:

```
root@raspberry ~ # needrestart -p
UNKN - Kernel: 5.15.76-v7+!=5.15.84-v8+ (!!), Microcode: unknown, Services: none, Containers: none, Sessions: none|Kernel=2;0;;0;2 Microcode=U;0;;0;1 Services=0;;0;0 Containers=0;;0;0 Sessions=0;0;;0
```

The "U" in the Microcode value triggers the aforementioned error. The proposed fix should catches this.

### DIFF
--- a/library/Icingadb/Util/PerfData.php
+++ b/library/Icingadb/Util/PerfData.php
@@ -312,7 +312,8 @@ class PerfData
      */
     public function isVisualizable(): bool
     {
-        return isset($this->minValue) && isset($this->maxValue) && isset($this->value);
+        return is_numeric($this->minValue ?? null) && is_numeric($this->maxValue ?? null)
+            && is_numeric($this->value ?? null);
     }
 
     /**


nilmerg (changes requested):
https://www.monitoring-plugins.org/doc/guidelines.html#AEN201 states the following:

> 8. value, min and max in class [-0-9.]. Must all be the same UOM. value may be a literal "U" instead, this would indicate that the actual value couldn't be determined

So the performance data you have an issue with is perfectly legal.

The `U` however is identified as unit by our parser. This is wrong and must be fixed. The remaining code needs to account for that, including `PerfData::isVisualizable()`.

Though your current proposal does more than this. The numeric checks for min and max are currently not required so please avoid them.

So if you want to continue:

1. Create an issue showing the error and stacktrace (not as screenshot but in a code block)
1. Fix the parser
1. Account for the value being possibly `'U'` then

EmTeedee:
@nilmerg My fault, I didn't know U was valid, never encountered it before.
I created a new issue #700
I'll submit a new PR with a fix to the parser later today.

PR #701, opened 2023-01-17
Fix parsing unknown value in PerfData

### LINKED ISSUE #700: PerfData with "U" in value throws error
## Describe the bug
"U" as the value in performance data results in an error when viewing service status.

## To Reproduce
1. Create a dummy service with the following `dummy_text`: `Test|DataPoint=U;1;2;0;3`
2. Let the service check run and try to view the result in icingaweb

## Expected behavior
Service status displays as usual.

## Actual behavior
An error is displayed: `A non-numeric value encountered`

Stack:

```
#0 /usr/share/icingaweb2/modules/icingadb.tt/library/Icingadb/Util/PerfData.php(467): Icinga\Application\ApplicationBootstrap->Icinga\Application\{closure}()
#1 /usr/share/icingaweb2/modules/icingadb.tt/library/Icingadb/Util/PerfData.php(491): Icinga\Module\Icingadb\Util\PerfData->calculatePieChartData()
#2 /usr/share/icingaweb2/modules/icingadb.tt/library/Icingadb/Widget/Detail/PerfDataTable.php(108): Icinga\Module\Icingadb\Util\PerfData->asInlinePie()
#3 /usr/share/icinga-php/ipl/vendor/ipl/html/src/HtmlDocument.php(295): Icinga\Module\Icingadb\Widget\Detail\PerfDataTable->assemble()
#4 /usr/share/icinga-php/ipl/vendor/ipl/html/src/HtmlDocument.php(497): ipl\Html\HtmlDocument->ensureAssembled()
#5 /usr/share/icinga-php/ipl/vendor/ipl/html/src/HtmlDocument.php(339): ipl\Html\HtmlDocument->render()
#6 /usr/share/icinga-php/ipl/vendor/ipl/html/src/BaseHtmlElement.php(255): ipl\Html\HtmlDocument->renderUnwrapped()
#7 /usr/share/icinga-php/ipl/vendor/ipl/html/src/BaseHtmlElement.php(323): ipl\Html\BaseHtmlElement->renderContent()
#8 /usr/share/icinga-php/ipl/vendor/ipl/html/src/HtmlDocument.php(499): ipl\Html\BaseHtmlElement->renderUnwrapped()
#9 /usr/share/icinga-php/ipl/vendor/ipl/html/src/HtmlDocument.php(339): ipl\Html\HtmlDocument->render()
#10 /usr/share/icinga-php/ipl/vendor/ipl/html/src/BaseHtmlElement.php(255): ipl\Html\HtmlDocument->renderUnwrapped()
#11 /usr/share/icinga-php/ipl/vendor/ipl/html/src/BaseHtmlElement.php(323): ipl\Html\BaseHtmlElement->renderContent()
#12 /usr/share/icinga-php/ipl/vendor/ipl/html/src/HtmlDocument.php(499): ipl\Html\BaseHtmlElement->renderUnwrapped()
#13 /usr/share/icinga-php/ipl/vendor/ipl/html/src/HtmlDocument.php(339): ipl\Html\HtmlDocument->render()
#14 /usr/share/icinga-php/ipl/vendor/ipl/html/src/BaseHtmlElement.php(255): ipl\Html\HtmlDocument->renderUnwrapped()
#15 /usr/share/icinga-php/ipl/vendor/ipl/html/src/BaseHtmlElement.php(323): ipl\Html\BaseHtmlElement->renderContent()
#16 /usr/share/icinga-php/ipl/vendor/ipl/html/src/HtmlDocument.php(499): ipl\Html\BaseHtmlElement->renderUnwrapped()
#17 /usr/share/icinga-php/ipl/vendor/ipl/html/src/HtmlDocument.php(339): ipl\Html\HtmlDocument->render()
#18 /usr/share/icinga-php/ipl/vendor/ipl/html/src/HtmlDocument.php(499): ipl\Html\HtmlDocument->renderUnwrapped()
#19 /usr/share/icinga-php/ipl/vendor/ipl/web/src/Compat/ViewRenderer.php(56): ipl\Html\HtmlDocument->render()
#20 /usr/share/icingaweb2/library/vendor/Zend/Controller/Action/Helper/ViewRenderer.php(964): ipl\Web\Compat\ViewRenderer->render()
#21 /usr/share/icingaweb2/library/vendor/Zend/Controller/Action/HelperBroker.php(272): Zend_Controller_Action_Helper_ViewRenderer->postDispatch()
#22 /usr/share/icingaweb2/modules/icingadb.tt/library/Icingadb/Web/Controller.php(506): Zend_Controller_Action_HelperBroker->notifyPostDispatch()
#23 /usr/share/php/Icinga/Web/Controller/Dispatcher.php(76): Icinga\Module\Icingadb\Web\Controller->dispatch()
#24 /usr/share/icingaweb2/library/vendor/Zend/Controller/Front.php(937): Icinga\Web\Controller\Dispatcher->dispatch()
#25 /usr/share/php/Icinga/Application/Web.php(290): Zend_Controller_Front->dispatch()
#26 /usr/share/php/Icinga/Application/webrouter.php(105): Icinga\Application\Web->dispatch()
#27 /usr/share/icingaweb2/public/index.php(4): require_once(String)
#28 {main}
```

## Your Environment

* Icinga Web 2 version (System - About): 2.11.3
* Web browser: Firefox 108.0.2
* Icinga 2 version (`icinga2 --version`): r2.13.6-1
* Icinga DB version (`icingadb --version`): 1.1.0
* PHP version used (`php --version`): 7.4.33
* Server operating system and version: Debian 11

## Additional context
Performance data value definition: https://www.monitoring-plugins.org/doc/guidelines.html#AEN201

See also #699

EmTeedee:
'U' is neither a valid SI unit nor a valid prefix, so excluding it in the UOM part of the regex should not cause any issues.

The UOM regex now also accepts the `μ` prefix. Icinga normalizes values, but the regex seems to try to include them anyway.

If there is no numeric value, the pie chart can't show anything, so we test for that.

closes #700

### DIFF
--- a/library/Icingadb/Util/PerfData.php
+++ b/library/Icingadb/Util/PerfData.php
@@ -426,11 +426,15 @@ class PerfData
         $parts = explode(';', $this->perfdataValue);
 
         $matches = array();
-        if (preg_match('@^(-?(?:\d+)?(?:\.\d+)?)([a-zA-Z%°]{1,3})$@u', $parts[0], $matches)) {
+        if (preg_match('@^(U|-?(?:\d+)?(?:\.\d+)?)([a-zA-TV-Z%°]{1,3})$@u', $parts[0], $matches)) {
             $this->unit = $matches[2];
             $this->value = $matches[1];
         } else {
             $this->value = $parts[0];
+        }
+
+        if (! is_numeric($this->value)) {
+            $this->value = null;
         }
 
         switch (count($parts)) {


library/Icingadb/Util/PerfData.php
```diff
@@ -426,12 +426,15 @@ protected function parse()
         $parts = explode(';', $this->perfdataValue);
 
         $matches = array();
-        if (preg_match('@^(-?(?:\d+)?(?:\.\d+)?)([a-zA-Z%°]{1,3})$@u', $parts[0], $matches)) {
+        if (preg_match('@^((?:U|(?:-?\d+)?(?:\.\d+)?))([a-zA-TV-Z%°μ]{1,3})$@u', $parts[0], $matches)) {
```
nilmerg, line 429: The `-?` can't be moved inside the first non-capture group. `.9kW` is a valid value. See https://github.com/Icinga/icingaweb2/issues/4690 for details.

What's `A-TV-Z` for? The regex currently doesn't match `'U'` anyway. And even if it does, the `U` is consumed early and the range can't consume it again.

The `μ` is not necessary. You already mentioned Icinga's normalization, so why do you think it is?

Please change the regex so that the following matches: `-.9kW`, `UkW`, `-1.9kW`, `.9kW`, `1.9kW`

library/Icingadb/Util/PerfData.php
```diff
@@ -426,12 +426,15 @@ protected function parse()
...
             $this->unit = $matches[2];
             $this->value = $matches[1];
         } else {
             $this->value = $parts[0];
         }
+        if ($this->value === 'U') {
```
nilmerg, line 435: An empty line please after control structures.

nilmerg (changes requested):

EmTeedee:
While testing perfdata values, I noticed that min/max get unexpectedly dropped when there is no warn/crit. Bug: https://github.com/Icinga/icinga2/issues/9634

library/Icingadb/Util/PerfData.php
nilmerg: `U|` is not optional. In my tests this always consumes a leading `U`, no matter what follows. `-kW` isn't valid anyway and just `kW` as well. (makes no sense) And even if supplied this way, it's correctly identified as value (`-`) and uom (`kW`).

So I'm fine with the usage of `is_numeric` to decide whether to null the value again. But the `A-Z` range can be kept in combination with `U|` at the front.

I've also optimized it a bit: `@^(U|-?(?:\d+)?(?:\.\d+)?)([a-zA-Z%°]{1,3})$@u`

I've run unittests against this and I don't get any (new) errors. Or have you an example that doesn't work with this?

EmTeedee:
The `U|` is not optional, but everything on the other side of the pipe (`-?(?:\d+)?(?:\.\d+)?`) is.
Since the UOM in the regex requires 1-3 characters, it will consume the U in this simple perfdata:
`test=U` (which should be indicating that the value of U is unknown and has _no_ unit).

It doesn't show up in the UI, but a quick test shows that the U ends up in the unit field of PerfData
The value is correctly set to null (because `is_numeric` doesn't accept an empty string).

For testing the dummy_text can be `example|test=U`

I added a line to print the result in the parse function (`printf("value: %s / unit: %s\n", $this->value ?? 'NULL', $this->unit ?? 'NULL');`) result:
icinga_perfdata_unit_U

Here is an example with `php -a`, note that the `U` is stuck in `matches[2]`, which gets assigned to `$this->unit`:
```
php > $perfdata="U";
php > $parts = explode(';', $perfdata); $matches = array(); preg_match('@^(U|-?(?:\d+)?(?:\.\d+)?)([a-zA-Z%°]{1,3})$@u', $parts[0], $matches);
php > var_dump($matches);
array(3) {
  [0]=>
  string(1) "U"
  [1]=>
  string(0) ""
  [2]=>
  string(1) "U"
}
```

I've perused the icinga2 source and found that perfdata is parsed by looking for the last `[.0-9]` in the value and treating everything after that as UOM and the front as value. Maybe that would be a better approach rather than the regex?
see https://github.com/Icinga/icinga2/blob/master/lib/base/perfdatavalue.cpp#L252-L258

I've also removed the offending commit.

nilmerg:
Ah, I didn't test just with an `U` :roll_eyes: 

> Maybe that would be a better approach rather than the regex?

The remaining regex has been proven correct in recent years. I'd rather not replace it just because of a minor enhancement ;)

It's fine now. But please remember to restore the `isset` call in `isVisualizable`.

nilmerg (approved):

PR #708, opened 2023-01-26
SlaReport: Handle `null` sla results correctly

yhabteab:
With [this PR](https://github.com/Icinga/icingadb/pull/560) of icingadb the sla might be `null` when generating reports within a timeframe for which no history events can be found, causing a fatal.

### Before
Bildschirm­foto 2023-01-26 um 11 28 45

### After
Bildschirm­foto 2023-01-26 um 11 27 50


yhabteab:
Superseded by https://github.com/Icinga/icingadb-web/pull/710

PR #710, opened 2023-02-03
WIP

yhabteab:


yhabteab:
refs https://github.com/Icinga/icingaweb2-module-reporting/pull/180

yhabteab:
refs https://github.com/Icinga/icingaweb2-module-reporting/pull/181

yhabteab:
Bildschirm­foto 2023-02-13 um 11 10 24

library/Icingadb/ProvidedHook/Reporting/Common/SlaTimeline.php
```diff
@@ -0,0 +1,204 @@
...
+        $totalTime = (int) $this->end->format('Uv') - $lastEventTime;
+
+        $lastHardState = $this->initialHardState;
+        $count = $this->count();
+        for ($i = 0; $i < $count; ++$i) {
+            $event = $this->event[$i];
+            $time = $this->time[$i];
+            $state = $this->state[$i];
+
+            if ($this->previousState[$i] === 99 || ($lastHardState === 99 && $event !== 'state_change')) {
```
julianbrost, line 81: This condition is changed to compared to the current one: https://github.com/Icinga/icingadb/blob/4ed4db3ab642622bb917797a8207589ce5f89f07/schema/mysql/schema.sql#L139

I think this is the change that's actually supposed to address https://github.com/Icinga/icingadb/issues/558. If we want to mix this in here, I'd try change the order of actions in the loop, the current condition was my suggestion for a quick test to see if this would bring an improvement. The general idea was to check if `$lastHardState` would be 99 at the end of the iteration. So it could make sense to just move this branch of the conditional after the update to `$lastHardState` (the update of `$lastEventTime` would have to be moved behind that then).

test/php/library/Icingadb/ProvidedHook/Reporting/SlaTest.php
```diff
@@ -0,0 +1,338 @@
...
+
+    public function testCriticalBeforeIntervalWithOverlappingDowntimes()
+    {
+        $this->insertSlaHistoryEvents([
+            'state'    => [['event_time' => 800000, 'hard_state' => 2, 'previous_hard_state' => 99]],
+            'downtime' => [
+                ['downtime_id' => $this->makeId(), 'downtime_start' => 600000, 'downtime_end' => 1000000],
+                ['downtime_id' => $this->makeId(), 'downtime_start' => 800000, 'downtime_end' => 1200000],
+                ['downtime_id' => $this->makeId(), 'downtime_start' => 1000000, 'downtime_end' => 1400000],
+                ['downtime_id' => $this->makeId(), 'downtime_start' => 1600000, 'downtime_end' => 2000000], // -20%
```
julianbrost, line 148: The -20% comment here on a downtime line makes little sense as downtimes never reduce the availability percentage.

julianbrost (commented):
I'm focusing only on the algorithm itself as I'm not very familiar with PHP and IPL.

General remarks:

* The [query for the timeline itself](https://github.com/Icinga/icingadb/blob/4ed4db3ab642622bb917797a8207589ce5f89f07/schema/mysql/schema.sql#L30-L81) is now split across two files (one with the column definitions and one with the filters). That would probably be better to understand if it was possible to keep this in one place and include comments.
* If I understand correctly, in case of a breakdown report, this currently fetches the timeline for each sub-interval separately. Generally, since now instead of a single value, the entire SLA history is fetches, it would be possible to do all this in a single query and split the result into the sub-intervals in PHP (could be an optimization, someone else has to judge if this is worth putting effort into).

The test data looks sensible but I didn't compare every single detail with the current test data on the Go side. Are there any intentional changes, especially to accommodate the change in handling 99/pending?

Also to anyone else reviewing this: if there are any question regarding the algorithm, just ask me.

test/php/library/Icingadb/ProvidedHook/Reporting/SlaTest.php
yhabteab: The comment should indicate the time period between the current downtime start event and the previous downtime end event. `downtime_end => 1400.....downtime_start => 1600`

library/Icingadb/ProvidedHook/Reporting/Common/SlaTimeline.php
yhabteab: > So it could make sense to just move this branch of the conditional after the update to `$lastHardState` (the update of `$lastEventTime` would have to be moved behind that then).

`$lastHardState` is only updated when the current row event type is a `state_change`. How would this make any difference compared the current implementation?

library/Icingadb/ProvidedHook/Reporting/Common/SlaTimeline.php
```diff
@@ -0,0 +1,204 @@
...
+                $totalTime -= $time - $lastEventTime;
+            } elseif ($lastHardState > 0 && $lastHardState !== 99 && $activeDowntimes === 0) {
```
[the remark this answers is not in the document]
yhabteab: Overseen the 🐘

library/Icingadb/ProvidedHook/Reporting/Common/SlaTimeline.php
julianbrost: It should do exactly the same. My hope is that the change makes the code easier to understand.

test/php/library/Icingadb/ProvidedHook/Reporting/SlaTest.php
julianbrost: That's not really related to that specific downtime but rather with a gap between the downtimes. That should be clearer when keeping the comment similar to how it was before: https://github.com/Icinga/icingadb/blob/4ed4db3ab642622bb917797a8207589ce5f89f07/tests/sql/sla_test.go#L91-L99

yhabteab:
> Are there any intentional changes, especially to accommodate the change in handling 99/pending?

Yes! https://github.com/Icinga/icingadb-web/pull/710/commits/8eb216548b78e1563c0d744d5a91545c0a1d18b2

PR #745, opened 2023-05-16
AcknowledgeProblemForm: Don't show datepicker if `expire` checkbox is unchecked

### LINKED ISSUE #743: settings.acknowledge_expire = true overrides "Use Expire Time" toggle in UI
## Describe the bug

If the  `settings.acknowledge_expire` option is set to `true`, the "Use Expire Time" toggle in the acknowledgement UI does not function to hide the "Expire Time" date picker. If the date picker is manually cleared, the form cannot be submitted due to the empty field.

## To Reproduce

Provide a link to a live example, or an unambiguous set of steps to reproduce this bug. Include configuration, logs, etc. to reproduce, if relevant.

1. Add the following block to `/etc/icingaweb2/modules/icingadb/config.ini`:
```ini
[settings]
acknowledge_expire = true
```
3. Acknowledge a service
4. The "Use Expire Time" toggle should already be set to true. Change it to false.
5. If a time was set in "Expire Time", the acknowledgement will expire at that time despite "Use Expire Time" being set to false.
6. If the "Expire Time" field was cleared, the acknowledgement cannot be submitted.

## Expected behavior

It should be possible to submit an acknowledgement without an expire time by disabling the "Use Expire Time" toggle.

## Screenshots

Default state of the acknowledgement box when `acknowledge_expire` is true - this is as expected:
Screenshot_20230512_153119

Toggling the "Use Expire Time" option off does not make the date picker disappear:
Screenshot_20230512_153134

If the date picker is cleared, the acknowledgement cannot be submitted:
Screenshot_20230512_153148

If the date picker is not cleared, an acknowledgement is created that expires on that date.
Screenshot_20230512_153210

## Your Environment

Include as many relevant details about the environment you experienced the problem in

* Icinga Web 2 version (System - About): 2.11.4
* Web browser: Chrome
* Icinga 2 version (`icinga2 --version`): r2.13.7-1
* Icinga DB version (`icingadb --version`): v1.1.0
* PHP version used (`php --version`): PHP 7.4.33 (cli) (built: Feb 22 2023 20:07:47) ( NTS )
* Server operating system and version: Docker, images based on Debian bullseye using https://packages.icinga.com/debian

## Additional context

* This worked as expected in the old `monitoring` module for icingaweb2.
* We also set `acknowledge_expire_time = PT24H`. This shouldn't have any effect on the desired behaviour.

## Baseless speculation

Looks like this if statement uses the value of the config option instead of the toggle in the UI to control whether the date picker is visible. 
https://github.com/Icinga/icingadb-web/blob/9041a646c73e79c053770722a5d1993996717a87/application/forms/Command/Object/AcknowledgeProblemForm.php#L144

The value of the date picker appears to ultimately be what controls whether an expiry time is set, rather than the value of the toggle.
https://github.com/Icinga/icingadb-web/blob/9041a646c73e79c053770722a5d1993996717a87/application/forms/Command/Object/AcknowledgeProblemForm.php#L196-L199

sukhwinder33445:
fixes #743

### DIFF
--- a/application/forms/Command/Object/AcknowledgeProblemForm.php
+++ b/application/forms/Command/Object/AcknowledgeProblemForm.php
@@ -126,22 +126,20 @@ class AcknowledgeProblemForm extends CommandForm
         );
         $decorator->decorate($this->getElement('sticky'));
 
-        $acknowledgeExpire = (bool) $config->get('settings', 'acknowledge_expire', false);
-
         $this->addElement(
             'checkbox',
             'expire',
             [
                 'ignore'        => true,
                 'class'         => 'autosubmit',
-                'value'         => $acknowledgeExpire,
+                'value'         => (bool) $config->get('settings', 'acknowledge_expire', false),
                 'label'         => t('Use Expire Time'),
                 'description'   => t('If the acknowledgement should expire, check this option.')
             ]
         );
         $decorator->decorate($this->getElement('expire'));
 
-        if ($acknowledgeExpire || $this->getPopulatedValue('expire') === 'y') {
+        if ($this->getElement('expire')->isChecked()) {
             $expireTime = new DateTime();
             $expireTime->add(new DateInterval($config->get('settings', 'acknowledge_expire_time', 'PT1H')));
 
--- a/application/forms/Command/Object/AddCommentForm.php
+++ b/application/forms/Command/Object/AddCommentForm.php
@@ -82,7 +82,6 @@ class AddCommentForm extends CommandForm
         $decorator->decorate($this->getElement('comment'));
 
         $config = Config::module('icingadb');
-        $commentExpire = (bool) $config->get('settings', 'comment_expire', false);
 
         $this->addElement(
             'checkbox',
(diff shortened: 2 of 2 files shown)


application/forms/Command/Object/AcknowledgeProblemForm.php
```diff
@@ -126,22 +126,20 @@ protected function assembleElements()
...
-                'value'         => $acknowledgeExpire,
+                'value'         => (bool) $config->get('settings', 'acknowledge_expire', false),
                 'label'         => t('Use Expire Time'),
                 'description'   => t('If the acknowledgement should expire, check this option.')
             ]
         );
         $decorator->decorate($this->getElement('expire'));
 
-        if ($acknowledgeExpire || $this->getPopulatedValue('expire') === 'y') {
+        if ($this->getValue('expire') === 'y') {
```
nilmerg, line 142: Please use `getElement()->isChecked()`

nilmerg (changes requested):
`AddCommentForm` is also affected.

nilmerg (approved):

PR #749, opened 2023-05-17
CheckStatistics: Show `retry_interval` if object in `problem && soft_state`

### LINKED ISSUE #684: Display retry interval when in soft state
Check statistics in the detail view should display the retry interval instead of the check interval when in soft state:
icingadb_check_statistics

In this case:
* check_interval = 60
* retry_interval = 30

# Environment
* Icinga Web: 2.11.2
* Icinga DB Web: 1.0.2
* Icinga DB: 1.1.0

sukhwinder33445:
fixes #684

### DIFF
--- a/library/Icingadb/Widget/Detail/CheckStatistics.php
+++ b/library/Icingadb/Widget/Detail/CheckStatistics.php
@@ -40,11 +40,12 @@ class CheckStatistics extends Card
 
         $overdueBar = null;
         $nextCheckTime = $this->object->state->next_check;
+        $checkInterval = $this->getCheckInterval();
         if ($this->object->state->is_overdue) {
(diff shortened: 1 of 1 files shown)


nilmerg (changes requested):
The progress is also incorrectly calculated:
image

library/Icingadb/Widget/Detail/CheckStatistics.php
```diff
@@ -120,13 +121,14 @@ protected function assembleBody(BaseHtmlElement $body)
                     : t('PENDING'))
             )
         );
+
```
nilmerg, line 124: Please remove this.

nilmerg (approved):

PR #754, opened 2023-05-22
Unset pagination parameters for service grid when `limit=0,0`

### LINKED ISSUE #714: Full Grid display lost on IcingaWeb2 2.11.3
## Describe the bug

- In IcingaWeb2 2.10.1, I have a dashlet pointing to: monitoring/list/servicegrid?limit=0%2C0 which displaying every hosts for every services
- In IcingaWeb2 2.11.3, I have set the same dashlet pointing to: icingadb/services/grid?limit=0%2C0 which displaying the default grid with limit=20 by x and by y

## To Reproduce
just create your dahslet with URL described above

## Expected behavior
Keep the same as before : a way to remove the default limit=20

## Screenshots
image

## Env
TAG_REDIS="7.0.5-alpine"
TAG_MARIADB="10.9.4"
TAG_ICINGA2="2.13.6"
TAG_ICINGADB="1.1.0"
TAG_ICINGAWEB="2.11"
Chrome 110.0.5481.78

raviks789:
The pagination parameters must be unset when the limit parameter is `0,0` and the service grid for all the hosts and services must be displayed.

fixes #714

### DIFF
--- a/application/controllers/ServicesController.php
+++ b/application/controllers/ServicesController.php
@@ -22,6 +22,7 @@ use Icinga\Module\Icingadb\Widget\ItemTable\ServiceItemTable;
 use Icinga\Module\Icingadb\Widget\ServiceStatusBar;
 use Icinga\Module\Icingadb\Widget\ShowMore;
 use Icinga\Module\Icingadb\Web\Control\ViewModeSwitcher;
+use Icinga\Util\Environment;
 use ipl\Html\HtmlString;
 use ipl\Orm\Query;
 use ipl\Stdlib\Filter;
@@ -222,6 +223,8 @@ class ServicesController extends Controller
 
     public function gridAction()
     {
+        Environment::raiseExecutionTime();
+
         $db = $this->getDb();
         $this->addTitleTab(t('Service Grid'));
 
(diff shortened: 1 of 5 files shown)


nilmerg (changes requested):
The changes you made leave the impression that you didn't quite know what to fix here.

The root cause is that a `LIMIT 0` on the selects return no results. The easy fix for this is not to apply such a limit. This is done in `PivotTable::paginateXAxis()` for example. No other changes needed for the x axis query.

Sure, this leads to further issues since most of the remaining code expects the limit not to be zero, or not set. But that needs to be accounted for then.

However, it is absolutely not acceptable to count all resulting rows first and then query them based on this count, just to avoid setting no limit or a zero limit. You've already showed me that the performance is inferior to what the monitoring module does in the same case, and you didn't think that counting first lowers it even further?

Please reduce your changes to what's absolutely required. Your current ones are not.

application/controllers/ServicesController.php
nilmerg, line 1: You only need to raise the execution time limit here

library/Icingadb/Data/PivotTable.php
nilmerg, line 1: You don't have to change `getPaginationParameter` at all. Only `paginateXAxis` and `paginateYAxis`.

nilmerg (changes requested):

PR #755, opened 2023-05-23
History: Sort also by type by default

### LINKED ISSUE #587: Additionally order history events by type

At the moment, history events seem to be ordered only by time. For events with the same timestamp, there seems to be no tie-breaker and events show up in some order (probably some implementation-defined ordering from the database).

I suggest additionally ordering events with the same timestamp by type in a way that reflects the typical order of causation. That is state changes should come first (can trigger downtimes, flapping events, acknowledgement removals) and notifications last (as they typically are the result of something that just happened).

nilmerg:
resolves #587

### DIFF
--- a/library/Icingadb/Model/History.php
+++ b/library/Icingadb/Model/History.php
@@ -63,7 +63,7 @@ class History extends Model
 
     public function getDefaultSort()
     {
-        return 'history.event_time desc';
+        return 'history.event_time desc, history.event_type desc';
     }
 
     public function createBehaviors(Behaviors $behaviors)
--- a/application/controllers/HistoryController.php
+++ b/application/controllers/HistoryController.php
@@ -51,7 +51,7 @@ class HistoryController extends Controller
         $sortControl = $this->createSortControl(
             $history,
             [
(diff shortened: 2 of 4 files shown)


julianbrost (commented):
So the overall plan here is to sort it as good as easily possible with `event_type` (i.e. without defining a separate sort order within the web module) and if other tweaks to the sort order are needed, these have to be done as a schema change?

Anyway, this change shouldn't cause problems as a sort order returned for `order by a, b` would also be valid for `order by a`.

I gave this a quick test (manually patched the file in my running container): no surprises regarding the sort order itself, but now the sort dropdown shows "Custom".

library/Icingadb/Model/History.php
```diff
@@ -63,7 +63,7 @@ public function getColumnDefinitions()
 
     public function getDefaultSort()
     {
-        return 'history.event_time desc';
+        return 'history.event_time desc, history.event_type desc';
```
raviks789, line 66: Use this sort also for `$sortControl` in `HistoryController:indexAction` or else as @julianbrost mentioned sort dropdown shows "Custom"

PR #756, opened 2023-05-24
Enhance global search

### LINKED ISSUE #690: find nothing of hosts when do global searching based on display name or address
## Describe the bug
Hello,

when doing a global search it will search in the old monitoring in the object, address and display_name for me.
In icingadb it will only search in the object name I think.

## To Reproduce
Disable the old monitoring that you only work in icingadb mode.
Search for something that is in the display_name or address , but is not in the object name (name or name_ci)

## Expected behavior
Find the horsts that find the text in the display name or address.

## Your Environment

* Icinga DB Web version (System - About): 2.11.2
* Icinga Web 2 version (System - About): 1.0.2
* Web browser: chrome
* Icinga 2 version (`icinga2 --version`): r2.13.6-1
* Icinga DB version (`icingadb --version`): v1.1.0
* PHP version used (`php --version`): 7.1.30
* Server operating system and version: Red Hat Enterprise Linux Server 7.9

sukhwinder33445:
fixes #690

### DIFF
--- a/library/Icingadb/Web/Controller.php
+++ b/library/Icingadb/Web/Controller.php
@@ -299,10 +299,11 @@ class Controller extends CompatController
      * Process a search request
      *
      * @param Query $query
+     * @param array $additionalColumns
      *
      * @return void
      */
-    public function handleSearchRequest(Query $query)
+    public function handleSearchRequest(Query $query, array $additionalColumns = [])
     {
         $q = trim($this->params->shift('q', ''), ' *');
         if (! $q) {
@@ -310,7 +311,7 @@ class Controller extends CompatController
         }
 
         $filter = Filter::any();
-        $this->prepareSearchFilter($query, $q, $filter);
+        $this->prepareSearchFilter($query, $q, $filter, $additionalColumns);
 
         $redirectUrl = Url::fromRequest();
         $redirectUrl->setQueryString(QueryString::render($filter));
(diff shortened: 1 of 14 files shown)


library/Icingadb/Web/Controller.php
```diff
@@ -327,12 +328,13 @@ public function handleSearchRequest(Query $query)
...
      * @param Filter\Any $filter
+     * @param array $columns
      *
      * @return void
      */
-    protected function prepareSearchFilter(Query $query, string $search, Filter\Any $filter)
+    protected function prepareSearchFilter(Query $query, string $search, Filter\Any $filter, array $columns)
     {
-        foreach ($query->getModel()->getSearchColumns() as $column) {
+        foreach ($columns as $column) {
```
nilmerg, line 337: `$columns` may be empty. If one doesn't pass them, not any column is used. Please use the model's search columns still as a fallback.

nilmerg (changes requested):
Seeing it now, I think we should still define most of the columns in the models as search columns. The problem with the current approach is that the search bar is not affected by it. And to search for the display name is also useful there.

The only exception I see, is the `address` and `address6` columns for the hosts overview. These should be exclusive to the global search. So the change in `HostsController` can be kept.

nilmerg (changes requested):
There may be no search integration for users and groups, but the search bar is still there and the display name should be one of the defaults there as well.

nilmerg (approved):

PR #759, opened 2023-05-25
Don't use table layout for group and user overviews

### LINKED ISSUE #192: `.item-table`s are still prone to overflow due long subjects

nilmerg:
fixes #192


flourish86 (commented):
Users and Usergroups are fine.

Host/Service Groups’ layout breaks, when there’s extreme content in the right aligned elements.

Screenshot 2023-05-25 at 15 51 30

Also even moderate differences in host and service counts make the layout uneven.

Screenshot 2023-05-25 at 16 00 17

nilmerg:
Fixed the total amounts to a maximum of 4 chars now.

PR #761, opened 2023-05-25
Fix/table list

flourish86:
Here’s a solution to the overflow problem.
UI wise this is the superior solution, event if it uses excess markup. Most users will never see the latter.

Still, what’s not solved is the overflow for the Host/Service Counts, but it should also work.

Screenshot 2023-05-25 at 17 00 51


nilmerg:
Sorry, table layout stinks.

PR #769, opened 2023-05-31
ObjectSuggestions: Improve column suggestions

nilmerg:


nilmerg:
Not near of being final yet.

Using `->isOne()` to detect candidate relations is far from perfect. It produces too much noise. (e.g. `comment.service_state`, even worse: it causes `service.state` to not appear!) There are also some duplicates (e.g. search for `user`).

A better heuristic to determine important relations is needed. Or some way to flag relations as only being *technical*, `comment.service_state` would be one example for this.

PR #772, opened 2023-06-05
Display downtime icon independent from object state

### LINKED ISSUE #706: Symbol doesn't show that Host/Service is in Downtime when it's okay
Hi,
in the past, using the monitoring module, our admins could see their hosts listed with symbols (a key symbol) when they are in Outage.
It's really nice, because one look and you see, that you can begin the work on the hosts, because it's in downtime.

Now, with the icingadb-web Module, you see in the 'list-view' that the host is in downtime, but ONLY, when it's not in okay state.

sukhwinder33445:
resolves #706


flourish86 (changes requested):
Please take a look. The icon in the detail header ist placed in the wrong visual. It should take place in the current state visual, like in the list.

Screenshot 2023-06-05 at 15 34 11

nilmerg (changes requested):
Please implement `State::getIcon(): ?Icon`, should simplify this change significantly.

library/Icingadb/Model/State.php
```diff
@@ -78,4 +80,47 @@ public function createBehaviors(Behaviors $behaviors)
...
+     */
+    public function getIcon(): ?Icon
+    {
+        $icon = null;
+        if ($this->is_handled) {
+            switch (true) {
+                case $this->is_acknowledged:
+                    $icon = new Icon(Icons::IS_ACKNOWLEDGED);
+                    break;
+                case $this->in_downtime:
```
nilmerg, line 97:
```suggestion
                case $this->is_handled && $this->in_downtime:
```

library/Icingadb/Model/State.php
```diff
@@ -78,4 +80,47 @@ public function createBehaviors(Behaviors $behaviors)
...
+                            '%s (%s)',
+                            strtoupper($this->getStateTextTranslated()),
+                            t('handled by Downtime')
+                        )]
+                    );
+                    break;
+                case $this->is_flapping:
+                    $icon = new Icon(Icons::IS_FLAPPING);
+                    break;
+                default:
```
nilmerg, line 110:
```suggestion
                case $this->is_handled:
```

library/Icingadb/Model/State.php
```diff
@@ -78,4 +80,47 @@ public function createBehaviors(Behaviors $behaviors)
...
+                    $icon = new Icon(Icons::HOST_DOWN);
+            }
+        } elseif ($this->in_downtime) {
```
nilmerg, line 113: Please add this with `! $this->is_handled && $this->in_downtime` to the switch case above.

library/Icingadb/Model/State.php
```diff
@@ -78,4 +80,47 @@ public function createBehaviors(Behaviors $behaviors)
...
+            switch (true) {
+                case $this->is_acknowledged:
+                    $icon = new Icon(Icons::IS_ACKNOWLEDGED);
+                    break;
+                case $this->in_downtime:
+                    $icon = new Icon(
+                        Icons::IN_DOWNTIME,
+                        ['title' => sprintf(
+                            '%s (%s)',
+                            strtoupper($this->getStateTextTranslated()),
```
nilmerg, line 102: why strtoupper? We don't do this anywhere else.

library/Icingadb/Widget/StateChange.php
```diff
@@ -55,9 +58,15 @@ public function setCurrentStateBallSize(string $size): self
         return $this;
     }
 
+    public function isReversed(): bool
```
nilmerg, line 61: Please drop this and add the icon here directly. The state knows which icon to use now, so use it please.

nilmerg (changes requested):

library/Icingadb/Widget/StateChange.php
sukhwinder33445: State (Model) does not exist here.

nilmerg: uhm, okay, but please get rid of this method. That's an implementation detail no-one should care about. Consider a `setIcon` method for example.

nilmerg (approved):

PR #773, opened 2023-06-05
Fix negative next check value when both active and passive checks are disabled

### LINKED ISSUE #403: CheckStatistics: Reflect disabled active checks
## Describe the bug
Disabling active checks afftects the “next check” value from the check statistics widget. Currently this leads to negative values for the “next check” or when the view is updated wrong values

## To Reproduce
1. Open a service/host detail view
2. disabled active checks
3. Wait until the check interval is over to see the negative “Next Check” value

## Expected behavior
We should consider the Check Statistics widget to reflect that active checks are disabled.

## Screenshots
Screen Shot 2021-10-20 at 13 55 45

raviks789:
resolves #403

### DIFF
--- a/library/Icingadb/Widget/Detail/CheckStatistics.php
+++ b/library/Icingadb/Widget/Detail/CheckStatistics.php
@@ -132,18 +132,30 @@ class CheckStatistics extends Card
                     : (new EmptyState(t('n. a.')))->setTag('span')
             )
         );
+
+        if ($this->isChecksDisabled()) {
+            $nextCheckBubbleContent = new VerticalKeyValue(
+                t('Next Check'),
+                t('n.a')
+            );
+
+            $this->addAttributes(['class' => 'checks-disabled']);
+        } else {
+            $nextCheckBubbleContent = $this->object->state->is_overdue
+                ? new VerticalKeyValue(t('Overdue'), new TimeSince($nextCheckTime))
+                : new VerticalKeyValue(
+                    t('Next Check'),
+                    $nextCheckTime !== null ? new TimeUntil($nextCheckTime) : t('PENDING')
+                );
+        }
+
         $nextCheck = Html::tag(
             'li',
             ['class' => 'end'],
             Html::tag(
                 'div',
                 ['class' => 'bubble upwards'],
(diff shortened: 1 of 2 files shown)


library/Icingadb/Widget/Detail/CheckStatistics.php
```diff
@@ -104,13 +104,6 @@ protected function assembleBody(BaseHtmlElement $body)
             'style' => sprintf('left: %F%%', $hPadding + $leftNow),
         ]);
 
-        $timeline->add([
-            $markerLast,
-            $markerNow,
-            $markerNext,
-            $overdueBar
-        ]);
```
nilmerg, line 112: Please restore this, you don't have to reorder it.

library/Icingadb/Widget/Detail/CheckStatistics.php
```diff
@@ -132,18 +125,52 @@ protected function assembleBody(BaseHtmlElement $body)
                     : (new EmptyState(t('n. a.')))->setTag('span')
             )
         );
+
+        if ($this->isChecksDisabled()) {
+            $markerLast->addAttributes([
+                'class' => 'state-' . $this->object->state->getStateText()
+            ]);
```
nilmerg, line 132: I don't think I talked about this change with @flourish86, that's just an artifact in the mockup. I see absolutely no reason to introduce this change here, it has nothing to do with active/passive checks being off.

public/css/mixin/progress-bar.less
```diff
@@ -177,3 +177,49 @@
...
 }
+
+.progress-bar.checks-disabled {
+  .above {
+    .checks-disabled-text {
+      display: block;
+      width: 100%;
+      text-align: center;
+    }
+  }
```
nilmerg, line 188: This is no general style. Please move this to `check-statistics.less`.

nilmerg (changes requested):

public/css/mixin/progress-bar.less
raviks789: The disabled note is moved to overlay container now.

nilmerg (approved):

PR #779, opened 2023-06-19
Render/suggest empty custom vars correctly

### LINKED ISSUE #778: Render empty customvars correctly
## Describe the bug

Empty custom vars of type `Array` (`[]`) and `Dictionary` (`{}`) aren't rendered correctly. They're just displayed as `[]` and `{}` while the monitoring module renders them correctly.

Bildschirmfoto 2023-06-13 um 17 34 51

Bildschirmfoto 2023-06-13 um 16 56 20

refs https://github.com/Icinga/icingadb/pull/601

yhabteab:
fixes #778

### DIFF
--- a/library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
+++ b/library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
@@ -212,15 +212,16 @@ class ObjectSuggestions extends Suggestions
 
         // Custom variables only after the columns are exhausted and there's actually a chance the user sees them
         $titleAdded = false;
+        $parsedArrayVars = [];
         foreach ($this->getDb()->select($this->queryCustomvarConfig($searchTerm)) as $customVar) {
             $search = $name = $customVar->flatname;
-            if (preg_match('/\w+\[(\d+)]$/', $search, $matches)) {
-                // array vars need to be specifically handled
-                if ($matches[1] !== '0') {
+            if (preg_match('/\w+(?:\[(\d*)])+$/', $search, $matches)) {
+                $name = substr($search, 0, -(strlen($matches[1]) + 2));
+                if (isset($parsedArrayVars[$name])) {
                     continue;
                 }
 
-                $name = substr($search, 0, -3);
+                $parsedArrayVars[$name] = true;
                 $search = $name . '[*]';
             }
 


library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -214,13 +214,11 @@ protected function fetchColumnSuggestions($searchTerm)
...
-                // array vars need to be specifically handled
-                if ($matches[1] !== '0') {
-                    continue;
-                }
-
-                $name = substr($search, 0, -3);
+            if (preg_match('/\w+\[(\d*)]$/', $search, $matches)) {
+                // Count the digits of the used array index and add the len of the curly brackets (2). The hardcoded
+                // previously used value `-3` may fail for array indices `>= 10`.
+                $length = $matches[1] === "" ? 2 : (int) log10((int) $matches[1]) + 1 + 2;
```
julianbrost, line 220: Isn't that just a complicated (and maybe even buggy) way of writing `$length = strlen($matches[1]) + 2;`

> The hardcoded previously used value `-3` may fail for array indices `>= 10`.

That sounds more like something for a commit message rather than an inline comment (there's no trace of -3 in the new code).

yhabteab: > Isn't that just a complicated (and maybe even buggy) way of writing `$length = strlen($matches[1]) + 2;`

Don't see what makes these changes " buggy". Yes, it might be quite complex for such a trivial task, but after all it only performs a simple mathematical operation.

julianbrost:
```
php > var_dump((int) log10((int) "0"));
int(0)
php > var_dump(log10((int) "0"));
float(-INF)
```

I mean probably works, but only because of some PHP quirks.

You want the length of a string to remove it from some other string, why would you want to resort to some floating point number calculations when you can just use the length directly?

library/Icingadb/Model/CustomvarFlat.php
```diff
@@ -113,7 +113,10 @@ public function unFlattenVars(Traversable $flattenedVars): array
 
                 $registerValue($data[$step], $source[$step] ?? null, $path, $value);
             } else {
-                $data[$step] = $value;
+                // Since empty custom vars of type dictionaries and arrays have null values in customvar_flat table,
+                // we won't be able to render them as such. Therefore, we have to use the value of the `customvar`
+                // table if it's not null, otherwise the current value.
```
nilmerg, line 118: And what's `$value` if the value in the customvar table is null? (Missing in the comment)

yhabteab:
Bildschirmfoto 2023-06-23 um 12 16 22

library/Icingadb/Model/CustomvarFlat.php
```diff
@@ -130,7 +134,33 @@ public function unFlattenVars(Traversable $flattenedVars): array
...
+
+                    // Join the custom var paths with the real name in reverse order.
+                    $path = array_merge([$realName], array_reverse($sourcePaths));
+                } else {
+                    if ($sourcePath === '[]') {
+                        $sourcePath = null;
+                    }
+
+                    $path = array_merge([$realName], $sourcePath ? explode('.', $sourcePath) : []);
+                }
```
nilmerg, line 163: Should suffice, please test:

```suggestion
                $path = array_merge([$realName], $sourcePath ? preg_split('/(?<=\w)\.|(?=\[)/', $sourcePath) : []);
```

library/Icingadb/Model/CustomvarFlat.php
```diff
@@ -97,7 +97,8 @@ public function unFlattenVars(Traversable $flattenedVars): array
             $step = array_shift($path);
 
             $isIndex = (bool) preg_match('/^\[(\d+)]$/', $step, $m);
-            if ($isIndex) {
+            // Don't compute custom var names like `vars."[4]"` as array index access.
+            if (! isset($source[$step]) && $isIndex) {
```
nilmerg, line 101: Then you don't have to perform the preg_match, do you?

```suggestion
            $isIndex = ! isset($source[$step]) && (bool) preg_match('/^\[(\d+)]$/', $step, $m);
            if ($isIndex) {
```

nilmerg (changes requested):

library/Icingadb/Model/CustomvarFlat.php
yhabteab: **Without your changes:**
Bildschirmfoto 2023-06-23 um 15 34 16

**With your changes:**
Bildschirmfoto 2023-06-23 um 15 33 30

nilmerg:
```suggestion
                $path = array_merge([$realName], $sourcePath ? preg_split('(?<=\w|\])\.|(?=\[)', $sourcePath) : []);
```

julianbrost:
One thing I noticed in the screenshot: empty arrays seem to miss the "(Array)" hint.

nilmerg:
That's unavoidable. PHP cannot differentiate between an empty dict and an empty list.

julianbrost:
Unavoidable as in you're reconstructing the whole `vars` data structure with PHP data types and Icinga 2 arrays and dicts are both represented by PHP arrays? And if there's an empty PHP array, you don't know whether it was an array or dict in Icinga 2?

nilmerg:
@yhabteab Please check once more. Should be working fine now.

nilmerg:
Didn't look at your/my code changes yet, but testing revealed this: 
image

nilmerg (approved):

PR #784, opened 2023-06-21
Avoid leakage of `state.check_commandline` to restricted users

raviks789:
Users who do not have permission to see the object's `Source` tab, must be restricted from accessing the object's `state.check_commandline` column.

fixes [ref/IP/44766](https://rt.icinga.com/Ticket/Display.html?id=44766)

### DIFF
--- a/library/Icingadb/Common/Auth.php
+++ b/library/Icingadb/Common/Auth.php
@@ -137,6 +137,9 @@ trait Auth
                 // Hence why the hosts restriction is also applied if only services are queried.
                 || $applyServiceRestriction;
 
+            $hostStateRelation = array_search('host_state', $relations, true);
+            $serviceStateRelation = array_search('service_state', $relations, true);
+
             $resolver = $query->getResolver();
 
             $queryFilter = Filter::any();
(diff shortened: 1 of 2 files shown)


library/Icingadb/Common/Auth.php
```diff
@@ -196,6 +199,40 @@ public function applyRestrictions(Query $query)
...
+                        $hostStateRelation
+                            ? $resolver->qualifyColumn('check_commandline', $hostStateRelation)
+                            : 'state.check_commandline'
+                    ];
+                }
+
+                if ($serviceStateRelation !== false) {
+                    $commandColumns[] = $serviceStateRelation
+                        ? $resolver->qualifyColumn('check_commandline', $serviceStateRelation)
+                        : 'state.check_commandline';
```
nilmerg, line 218: same here

nilmerg (changes requested):
I've removed the changes from `createColumnControl`. It's been `VolatileStateResults` which reapplied the commandline value read from redis.

nilmerg (approved):

PR #830, opened 2023-08-09
Use ipl web widgets

yhabteab:
## Blocked by
- https://github.com/Icinga/ipl-web/pull/183


public/css/mixins.less
```diff
@@ -1,22 +1,5 @@
...
-
-.text-ellipsis() {
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
-}
-
 .monospace() {
   font-family: SFMono-Regular, Consolas, "Liberation Mono", Menlo, monospace;
 }
```
sukhwinder33445, line 5: Why is this not moved as well? It has nothing icingadb-web specific.

sukhwinder33445 (changes requested):
Please commit new changes in a new commit.

library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -99,6 +99,8 @@ protected function createTimestamp(): ?BaseHtmlElement
         } elseif ($this->state->last_state_change !== null && $this->state->last_state_change->getTimestamp() > 0) {
             return new TimeSince($this->state->last_state_change->getTimestamp());
         }
+
+        return null;
```
sukhwinder33445, line 103:
```suggestion
$since = null;
if (..) {
    $since = $val;
} elseif (..) {
    $since = $val;
}

return $since;
```

sukhwinder33445 (changes requested):

nilmerg (dismissed):

nilmerg (approved):

sukhwinder33445 (approved):

Issue #818, opened 2023-07-25
View tacticalview like the hostgroups 


gbin2265:
Hello, 

## New feature layout

I don't know if the question has already been asked, but wouldn't it be possible to also present the tacticalview as it is now done for the hostgroups.

## Describe the solution you'd like

Do the search as it is done now for the tacticalview but the presentation as the hostgroups (see photo).

![tacticallist](https://github.com/Icinga/icingadb-web/assets/29303758/270fdeb6-6aa4-4110-8ad5-56162d6f00b8)

## Reason for this question

* We currently solve by creating hostgroups and filtering on them, but this puts enormous pressure on the database due to the number of hostgroups.
  Each question the full hostgroup calculation is done in the database and if you have more than 500hostgroups it can take a long time even if we only filter one out.

* for each user and question should we create a hostgroup , reload from the icinga .....
  If we can do this like the tacticalview then there are direct links that we can place in the dashboard.
  And we can also remove all the users hostgroups so that these are again only the hostgroups that are really needed.

* I think that the question via the tactical view is much lighter to the database than the sql query of the hostgroups.
  (like filtering based on variables)

* if we can have the tactical view in list form we can put a lot more info on our screens.
  Now we have to physically install extra screens for the 24/24 operators to be able to visualize everything.

## Question

Do you think it is possible to have a tacticalview but with the look of the hostgroups !?

regards,
Geert

PR #841, opened 2023-08-16
Tactical with layout like hostgroups

gbin2265:
Hello, 

I have already asked that it is not possible to display the tactical view like the hostgroups. (https://github.com/Icinga/icingadb-web/issues/818)

I'm not a PHP programmer, but with some searching and pasting I came to this result.

tacticalline_20230816

One line /icingaweb2/icingadb/tacticalline
Based on a hostgroup /icingaweb2/icingadb/tacticalline?hostgroup.name=hostgroup_all_linux

Currently my version works except when I want to activate the search bar.
Then I have an error :

**Relation 'timeperiod' already exists**

This error I have when activaten te line  
$searchBar = $this->createSearchBar($tacticallineSummary); 
in the file TacticallineController.php

I see in the file TimeperiodOverrideExclude.php and TimeperiodOverrideInclude.php this message 
**TODO: `timeperiod` cannot be used again, find a better name**

I'll have to search why or if anyone knows why i have this error would be a big help.


gbin2265:
Hello,

I'm going to close this request because I have created other views outside the tacticalline.

views based on the service group , but with the HOST or CHECKCOMMAND as the base table.

regards,
Geert

PR #846, opened 2023-08-21
ObjectInspectionDetail: Render 'n. a.' if $command is empty

### LINKED ISSUE #845: Executed command section in source tab looks odd
## Describe the bug
When viewing the Source tab after sending a process check result, the Executed Command section looks like this:

Bildschirmfoto 2023-08-17 um 17 23 32

sukhwinder33445:
fixes #845


nilmerg (changes requested):
This should look the same as the other `n.a.` occurrences. It makes also no sense to allow copying it.

nilmerg:
I've also meant you to remove the code block look:
image

Otherwise the contrast ratio is too low.

nilmerg (approved):

PR #859, opened 2023-09-04
Allow to reset active db connection

### LINKED ISSUE #805: Provide a way to reset DB connection

Icinga DB Web reuses the database connection and results in an already killed connection being used when restarting the database (or any other temporary server error) in the meantime. See here for reference https://github.com/Icinga/icingaweb2-module-x509/issues/188.

There should be an option to reset this dead connection by the consumers.

Don't reuse database connections?

yhabteab:
resolves #805

### DIFF
--- a/library/Icingadb/ProvidedHook/X509/Sni.php
+++ b/library/Icingadb/ProvidedHook/X509/Sni.php
@@ -22,6 +22,8 @@ class Sni extends SniHook
      */
     public function getHosts(Filter $filter = null): Generator
     {
+        $this->getDb()->ping();
+
         $queryHost = Host::on($this->getDb());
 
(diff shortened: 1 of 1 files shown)


library/Icingadb/ProvidedHook/X509/Sni.php
```diff
@@ -22,6 +22,7 @@ class Sni extends SniHook
      */
     public function getHosts(Filter $filter = null): Generator
     {
+        $this->resetConn();
```
nilmerg, line 25: Please use `$this->getDb()->ping()` instead.

nilmerg (changes requested):

nilmerg (approved):

PR #867, opened 2023-09-12
action-list.js: Don't add footer for list in `.controls > .action-list`

sukhwinder33445:


nilmerg:
Is this an oversight and lists not in the content weren't navigable anyway? Or may now some lists not navigable anymore?

sukhwinder33445:
As discussed offline, only primary lists are navigable. We will stick to this behaviour for now.

library/Icingadb/Common/DetailActions.php
```diff
@@ -132,8 +132,11 @@ public function addMultiselectFilterAttribute(BaseHtmlElement $element, Filter\R
      */
     public function addDetailFilterAttribute(BaseHtmlElement $element, Filter\Rule $filter): self
     {
+        if (! $this->getDetailActionsDisabled()) {
+            $element->getAttributes()->set('data-action-item', true);
+        }
```
nilmerg, line 137: I thought you said you'll apply this the same as the other attribute? Where's the callback then?

public/js/action-list.js
```diff
@@ -719,7 +719,7 @@
             }
 
             if (isTopLevelContainer) {
-                let footerList = list ?? container.querySelector('.action-list');
+                let footerList = list ?? container.querySelector('.content .action-list');
```
nilmerg, line 722: Not `.content > .action-list`?

nilmerg (changes requested):

nilmerg (approved):

nilmerg:
@cla-bot check

nilmerg:
@cla-bot check

PR #884, opened 2023-09-27
IcingaRedis: Don't throw in method `isUnavailable`

### LINKED ISSUE #883: `Host/Service` views are not usable when Redis is unavailable
## Describe the bug

Bildschirmfoto 2023-09-26 um 16 51 38

nilmerg:
fixes #883

### DIFF
--- a/library/Icingadb/Common/IcingaRedis.php
+++ b/library/Icingadb/Common/IcingaRedis.php
@@ -44,8 +44,12 @@ class IcingaRedis
     {
         $self = self::instance();
 
-        if ($self->redis === null) {
-            $self->getConnection();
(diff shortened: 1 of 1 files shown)


yhabteab (commented):
I know this has nothing to do with this PR, but accessing the `Source` tab while Icinga 2 is stopped results in the following stack trace, you may want to fix that here as well.

Bildschirmfoto 2023-09-27 um 10 08 32

nilmerg:
Nope, that doesn't need fixing. The error message is clear enough and without Icinga 2 the view is useless anyway. Just like any other view if the DB is not reachable.

yhabteab (approved):

PR #904, opened 2023-10-09
Fix that some macros break the host detail

### LINKED ISSUE #900: Host details page not working since upgrade to 1.1.0
## Describe the bug
Since the upgrade to the new version 1.1.0 of IcingaDB Web the host overview page doesn't work anymore and only an error message is displayed.

2023-10-02 08_53_07-REZ01SR0266 - rez01sr0266 patrizia ag - Remotedesktopverbindung

However the details page of services still works and the new feature (execution start/end time) is also shown.
(I can click on the service itself in the Overview -> Services pane.)

## To Reproduce
Update to IcingaDB Web via apt update & apt upgrade.
(No other updates were available in my case, so it's really only the icingadb-web package.)

## Expected behavior
No error message should appear and the details page should work like before.

## Your Environment

* Icinga DB Web version (System - About): 1.1.0 (before 1.0.2)
* Icinga Web 2 version (System - About): 2.12.0
* Web browser: Microsoft Edge (latest)
* Icinga 2 version (`icinga2 --version`):  r2.14.0-1
* Icinga DB version (`icingadb --version`): v1.1.1
* PHP version used (`php --version`): 8.1.2-1ubuntu2.14
* Server operating system and version: Ubuntu 22.04.3 LTS

## Additional context
There are no outstanding migrations in the new System -> Migrations pane. (Neither before nor after the update.)
A system reboot doesn't help either.

nilmerg:
fixes #900

### DIFF
--- a/library/Icingadb/Common/Macros.php
+++ b/library/Icingadb/Common/Macros.php
@@ -48,7 +48,7 @@ trait Macros
      */
     public function resolveMacro(string $macro, $object): string
     {
-        if ($object instanceof Host) {
+        if ($object instanceof Host || (property_exists($object, 'type') && $object->type === 'host')) {
             $objectType = 'host';
         } else {
             $objectType = 'service';
--- a/test/php/library/Icingadb/Common/MacrosTest.php
+++ b/test/php/library/Icingadb/Common/MacrosTest.php
@@ -5,6 +5,8 @@
 namespace Tests\Icinga\Modules\Icingadb\Common;
 
 use Icinga\Module\Icingadb\Common\Macros;
+use Icinga\Module\Icingadb\Compat\CompatHost;
+use Icinga\Module\Icingadb\Compat\CompatService;
 use Icinga\Module\Icingadb\Model\Host;
 use Icinga\Module\Icingadb\Model\Service;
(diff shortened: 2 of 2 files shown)


yhabteab (dismissed):
I think these are not only some macros but every single host macros🤦.

yhabteab:
One more thing: If monitoring is not there the tests fails fatal like this (at least locally):
```bash
Error : Class "Icinga\Module\Monitoring\Object\Host" not found
 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Compat/CompatHost.php:9
 /icingaweb2/library/Icinga/Application/ClassLoader.php:275
 /usr/share/icingaweb2-modules/icingadb/test/php/library/Icingadb/Common/MacrosTest.php:44
 

Error : Class "Icinga\Module\Monitoring\Object\Service" not found
 /usr/share/icingaweb2-modules/icingadb/library/Icingadb/Compat/CompatService.php:9
 /icingaweb2/library/Icinga/Application/ClassLoader.php:275
 /usr/share/icingaweb2-modules/icingadb/test/php/library/Icingadb/Common/MacrosTest.php:110
```

yhabteab (approved):

PR #906, opened 2023-10-10
Fix that the dashboard loses selection, again

### LINKED ISSUE #696: Dashboard loses selection
## Describe the bug
When selecting an item on the dashboard it becomes highlighted but loses it when the page refreshes.

## To Reproduce
1. Go to the dashboard and select something.
2. Wait for the page to auto-refresh or press F5.

## Expected behavior
Keep the highlight like in the monitoring-module.

## Your Environment
* Icinga Web 2 version (System - About): 2.11.3
* Web browser: Firefox
* Icinga 2 version (`icinga2 --version`): 2.13.6
* Icinga DB version (`icingadb --version`): 1.1.0
* PHP version used (`php --version`): 7.3.33
* Server operating system and version: Oracle Linux 7

nilmerg:
@sukhwinder33445 I've changed the refresh suspension quite a bit, please re-test if the key navigation still behaves like you intended.

fixes #696

### DIFF
--- a/public/js/action-list.js
+++ b/public/js/action-list.js
@@ -29,8 +29,8 @@
 
             this.lastActivatedItemUrl = null;
             this.lastTimeoutId = null;
-            this.isProcessingRequest = false;
             this.isProcessingLoadMore = false;
+            this.activeRequests = {};
         }
 
         /**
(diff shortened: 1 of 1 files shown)


public/js/action-list.js
```diff
@@ -116,15 +148,12 @@
                     : activeItems[activeItems.length - 1].dataset.icingaDetailFilter;
             }
 
-            if (isBeingMultiSelected && container.id === 'col1' && list.matches('.content > :scope')) {
-                // The restriction on col1 is currently only in place as it's not defined what should happen
-                // with primary lists in col2. (as a detail load then moves the list to col1)
-                container.dataset.suspendAutorefresh = '';
+            if (! dashboard) {
+                _this.addSelectionCountToFooter(list);
```
sukhwinder33445, line 152: This must be called after `clearSelection()` and `setActive()`. Otherwise the counter in the footer is not up to date.

sukhwinder33445 (changes requested):
There is another small bug that the first click on comment/downtime in the detail view loads nothing and displays empty #col2:
1. Select a host/service from the list that has at least one comment/downtime.
2. Click on the comment/downtime in the detail view.
3. There is an error in the console.

sukhwinder33445 (approved):

PR #968, opened 2024-02-02
action-list.js: Update the history explicitly when closing col2

nilmerg:
With https://github.com/Icinga/icingaweb2/pull/5179 this will not be done by default anymore


nilmerg:
Merged with 2e2b0534b68d0639a89f68f6ddff23d7f8ec0853

PR #992, opened 2024-03-21
Fix a mysterious `MigrateCommand` issue

nilmerg:


nilmerg:
@raviks789 I remember that we found an issue which we didn't fix at the time this got introduced. I found this test in my stash. Is this nonsense or do you see what this is supposed to test?

raviks789:
> @raviks789 I remember that we found an issue which we didn't fix at the time this got introduced. I found this test in my stash. Is this nonsense or do you see what this is supposed to test?

If I remember correctly, the case was if a navigation item already exists in `icingadb-host-action` but is not a shared item and the same navigation item is present in `host-action` as shared item, then only the filter needs to be transformed, instead of creating a new shared navigation item for `icingadb-host-action`. If I am understanding correctly the test is written for this case.

So I think this is a valid test case.

PR #1019, opened 2024-05-07
Add configuration settings for maximal number of characters in plugin…

### LINKED ISSUE #1017: Plugin Output limits limit too much
## Describe the bug
The changes introduced in #998 limit the Plugin Output to 10.000 symbols (chars?) by default,
which, in some edge cases is not enough ( ref NC/817322 ).

## Possible solution
Add a configuration parameter (config.ini) to set the limit manually for a setup.

RincewindsHat:
… output

closes #1017


RincewindsHat:
Added some changes, the character limit from the configuration should now be applied to every new instance of `PluginOutput`.

PR #1026, opened 2024-06-17
PluginOutput::render():  Return correctly cut off output

### LINKED ISSUE #1012: filtered views behave random after update
## Describe the bug

After updating from icingadb-web 1.1.1 to 1.1.2 filtered list views randomly shows less objects than expected. 

image002

The image shows the problem, the list should show 37 Warnings and only shows 1 regarding the page limit of 50. 

image003

Secondly if the items exceed the page limit, every page shows a random number of items. 

## Your Environment

* Icinga DB Web version (System - About): 1.1.2
* Icinga Web 2 version (System - About): 2.12.1
* Web browser: Chrome/Edge/Firefox
* Icinga 2 version (`icinga2 --version`): 2.14.2
* Icinga DB version (`icingadb --version`): 1.2.0
* PHP version used (`php --version`): 7.4.33
* Server operating system and version: Rhel 8

sukhwinder33445:
fixes #1012

### DIFF
--- a/library/Icingadb/Util/PluginOutput.php
+++ b/library/Icingadb/Util/PluginOutput.php
@@ -16,7 +16,7 @@ use LogicException;
 class PluginOutput extends HtmlString
 {
     /** @var string[] Patterns to be replaced in plain text plugin output */
-    const TEXT_PATTERNS = [
+    protected const TEXT_PATTERNS = [
         '~\\\t~',
         '~\\\n~',
         '~(\[|\()OK(\]|\))~',
@@ -29,7 +29,7 @@ class PluginOutput extends HtmlString
     ];
 
     /** @var string[] Replacements for {@see PluginOutput::TEXT_PATTERNS} */
-    const TEXT_REPLACEMENTS = [
+    protected const TEXT_REPLACEMENTS = [
         "\t",
         "\n",
         '<span class="state-ball ball-size-m state-ok"></span>',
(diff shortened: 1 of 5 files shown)


tbauriedel:
ref/NC/818175

tbauriedel:
There is already a PR to change the PluginOutput #1019

sukhwinder33445:
As discussed offline, linked PR above is an extension and independent.

library/Icingadb/Util/PluginOutput.php
nilmerg, line 1: This still has numerous problems:

1. `PluginOutput::$enrichOutput` was previously a prerequisite for the replacement of state macros. Now it's ignored.
2. In case of plain text, `htmlspecialchars` was utilized. Now it's not.
3. Since replacement happens now before the check for HTML, once a macro is matched in plain text, it's now magically HTML and being purified for no reason. Or is 2) the reason? But then, why only if a macro has been replaced??

test/php/library/Icingadb/Util/PluginOutputTest.php
nilmerg, line 1: monitoring also has a test, please migrate this over and see if it succeeds here afterwards

nilmerg (changes requested):

library/Icingadb/Util/PluginOutput.php
nilmerg, line 1: While you already added a return type to method `render`, you can now also add visibility to constants.

nilmerg (changes requested):
Update phpstan's baseline completely please, not just partially.

nilmerg (approved):

PR #1032, opened 2024-07-19
JsonResultSetUtils: When query offset is null, Use 0 as default

sukhwinder33445:
fixes https://github.com/Icinga/icingadb-web/issues/1014

### DIFF
--- a/library/Icingadb/Data/JsonResultSetUtils.php
+++ b/library/Icingadb/Data/JsonResultSetUtils.php
@@ -70,7 +70,7 @@ trait JsonResultSetUtils
         if ($query->hasLimit()) {
             // Custom limits should still apply
             $query->peekAhead(false);
-            $offset = $query->getOffset();
+            $offset = $query->getOffset() ?? 0;
         } else {
             $query->limit(1000);
             $query->peekAhead();


library/Icingadb/Data/JsonResultSetUtils.php
```diff
@@ -70,7 +70,7 @@ public static function stream(Query $query): void
         if ($query->hasLimit()) {
             // Custom limits should still apply
             $query->peekAhead(false);
-            $offset = $query->getOffset();
+            $offset = $query->getOffset() ?: 0;
```
nilmerg, line 73: Why not `??`?

sukhwinder33445: In this case, I see no advantage/disadvantage, as the method always returns a value. `(null|int`).

nilmerg: I do. `0 ?? 0` only checks the type whereas `0 ?: 0` also the value. And since the method only returns a nullable int and a zero is fine, the latter is less efficient. But it's not about efficiency, but about clarity.

The next time you write something like this, and your fallback value doesn't match the default, this will bite you: `$value ?: 100`

In short: If a method may return null and you are fine with anything else, use `??`.

sukhwinder33445: I see, thanks for the explanation.

nilmerg (approved):

PR #1033, opened 2024-07-22
Configurable output limit

### LINKED ISSUE #1017: Plugin Output limits limit too much

nilmerg:
resolves #1017
closes #1019

### DIFF
--- a/doc/03-Configuration.md
+++ b/doc/03-Configuration.md
@@ -69,26 +69,25 @@ configuration file `/etc/icingaweb2/modules/icingadb/config.ini`.
 
 ### Available Settings and defaults
 
-Option                            | Description                      | Default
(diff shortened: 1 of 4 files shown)


sukhwinder33445 (changes requested):
The `setCharacterLimit` call is missing in `PluginOutput::fromObject()`, which causes the host/service lists to still render the full output.

nilmerg:
There is a default of `1000` which should be in effect in lists.

sukhwinder33445 (approved):

PR #1039, opened 2024-07-31
Fix unstable service order

nilmerg:
Services usually have similar (if not equal) names across different hosts. Sorting such after (display_)name, might result in unexpected results depending on the used database backend.

To guarantee a deterministic order we have to use more than a single column if said column isn't variant enough.


nilmerg:
Currently not tested on PostgreSQL

nilmerg:
We'll need to verify query plans of several backends and versions in order to make sure this has no disadvantages.

PR #1046, opened 2024-08-20
Add hook to include custom variables by default when searching under Overview entries

### LINKED ISSUE #1045: Include custom variables automatically when searching on Overview entries
I'd like to include some custom variables by default when searching in Overview sections.

For example, when I search Hosts by default the included columns are `name_ci` and `display_name`. I'd like to have for example `host.vars.hostfoo` and `host.vars.hostbar` that are two custom variables I created for the hosts.

Example:
image

oskarnrk:
The hook can be implemented in multiple ways. For example:

Add to **run.php**
```php
$this->provideHook('Icingadb/SearchColumnsProvider');
```

**SearchColumnsProvider.php**
```php
<?php

namespace Icinga\Module\Icingadb\ProvidedHook;

use Icinga\Module\Icingadb\Hook\SearchColumnsProviderHook;
use ipl\Orm\Model;

class SearchColumnsProvider extends SearchColumnsProviderHook
{
    public function retrieveCustomVars(Model $model): array
    {
        switch ($model->getTableName()) {
            case 'host':
                return ['host.vars.hostfoo', 'host.vars.hostbar'];
            case 'service':
                return ['service.vars.servicefoo', 'service.vars.servicebar'];
            default:
                return [];
        }
    }
}
```


oskarnrk:
I signed the ICLA ;)

oskarnrk:
> Thank you for your pull request. Before we can look at it, you'll need to sign a Contributor License Agreement (CLA).
> 
> **Please follow instructions at https://icinga.com/company/contributor-agreement to sign the CLA.**
> 
> After that, please reply here with a comment and we'll verify.
> 
> Contributors that have not signed yet: @oskarnrk

I signed the ICLA, thanks in advance.

oskarnrk:
Is it possibile to retrigger the CLA check? Thanks

bobapple:
@cla-bot check

oskarnrk:
Please, is it possible to check this quite old PR? Thanks 😇

nilmerg:
Hi, the reason why (at least for me) no-one looked at this yet, is that this isn't just a small feature. This has so many other parts which might be affected, that I don't want to introduce a half baked solution. (Besides that a new hook also requires us to cover support for it, which I cannot guarantee at all)

library/Icingadb/Hook/CustomVarsRetrieverHook.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
```
nilmerg: Please add a license header.

```suggestion

// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
// SPDX-License-Identifier: GPL-3.0-or-later

```

library/Icingadb/Hook/CustomVarsRetrieverHook.php
```diff
@@ -0,0 +1,37 @@
...
+namespace Icinga\Module\Icingadb\Hook;
+
+use Icinga\Application\Hook;
+use Icinga\Application\Logger;
+use ipl\Orm\Model;
+use Throwable;
+
+abstract class CustomVarsRetrieverHook
```
nilmerg: Please rename this to `SearchColumnsProvider`.

Also add a class doc block containing a deprecation notice:

We have to make sure everyone knows this is an unfinished implementation. (Please don't accept the suggestion as-is, I didn't make sure the line length is fine…)

```suggestion
/**
  * This is a first attempt to generalize support for modules to provide additional search columns, such as custom variables. The end user won't be able to customize the set provided and every global or route search will always use them. Any hook implementation will also need to ensure that only columns are returned, especially custom variables, the user has access to and that are valid in the current route. ([As otherwise the user sees an error](https://github.com/Icinga/icingadb-web/security/advisories/GHSA-w57j-28jc-8429)) A future version will not re-use this hook and integrations need to be adapted manually.

  * @deprecated This will not be used by any future version without any further notice.
  */
abstract class SearchColumnsProviderHook
```

library/Icingadb/Hook/CustomVarsRetrieverHook.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    abstract public function retrieveCustomVars(Model $model): array;
+
+    final public static function getCustomVarColumns(Model $model): array
```
nilmerg: It makes sense to pass the current default set. Then, de-duplication can happen as we have to ensure the view won't register the same columns over and over just because defaults are provided again or multiple implementations provide the same ones.

So pass over the current default columns where this method is called and before it returns the result apply `array_unique` to it.

nilmerg (changes requested):
Since https://github.com/Icinga/ipl-web/commit/91221e206555150851991552e7747fc5d3f9f8fd there's an additional usage of `Model::getSearchColumns()`. It's intended to provide the same set of columns the searchbar will provide as a default in the same route, but applied in case the user hits enter faster than the default suggestion appears (or rather can be interacted with).

My suggestion to extend this, isolated to Icinga DB Web, is to change `\Icinga\Module\Icingadb\Common\SearchControls::createSearchBar` to override what the base implementation did. However, for this to work at all, `\ipl\Web\Compat\SearchControls::callHandleRequest` has be overridden and you have to call `$searchBar->handleRequest($this->getServerRequest())`.

library/Icingadb/Web/Controller.php
```diff
@@ -303,7 +304,10 @@ public function handleSearchRequest(Query $query, array $additionalColumns = [])
      */
     protected function prepareSearchFilter(Query $query, string $search, Filter\Any $filter, array $additionalColumns)
     {
-        $columns = array_merge($query->getModel()->getSearchColumns(), $additionalColumns);
+        $model = $query->getModel();
+        $defaultColumns = $model->getSearchColumns();
+        $customVarColumns = SearchColumnProviderHook::getCustomVarColumns($model, $defaultColumns);
+        $columns = array_merge($additionalColumns, $customVarColumns);
```
nilmerg: this merge is rather obsolete, as the *additional* columns should already be part of the default set.

nilmerg (changes requested):
looking fine now, but please rebase and fix the conflicts

nilmerg (approved):
Looks fine, thanks :)

PR #1055, opened 2024-09-16
db: Cleanup parent/child relations in matching models

nilmerg:
The previous relations didn't work as expected.
Now, filtering with `child.host.name` or
`parent.service.name` works fine.

### DIFF
--- a/library/Icingadb/Model/DependencyEdge.php
+++ b/library/Icingadb/Model/DependencyEdge.php
@@ -17,8 +17,8 @@ use ipl\Orm\Relations;
  * @property string $from_node_id
  * @property ?string $dependency_id
  *
- * @property DependencyNode|Query $from
- * @property DependencyNode|Query $to
+ * @property DependencyNode|Query $child
+ * @property DependencyNode|Query $parent
  * @property (?Dependency)|Query $dependency
  */
 class DependencyEdge extends Model
@@ -53,11 +53,17 @@ class DependencyEdge extends Model
 
     public function createRelations(Relations $relations): void
     {
+        $relations->belongsTo('child', DependencyNode::class)
+            ->setCandidateKey('from_node_id');
+        $relations->belongsTo('parent', DependencyNode::class)
+            ->setCandidateKey('to_node_id');
+        $relations->belongsTo('dependency', Dependency::class)
+            ->setJoinType('LEFT');
+
+        // "from" and "to" are only necessary for sub-query filters.
         $relations->belongsTo('from', DependencyNode::class)
             ->setCandidateKey('from_node_id');
         $relations->belongsTo('to', DependencyNode::class)
             ->setCandidateKey('to_node_id');
-        $relations->belongsTo('dependency', Dependency::class)
-            ->setJoinType('LEFT');
     }
 }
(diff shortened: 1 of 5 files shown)


sukhwinder33445:
# Test:
I have used the example configuration but reduced the number of objects (loop range).

Tested this PR with the filter `icingadb/hosts?parent.host.name=big-switch-server-0`.
### Expected behavior: 
No result, as there is no child existing with this host as parent.
### Actual behavior: 
One host `big-switch-server-0` (self) as result. //  There are services that have this host as a parent

## The sql query:
(icingadb/hosts?parent.host.name=big-switch-server-0&format=sql)

``` sql
...
WHERE host.id IN ((SELECT sub_host_to_from_host.id AS sub_host_to_from_host_id
                   FROM host sub_host
                            INNER JOIN dependency_node sub_host_dependency_node # here we have 1:n (host : dependency_node) relation
                                       ON sub_host_dependency_node.host_id = sub_host.id
                            INNER JOIN dependency_edge sub_host_to
                                       ON sub_host_to.to_node_id = sub_host_dependency_node.id
                            INNER JOIN dependency_node sub_host_to_from
                                       ON sub_host_to_from.id = sub_host_to.from_node_id
                            LEFT JOIN host sub_host_to_from_host ON sub_host_to_from_host.id = sub_host_to_from.host_id
                   WHERE sub_host.name = 'big-switch-server-0'))
ORDER BY host.display_name
LIMIT 25;
```

The column `dependency_node.host_id` is not unique and there are several identical entries in case services exist for this host. This leads to a ‘1:n’ relationship, which leads to false positive results.
Adding a service filter in the case of a host solves this problem.

sukhwinder33445 (commented):
## Test:
I have used the example configuration but reduced the number of objects (loop range).

Tested this PR with the filter `icingadb/hosts?parent.host.name=big-switch-server-0`.
### Expected behavior: 
No result, as there is no child existing with this host as parent.
### Actual behavior: 
One host `big-switch-server-0` (self) as result. //  There are services that have this host as a parent

## The sql query:
(icingadb/hosts?parent.host.name=big-switch-server-0&format=sql)

``` sql
...
WHERE host.id IN ((SELECT sub_host_to_from_host.id AS sub_host_to_from_host_id
                   FROM host sub_host
                            INNER JOIN dependency_node sub_host_dependency_node # here we have 1:n (host : dependency_node) relation
                                       ON sub_host_dependency_node.host_id = sub_host.id
                            INNER JOIN dependency_edge sub_host_to
                                       ON sub_host_to.to_node_id = sub_host_dependency_node.id
                            INNER JOIN dependency_node sub_host_to_from
                                       ON sub_host_to_from.id = sub_host_to.from_node_id
                            LEFT JOIN host sub_host_to_from_host ON sub_host_to_from_host.id = sub_host_to_from.host_id
                   WHERE sub_host.name = 'big-switch-server-0'))
ORDER BY host.display_name
LIMIT 25;
```

The column `dependency_node.host_id` is not unique and there are several identical entries in case services exist for this host. This leads to a ‘1:n’ relationship, which leads to false positive results.
Adding a service filter in the case of a host solves this problem.

sukhwinder33445 (approved):

Issue #1050, opened 2024-09-03
Show a list of root problems in the host/service detail if it has a problem and is part of a dependency


nilmerg:
In order to aid users to resolve an incident we want to list all parents, that may be the cause of the problem, in its detail area.

**It should look like this:**
<img width=800 src="https://github.com/user-attachments/assets/3714a2b6-0ad4-442f-a93d-4a2ef7a35b98">

## Notes

* The list may contain objects of type *host*, *service* and (new) *redundancy group*
  * Members of a redundancy group must not be shown, it's the group's detail area which is responsible for this
* Use `\Icinga\Module\Icingadb\Model\UnreachableParent` to fetch the content
  * Restrictions must be applied as usual
* Each item links to its detail area
  * Redundancy groups have none yet and will get one separately
  * For now, they won't have a link associated with them
* The mockup above includes a link to *all 123* items. This is not required and will also be solved separately
  * This means, all items must be shown
* The list should be entirely replaceable by hooks of type `ObjectDetailExtensionHook`
  * Currently the plugin output is at the top, where it should still be in case there are no root problems
    * It is using position `0` in all details (e.g. in [`HostDetail`](https://github.com/Icinga/icingadb-web/blob/334aaf848eeba3ca7efcf11a512377e540e9fa46/library/Icingadb/Widget/Detail/HostDetail.php#L44)). The root problems should use this position instead, and the plugin output be moved to `1`

PR #1057, opened 2024-09-17
Show root problem list for objects with problem and are part of dependency

raviks789:
ref #1050

blocked by: #1055
blocked by: Icinga/ipl-web#231,  Icinga/ipl-web#234

### DIFF
--- a/application/controllers/ServiceController.php
+++ b/application/controllers/ServiceController.php
@@ -38,13 +38,15 @@ class ServiceController extends Controller
         $name = $this->params->getRequired('name');
         $hostName = $this->params->getRequired('host.name');
 
-        $query = Service::on($this->getDb())->with([
-            'state',
-            'icon_image',
-            'host',
-            'host.state',
-            'timeperiod'
-        ]);
+        $query = Service::on($this->getDb())
+            ->withColumns(['has_problematic_parent'])
+            ->with([
+                'state',
+                'icon_image',
+                'host',
+                'host.state',
+                'timeperiod'
+            ]);
         $query
             ->setResultSetClass(VolatileStateResults::class)
             ->filter(Filter::all(
--- a/library/Icingadb/Common/StateBadges.php
+++ b/library/Icingadb/Common/StateBadges.php
@@ -4,11 +4,12 @@
 
 namespace Icinga\Module\Icingadb\Common;
 
+use InvalidArgumentException;
 use ipl\Html\BaseHtmlElement;
 use ipl\Html\Html;
+use ipl\Html\HtmlElement;
 use ipl\Stdlib\BaseFilter;
 use ipl\Stdlib\Filter;
-use ipl\Web\Filter\QueryString;
 use ipl\Web\Url;
 use ipl\Web\Widget\Link;
 use ipl\Web\Widget\StateBadge;
@@ -26,7 +27,7 @@ abstract class StateBadges extends BaseHtmlElement
     /** @var string Prefix */
     protected $prefix;
 
-    /** @var Url Badge link */
+    /** @var ?Url Badge link */
     protected $url;
 
     protected $tag = 'ul';
@@ -47,13 +48,6 @@ abstract class StateBadges extends BaseHtmlElement
     }
 
     /**
-     * Get the badge base URL
-     *
-     * @return Url
-     */
-    abstract protected function getBaseUrl(): Url;
-
-    /**
      * Get the type of the items
      *
      * @return string
@@ -68,20 +62,35 @@ abstract class StateBadges extends BaseHtmlElement
     abstract protected function getPrefix(): string;
 
     /**
+     * Get the badge base URL
+     *
+     * @return ?Url
+     */
+    protected function getBaseUrl(): ?Url
+    {
+        return null;
+    }
+
+    /**
      * Get the integer of the given state text
      *
      * @param string $state
      *
      * @return int
+     *
+     * @throws InvalidArgumentException if the given state is not valid
      */
-    abstract protected function getStateInt(string $state): int;
+    protected function getStateInt(string $state): int
+    {
+        throw new InvalidArgumentException(sprintf('%s is not a valid state', $state));
+    }
 
     /**
      * Get the badge URL
      *
-     * @return Url
+     * @return ?Url
      */
-    public function getUrl(): Url
+    public function getUrl(): ?Url
     {
         return $this->url;
     }
@@ -108,7 +117,7 @@ abstract class StateBadges extends BaseHtmlElement
      *
      * @return Link
      */
-    public function createLink($content, Filter\Rule $filter = null): Link
+    protected function createLink($content, Filter\Rule $filter = null): Link
     {
         $url = clone $this->getUrl();
 
@@ -135,18 +144,23 @@ abstract class StateBadges extends BaseHtmlElement
      *
      * @return ?BaseHtmlElement
      */
-    protected function createBadge(string $state)
+    protected function createBadge(string $state): ?BaseHtmlElement
     {
         $key = $this->prefix . "_{$state}";
-
-        if (isset($this->item->$key) && $this->item->$key) {
-            return Html::tag('li', $this->createLink(
-                new StateBadge($this->item->$key, $state),
-                Filter::equal($this->type . '.state.soft_state', $this->getStateInt($state))
-            ));
+        if (empty($this->item->$key)) {
+            return null;
         }
 
-        return null;
+        $stateBadge = new StateBadge($this->item->$key, $state);
+
+        if ($this->url !== null) {
+            $this->createLink(
+                $stateBadge,
+                Filter::equal($this->type . '.state.soft_state', $this->getStateInt($state))
+            );
+        }
+
+        return new HtmlElement('li', null, $stateBadge);
     }
 
     /**
@@ -156,34 +170,46 @@ abstract class StateBadges extends BaseHtmlElement
      *
      * @return ?BaseHtmlElement
      */
-    protected function createGroup(string $state)
+    protected function createGroup(string $state): ?BaseHtmlElement
     {
         $content = [];
         $handledKey = $this->prefix . "_{$state}_handled";
         $unhandledKey = $this->prefix . "_{$state}_unhandled";
 
         if (isset($this->item->$unhandledKey) && $this->item->$unhandledKey) {
-            $content[] = Html::tag('li', $this->createLink(
-                new StateBadge($this->item->$unhandledKey, $state),
-                Filter::all(
-                    Filter::equal($this->type . '.state.soft_state', $this->getStateInt($state)),
-                    Filter::equal($this->type . '.state.is_handled', 'n'),
-                    Filter::equal($this->type . '.state.is_reachable', 'y')
-                )
-            ));
+            $unhandledStateBadge = new StateBadge($this->item->$unhandledKey, $state);
+
+            if ($this->url !== null) {
+                $unhandledStateBadge = $this->createLink(
+                    $unhandledStateBadge,
+                    Filter::all(
+                        Filter::equal($this->type . '.state.soft_state', $this->getStateInt($state)),
+                        Filter::equal($this->type . '.state.is_handled', 'n'),
+                        Filter::equal($this->type . '.state.is_reachable', 'y')
+                    )
+                );
+            }
+
+            $content[] = new HtmlElement('li', null, $unhandledStateBadge);
         }
 
         if (isset($this->item->$handledKey) && $this->item->$handledKey) {
-            $content[] = Html::tag('li', $this->createLink(
-                new StateBadge($this->item->$handledKey, $state, true),
-                Filter::all(
-                    Filter::equal($this->type . '.state.soft_state', $this->getStateInt($state)),
-                    Filter::any(
-                        Filter::equal($this->type . '.state.is_handled', 'y'),
-                        Filter::equal($this->type . '.state.is_reachable', 'n')
+            $handledStateBadge = new StateBadge($this->item->$handledKey, $state, true);
+
+            if ($this->url !== null) {
+                $handledStateBadge = $this->createLink(
+                    $handledStateBadge,
+                    Filter::all(
+                        Filter::equal($this->type . '.state.soft_state', $this->getStateInt($state)),
+                        Filter::any(
+                            Filter::equal($this->type . '.state.is_handled', 'y'),
+                            Filter::equal($this->type . '.state.is_reachable', 'n')
+                        )
                     )
-                )
-            ));
+                );
+            }
+
+            $content[] = new HtmlElement('li', null, $handledStateBadge);
         }
 
         if (empty($content)) {
--- a/library/Icingadb/Model/Behavior/HasProblematicParent.php
+++ b/library/Icingadb/Model/Behavior/HasProblematicParent.php
@@ -1,0 +1,94 @@
+<?php
+
+/* Icinga DB Web | (c) 2024 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Icingadb\Model\Behavior;
+
+use Icinga\Module\Icingadb\Model\DependencyEdge;
+use ipl\Orm\AliasedExpression;
+use ipl\Orm\ColumnDefinition;
+use ipl\Orm\Exception\InvalidColumnException;
+use ipl\Orm\Query;
+use ipl\Sql\Expression;
+use ipl\Stdlib\Filter;
+use ipl\Orm\Contract\RewriteColumnBehavior;
+use ipl\Orm\Contract\QueryAwareBehavior;
+
+/**
+ * Behavior to check if the service has a problematic parent
+ */
+class HasProblematicParent implements RewriteColumnBehavior, QueryAwareBehavior
+{
+    /** @var Query */
+    protected $query;
+
+    public function setQuery(Query $query): self
+    {
+        $this->query = $query;
+
+        return $this;
+    }
+
+    public function rewriteColumn($column, ?string $relation = null): ?AliasedExpression
+    {
+        if (! $this->isSelectableColumn($column)) {
+            return null;
+        }
+
+        $resolver = $this->query->getResolver();
+        if ($relation !== null) {
+            $serviceTableAlias = $resolver->getAlias($resolver->resolveRelation($relation)->getTarget());
+            $column = $resolver->qualifyColumnAlias($column, $serviceTableAlias);
+        } else {
+            $serviceTableAlias = $resolver->getAlias($this->query->getModel());
+        }
+
+        $subQueryModel = new DependencyEdge();
+        $subQuery = (new Query())
+            ->setDb($this->query->getDb())
+            ->setModel($subQueryModel)
+            ->columns([new Expression('1')])
+            ->utilize('from')
+            ->limit(1)
+            ->filter(Filter::equal('dependency.state.failed', 'y'));
+
+        $subQueryResolver = $subQuery->getResolver()->setAliasPrefix('hpp_');
+        $subQueryTarget = $subQueryResolver->resolveRelation($subQueryModel->getTableName() . '.from')->getTarget();
+        $targetForeignKey = $subQueryResolver->qualifyColumn(
+            'service_id',
+            $subQueryResolver->getAlias($subQueryTarget)
+        );
+
+        $subQuery->getSelectBase()
+            ->where("$targetForeignKey = {$resolver->qualifyColumn('id', $serviceTableAlias)}");
+
+        [$select, $values] = $this->query->getDb()
+            ->getQueryBuilder()
+            ->assembleSelect($subQuery->assembleSelect());
+
+        return new AliasedExpression(
+            $this->query->getDb()->quoteIdentifier([$column]),
+            "($select)",
+            null,
+            ...$values
+        );
+    }
+
+    public function isSelectableColumn(string $name): bool
+    {
+        return $name === 'has_problematic_parent';
+    }
+
+    public function rewriteColumnDefinition(ColumnDefinition $def, string $relation): void
+    {
+    }
+
+    public function rewriteCondition(Filter\Condition $condition, $relation = null)
+    {
+        $column = substr($condition->getColumn(), strlen($relation ?? ''));
+
+        if ($this->isSelectableColumn($column)) {
+            throw new InvalidColumnException($column, $this->query->getModel());
+        }
+    }
+}
--- a/library/Icingadb/Model/RedundancyGroupSummary.php
+++ b/library/Icingadb/Model/RedundancyGroupSummary.php
@@ -1,0 +1,159 @@
+<?php
+
+/* Icinga DB Web | (c) 2024 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Icingadb\Model;
+
+use ipl\Orm\Query;
+use ipl\Sql\Connection;
+use ipl\Sql\Expression;
+use ipl\Sql\Select;
+
+/**
+ * Redundancy group's summary
+ *
+ * @property int $nodes_total
+ * @property int $nodes_ok
+ * @property int $nodes_problem_handled
+ * @property int $nodes_problem_unhandled
+ * @property int $nodes_pending
+ * @property int $nodes_unknown_handled
+ * @property int $nodes_unknown_unhandled
+ * @property int $nodes_warning_handled
+ * @property int $nodes_warning_unhandled
+ */
+class RedundancyGroupSummary extends RedundancyGroup
+{
+    public function getSummaryColumns(): array
+    {
+        return [
+            'nodes_total' => new Expression('COUNT(*)'),
+            'nodes_ok' => new Expression(
+                'SUM(CASE'
+                . ' WHEN %s IS NOT NULL THEN (CASE WHEN %s = 0 THEN 1 ELSE 0 END)'
+                . ' WHEN %s = 0 THEN 1'
+                . ' ELSE 0'
+                . ' END)',
+                [
+                    'from.to.service_id',
+                    'from.to.service.state.soft_state',
+                    'from.to.host.state.soft_state',
+                ]
+            ),
+            'nodes_problem_handled' => new Expression(
+                'SUM(CASE'
+                . " WHEN %s IS NOT NULL THEN (CASE WHEN %s = 2 AND (%s = 'y' OR %s = 'n') THEN 1 ELSE 0 END)"
+                . " WHEN %s = 1 AND (%s = 'y' OR %s = 'n') THEN 1"
+                . ' ELSE 0'
+                . ' END)',
+                [
+                    'from.to.service_id',
+                    'from.to.service.state.soft_state',
+                    'from.to.service.state.is_handled',
+                    'from.to.service.state.is_reachable',
+                    'from.to.host.state.soft_state',
+                    'from.to.host.state.is_handled',
+                    'from.to.host.state.is_reachable',
+                ]
+            ),
+            'nodes_problem_unhandled' => new Expression(
+                'SUM(CASE'
+                . " WHEN %s IS NOT NULL THEN (CASE WHEN %s = 2 AND (%s = 'n' AND %s = 'y') THEN 1 ELSE 0 END)"
+                . " WHEN %s = 1 AND (%s = 'n' AND %s = 'y') THEN 1"
+                . ' ELSE 0'
+                . ' END)',
+                [
+                    'from.to.service_id',
+                    'from.to.service.state.soft_state',
+                    'from.to.service.state.is_handled',
+                    'from.to.service.state.is_reachable',
+                    'from.to.host.state.soft_state',
+                    'from.to.host.state.is_handled',
+                    'from.to.host.state.is_reachable',
+                ]
+            ),
+            'nodes_pending' => new Expression(
+                'SUM(CASE'
+                . ' WHEN %s IS NOT NULL THEN (CASE WHEN %s = 99 THEN 1 ELSE 0 END)'
+                . ' WHEN %s = 99 THEN 1'
+                . ' ELSE 0'
+                . ' END)',
+                [
+                    'from.to.service_id',
+                    'from.to.service.state.soft_state',
+                    'from.to.host.state.soft_state',
+                ]
+            ),
+            'nodes_unknown_handled' => new Expression(
+                'SUM(CASE'
+                . " WHEN %s IS NOT NULL THEN (CASE WHEN %s = 3 AND (%s = 'y' OR %s = 'n') THEN 1 ELSE 0 END)"
+                . ' ELSE 0'
+                . ' END)',
+                [
+                    'from.to.service_id',
+                    'from.to.service.state.soft_state',
+                    'from.to.service.state.is_handled',
+                    'from.to.service.state.is_reachable'
+                ]
+            ),
+            'nodes_unknown_unhandled' => new Expression(
+                'SUM(CASE'
+                . " WHEN %s IS NOT NULL THEN (CASE WHEN %s = 3 AND (%s = 'n' AND %s = 'y') THEN 1 ELSE 0 END)"
+                . ' ELSE 0'
+                . ' END)',
+                [
+                    'from.to.service_id',
+                    'from.to.service.state.soft_state',
+                    'from.to.service.state.is_handled',
+                    'from.to.service.state.is_reachable'
+                ]
+            ),
+            'nodes_warning_handled' => new Expression(
+                'SUM(CASE'
+                . " WHEN %s IS NOT NULL THEN (CASE WHEN %s = 1 AND (%s = 'y' OR %s = 'n') THEN 1 ELSE 0 END)"
+                . ' ELSE 0'
+                . ' END)',
+                [
+                    'from.to.service_id',
+                    'from.to.service.state.soft_state',
+                    'from.to.service.state.is_handled',
+                    'from.to.service.state.is_reachable'
+                ]
+            ),
+            'nodes_warning_unhandled' => new Expression(
+                'SUM(CASE'
+                . " WHEN %s IS NOT NULL THEN (CASE WHEN %s = 1 AND (%s = 'n' AND %s = 'y') THEN 1 ELSE 0 END)"
+                . ' ELSE 0'
+                . ' END)',
+                [
+                    'from.to.service_id',
+                    'from.to.service.state.soft_state',
+                    'from.to.service.state.is_handled',
+                    'from.to.service.state.is_reachable'
+                ]
+            )
+        ];
+    }
+
+    public static function on(Connection $db): Query
+    {
+        $q = parent::on($db);
+
+        /** @var static $m */
+        $m = $q->getModel();
+        $q->columns($m->getSummaryColumns());
+
+        $q->on($q::ON_SELECT_ASSEMBLED, function (Select $select) use ($q) {
+            $model = $q->getModel();
+            $groupBy = $q->getResolver()->qualifyColumnsAndAliases((array) $model->getKeyName(), $model, false);
+            $select->groupBy($groupBy);
+        });
+
+        return $q;
+    }
+
+    public function getColumns(): array
+    {
+        return array_merge(parent::getColumns(), $this->getSummaryColumns());
+    }
+}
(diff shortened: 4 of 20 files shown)


library/Icingadb/Widget/RedundancyGroupStatistics.php
```diff
@@ -0,0 +1,52 @@
...
+namespace Icinga\Module\Icingadb\Widget;
+
+use Icinga\Chart\Donut;
+
+use Icinga\Module\Icingadb\Widget\Detail\ObjectStatistics;
+use ipl\Html\Text;
+use ipl\Html\ValidHtml;
+use ipl\Html\HtmlString;
+
+class RedundancyGroupStatistics extends ObjectStatistics
```
nilmerg, line 14: Please rename this to *Object**s**Statistics*. There is not much which makes this redundancy group specific.

It's just not specific to hosts and services anymore. We will benefit from a generalized implementation somewhere else ;)

library/Icingadb/Widget/RedundancyGroupStatistics.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $donut = (new Donut())
+            ->addSlice($this->summary->parents_ok, ['class' => 'slice-state-ok'])
+            ->addSlice($this->summary->parents_warning_handled, ['class' => 'slice-state-warning-handled'])
+            ->addSlice($this->summary->parents_warning_unhandled, ['class' => 'slice-state-warning'])
+            ->addSlice($this->summary->parents_problem_handled, ['class' => 'slice-state-critical-handled'])
+            ->addSlice($this->summary->parents_problem_unhandled, ['class' => 'slice-state-critical'])
+            ->addSlice($this->summary->parents_unknown_handled, ['class' => 'slice-state-unknown-handled'])
+            ->addSlice($this->summary->parents_unknown_unhandled, ['class' => 'slice-state-unknown'])
+            ->addSlice($this->summary->parents_pending, ['class' => 'slice-state-pending']);
```
nilmerg, line 33: Replace `parents_` with `objects_`

library/Icingadb/Widget/RedundancyGroupParentStateBadges.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+/* Icinga DB Web | (c) 2024 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Icingadb\Widget;
+
+use Icinga\Module\Icingadb\Common\StateBadges;
+
+class RedundancyGroupParentStateBadges extends StateBadges
```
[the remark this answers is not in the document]
nilmerg: Since I just requested that *RedundancyGroupStatistics* should be renamed to *ObjectsStatistics*, these should be *ObjectsStateBadges*.

library/Icingadb/Widget/RedundancyGroupParentStateBadges.php
```diff
@@ -0,0 +1,33 @@
...
+{
+    protected function getType(): string
+    {
+        return 'redundancy_group_parent';
+    }
+
+    protected function getPrefix(): string
+    {
+        return 'parents';
```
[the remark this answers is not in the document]
nilmerg: Well, objects :sweat_smile:

library/Icingadb/Widget/ItemList/RedundancyGroupListItem.php
```diff
@@ -0,0 +1,153 @@
...
+            ->filter($filter)
+            ->orderBy([
+                'parent_severity',
+                'parent_last_state_change',
+            ], 'DESC');
+
+        $this->applyRestrictions($members);
+
+        /** @var RedundancyGroup $data */
+        $data = $members->first();
```
nilmerg, line 117: I don't quite like that this is evaluated per group result.

Keep it as it is for now, but update it so that it matches the changes made to the statistics and badges implementation.

I have an alternative implementation in mind, but that's now far to complex for this change and so will be implemented separately. Prior approval of this change, I will request to remove this to not show any badges or output for groups. Until then, it's kept to test the statistics and badges implementation changes.

library/Icingadb/Widget/ItemList/RedundancyGroupListItem.php
```diff
@@ -0,0 +1,153 @@
...
+            $verb = 'has';
+        } else {
+            $verb = 'is';
+        }
+
+        $title->addHtml(Html::sprintf(
+            t('%s %s %s', '<hostname> has/is <state-text>'),
+            $this->createSubject(),
+            $verb,
+            Html::tag('span', ['class' => 'state-text'], $this->state->failed ? 'Failed' : 'OK')
```
nilmerg, line 141: Translatable strings should never be concatenated. Instead, the text to translate should always be complete.

nilmerg (changes requested):
please rebase, #1055 is merged.

library/Icingadb/Widget/ObjectsStateBadges.php
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Icinga\Module\Icingadb\Widget;
+
+use Icinga\Module\Icingadb\Common\StateBadges;
+
+class ObjectsStateBadges extends StateBadges
+{
+    protected function getType(): string
+    {
+        return 'redundancy_group_objects';
```
sukhwinder33445, line 13: This class should be generalized.

library/Icingadb/Widget/ObjectsStateBadges.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    protected function getPrefix(): string
+    {
+        return 'objects';
+    }
+
+    protected function assemble()
```
sukhwinder33445, line 21: Add return type, please.

library/Icingadb/Widget/ItemList/DependencyNodeList.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        $this->initializeDetailActions();
+    }
+
+    protected function getItemClass(): string
+    {
+        return '';
+    }
+
+    protected function createListItem(object $data)
```
sukhwinder33445, line 24: Please add return type.

library/Icingadb/Model/RedundancyGroupParentStateSummary.php
```diff
@@ -0,0 +1,110 @@
...
+
+        return $q;
+    }
+
+    public function getColumns(): array
+    {
+        return array_merge(parent::getColumns(), $this->getSummaryColumns());
+    }
+
+    public function getDefaultSort()
```
sukhwinder33445, line 106: Please add return type.

library/Icingadb/Model/RedundancyGroupParentStateSummary.php
```diff
@@ -0,0 +1,110 @@
...
+            ),
+            'objects_warning_unhandled' => new Expression(
+                'SUM(CASE WHEN redundancy_group_from_to_service_state.soft_state = 1'
+                . ' AND redundancy_group_from_to_service_state.is_handled = \'n\''
+                . ' AND redundancy_group_from_to_service_state.is_reachable = \'y\' THEN 1 ELSE 0 END)'
+            )
+        ];
+    }
+
+    public static function on(Connection $db)
```
sukhwinder33445, line 84: Please add return type.

library/Icingadb/Model/RedundancyGroupParentStateSummary.php
```diff
@@ -0,0 +1,110 @@
...
+ * @property int $objects_total
+ * @property int $objects_up_ok
+ * @property int $objects_unknown_handled
+ * @property int $objects_unknown_unhandled
+ * @property int $objects_warning_handled
+ * @property int $objects_warning_unhandled
+ */
+class RedundancyGroupParentStateSummary extends RedundancyGroup
+{
+    public function getSummaryColumns()
```
sukhwinder33445, line 24: Please add return type.

library/Icingadb/Model/Behavior/HasRootProblem.php
```diff
@@ -0,0 +1,75 @@
...
+use ipl\Orm\AliasedExpression;
+use ipl\Orm\ColumnDefinition;
+use ipl\Orm\Exception\InvalidColumnException;
+use ipl\Orm\Query;
+use ipl\Sql\Expression;
+use ipl\Stdlib\Filter;
+use ipl\Orm\Contract\RewriteColumnBehavior;
+use ipl\Orm\Contract\QueryAwareBehavior;
+
+class HasRootProblem implements RewriteColumnBehavior, QueryAwareBehavior
```
sukhwinder33445, line 17: Please add return type to the methods.

library/Icingadb/Common/StateBadges.php
```diff
@@ -156,34 +170,46 @@ protected function createBadge(string $state)
...
+                            Filter::equal($this->type . '.state.is_handled', 'y'),
+                            Filter::equal($this->type . '.state.is_reachable', 'n')
+                        )
                     )
-                )
-            ));
+                );
+            }
+
+            $content[] = Html::tag('li', $handledStateBadge);
```
sukhwinder33445, line 212: Please use `new HtmlElement()` instead.

library/Icingadb/Common/StateBadges.php
```diff
@@ -156,34 +170,46 @@ protected function createBadge(string $state)
...
+                    $unhandledStateBadge,
+                    Filter::all(
+                        Filter::equal($this->type . '.state.soft_state', $this->getStateInt($state)),
+                        Filter::equal($this->type . '.state.is_handled', 'n'),
+                        Filter::equal($this->type . '.state.is_reachable', 'y')
+                    )
+                );
+            }
+
+            $content[] = Html::tag('li', $unhandledStateBadge);
```
sukhwinder33445, line 193: Please use `new HtmlElement()` instead.

library/Icingadb/Common/StateBadges.php
```diff
@@ -135,15 +143,21 @@ public function createLink($content, Filter\Rule $filter = null): Link
...
+            $stateBadge = new StateBadge($this->item->$key, $state);
+
+            if ($this->url !== null) {
+                $this->createLink(
+                    $stateBadge,
+                    Filter::equal($this->type . '.state.soft_state', $this->getStateInt($state))
+                );
+            }
+
+            return Html::tag('li', $stateBadge);
```
sukhwinder33445, line 160: Please use `new HtmlElement()` instead.

sukhwinder33445 (commented):
Please add PhpDoc and method return type to newly introduced classes/methods.

library/Icingadb/Model/RedundancyGroupParentStateSummary.php
raviks789: IMO, I think there is no need to add return type to this method. But, now that I look at it, I don't think it is necessary to have this method at all.

library/Icingadb/Model/RedundancyGroupParentStateSummary.php
```diff
@@ -0,0 +1,130 @@
...
+ * @property int $objects_down_critical_handled
+ * @property int $objects_down_critical_unhandled
+ * @property int $objects_pending
+ * @property int $objects_problems_unacknowledged
+ * @property int $objects_total
+ * @property int $objects_up_ok
+ * @property int $objects_unknown_handled
+ * @property int $objects_unknown_unhandled
+ * @property int $objects_warning_handled
+ * @property int $objects_warning_unhandled
```
sukhwinder33445, line 25: Please fix these properties.

library/Icingadb/Widget/ObjectsStatistics.php
```diff
@@ -0,0 +1,55 @@
...
+use ipl\Html\Text;
+use ipl\Html\ValidHtml;
+use ipl\Html\HtmlString;
+
+/**
+ * Objects statistics
+ */
+class ObjectsStatistics extends ObjectStatistics
+{
+    protected $summary;
```
sukhwinder33445, line 19: Please add phpDoc.

public/css/common.less
```diff
@@ -403,3 +403,10 @@ div.show-more {
 form[name="form_confirm_removal"] {
   text-align: center;
 }
+
+.state-ball,
+.state-badge {
+  &.state-problem {
```
sukhwinder33445, line 409: As discussed offline, please change the class to `.state-(reachable|unreachable)`.
Please change the method `RedundencyGroupState::getTextState()` accordingly.

sukhwinder33445 (changes requested):

library/Icingadb/Widget/ItemList/DependencyNodeList.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        /** @var UnreachableParent|DependencyNode $data */
+        if ($data->redundancy_group_id !== null) {
+            return new RedundancyGroupListItem($data->redundancy_group, $this);
+        } elseif ($data->service_id !== null) {
+            return new ServiceListItem($data->service, $this);
+        } else {
+            return new HostListItem($data->host, $this);
+        }
+    }
```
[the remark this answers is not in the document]
raviks789: This method was introduced to check the instances of each row in the root problem list. Which in case of a dependency node could be host, service, or redundancy group.

application/controllers/ServiceController.php
```diff
@@ -38,13 +38,15 @@ public function init()
...
+            ->withColumns(['has_root_problem'])
+            ->with([
+                'state',
+                'icon_image',
+                'host',
+                'host.state',
+                'timeperiod'
+            ]);
         $query
             ->setResultSetClass(VolatileStateResults::class)
```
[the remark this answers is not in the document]
raviks789: There was no newline earlier. I just added `->withColumns(['has_root_problem'])`. Please check this again on the main branch.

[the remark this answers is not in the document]
sukhwinder33445: Ohh yes. You are right.

library/Icingadb/Widget/ItemList/RedundancyGroupListItem.php
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    protected function assembleTitle(BaseHtmlElement $title): void
+    {
+        $subject = $this->createSubject();
+        if ($this->state->failed) {
+            $stateTextElement = Html::sprintf(
+                t('%s has %s', '<hostname> has <state-text>'),
+                $subject,
+                new HtmlElement('span', Attributes::create(['class' => 'state-text']), Text::create('FAILED'))
```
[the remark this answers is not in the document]
raviks789: As discussed offline with @nilmerg this will not be a state-text instead the text will be `has no working objects` if the state is failed else `has working objects`.

library/Icingadb/Widget/ObjectsStatistics.php
```diff
@@ -0,0 +1,55 @@
...
+
+use Icinga\Module\Icingadb\Widget\Detail\ObjectStatistics;
+use ipl\Html\Text;
+use ipl\Html\ValidHtml;
+use ipl\Html\HtmlString;
+
+/**
+ * Objects statistics
+ */
+class ObjectsStatistics extends ObjectStatistics
```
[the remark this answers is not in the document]
raviks789: IMO, this is not necessary at this point of time. We could add the class if other CSS rules specific to this class is required.

library/Icingadb/Widget/ObjectsStatistics.php
raviks789: Sure, but don't forget to add the `DependencyNodeSummary` as a possible type for this property in #1069.

public/css/common.less
raviks789: I think this only applies for state-ball. In case of state-badges, the objects summary gives the number of objects in problem (down or critical) state. So the css class state-problem is still necessary in case of state-badges.

library/Icingadb/Widget/ItemList/RedundancyGroupListItem.php
```diff
@@ -0,0 +1,176 @@
...
+                'objects_last_state_change' => new Expression(
+                    'CASE WHEN redundancy_group_from_to_host_state.last_state_change IS NULL'
+                    . ' THEN redundancy_group_from_to_service_state.last_state_change'
+                    . ' ELSE redundancy_group_from_to_host_state.last_state_change END'
+                ),
+                'objects_severity' => new Expression(
+                    'CASE WHEN redundancy_group_from_to_host_state.severity IS NULL'
+                    . ' THEN redundancy_group_from_to_service_state.severity'
+                    . ' ELSE redundancy_group_from_to_host_state.severity END'
+                )
```
nilmerg, line 128: Don't you fetch only host state details here? `redundancy_group_from_to_host_state` should always succeed for services.

library/Icingadb/Widget/ItemList/RedundancyGroupListItem.php
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    protected function assembleTitle(BaseHtmlElement $title): void
+    {
+        $title->addHtml($this->createSubject());
+        if ($this->state->failed) {
+            $title->addHtml(HtmlElement::create('span', null, Text::create(t('has no working objects'))));
+        } else {
+            $title->addHtml(HtmlElement::create('span', null, Text::create(t('has working objects'))));
+        }
```
nilmerg, line 165: Please use ipl-i18n's `Translation` trait here instead of `t()`

library/Icingadb/Widget/ObjectsStateBadges.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+/* Icinga DB Web | (c) 2024 Icinga GmbH | GPLv2 */
+
+namespace Icinga\Module\Icingadb\Widget;
+
+use Icinga\Module\Icingadb\Common\StateBadges;
+
+/**
+ * State badges for the objects
```
nilmerg, line 10: I know I've requested to rename this class. But now I wonder what *the objects* are actually?

I believe you don't want to hear this, but what about `DependencyNodeStateBadges`? :rofl: 

It was about to generalize the name, so *members*, *parents* and *children* were all too specific. Though, *objects* is way to arbitrary now. *dependency node* on the other hand fits all three specific names and is still general enough. :sweat_smile:

library/Icingadb/Widget/ObjectsStatistics.php
```diff
@@ -0,0 +1,57 @@
...
+use Icinga\Chart\Donut;
+
+use Icinga\Module\Icingadb\Model\RedundancyGroupParentStateSummary;
+use Icinga\Module\Icingadb\Widget\Detail\ObjectStatistics;
+use ipl\Html\Text;
+use ipl\Html\ValidHtml;
+use ipl\Html\HtmlString;
+
+/**
+ * Objects statistics
```
nilmerg, line 16: Same here :sweat_smile: `DependencyNodeStatistics` and inside, the properties can be prefixed by `nodes_`

library/Icingadb/Widget/ObjectsStatistics.php
```diff
@@ -0,0 +1,57 @@
...
+    {
+        return Text::create($this->shortenAmount($this->summary->objects_total));
+    }
+
+    protected function createBadges(): ValidHtml
+    {
+        $badges = new ObjectsStateBadges($this->summary);
+        if ($this->hasBaseFilter()) {
+            $badges->setBaseFilter($this->getBaseFilter());
+        }
```
nilmerg, line 53: Are the badges clickable? No? Why do they need a base filter then?

public/css/common.less
```diff
@@ -197,7 +197,7 @@ div.show-more {
   margin-left: 1em / 1.333em; // 1em / h2 font size
 }
 
-.object-detail .plugin-output {
+.object-detail .check-command-output .plugin-output {
```
nilmerg, line 200: Instead of adding a class to the collapsible container, please use `:not(.caption) > .plugin-output`

public/css/list/redundancy-group-list-item.less
```diff
@@ -0,0 +1,21 @@
+.redundancy-group-list-item.list-item {
```
nilmerg, line 1: Is this a specificity thing or why is `.list-item` added here?

public/css/list/redundancy-group-list-item.less
```diff
@@ -0,0 +1,21 @@
...
+  .caption {
+    display: flex;
+    justify-content: space-between;
+
+    .plugin-output {
+      vertical-align: middle;
+      .text-ellipsis();
+    }
```
nilmerg, line 9: Hmm... shouldn't this be handled already because it's a list item? Either way, it should not behave differently than output shown as part of a host or service.

nilmerg (changes requested):

public/css/list/redundancy-group-list-item.less
raviks789: Yes, I wanted to make the rules more specific. But now that you point it out, this seems unnecessary.

public/css/list/redundancy-group-list-item.less
raviks789: `display: flex` rule added to the caption actually breaks `text-overflow: ellipsis`. Hence, I added it here again.

library/Icingadb/Widget/ItemList/DependencyNodeList.php
```diff
@@ -0,0 +1,39 @@
...
+use Icinga\Module\Icingadb\Model\DependencyNode;
+use Icinga\Module\Icingadb\Model\UnreachableParent;
+use ipl\Web\Common\BaseListItem;
+
+/**
+ * Dependency node list
+ */
+class DependencyNodeList extends StateList
+{
+    protected $defaultAttributes = ['class' => ['root-problem-list']];
```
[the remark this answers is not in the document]
raviks789: You are right. And here I don't need to explicitly add `root-problem-list` class either.

library/Icingadb/Widget/ObjectsStatistics.php
raviks789: Its an oversight from me. This is totally unnecessary.

public/css/list/redundancy-group-list-item.less
raviks789: I have just noticed that I have to also clamp the lines down to 2 to replicate the common view mode.

sukhwinder33445 (changes requested):

library/Icingadb/Common/StateBadges.php
```diff
@@ -135,15 +144,21 @@ public function createLink($content, Filter\Rule $filter = null): Link
      *
      * @return ?BaseHtmlElement
      */
-    protected function createBadge(string $state)
+    protected function createBadge(string $state): ?BaseHtmlElement
     {
         $key = $this->prefix . "_{$state}";
 
         if (isset($this->item->$key) && $this->item->$key) {
```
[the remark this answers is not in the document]
raviks789: There was no changes on this line of code.

library/Icingadb/Common/StateBadges.php
```diff
@@ -156,34 +171,46 @@ protected function createBadge(string $state)
...
+                        Filter::equal($this->type . '.state.is_handled', 'n'),
+                        Filter::equal($this->type . '.state.is_reachable', 'y')
+                    )
+                );
+            }
+
+            $content[] = new HtmlElement('li', null, $unhandledStateBadge);
         }
 
         if (isset($this->item->$handledKey) && $this->item->$handledKey) {
```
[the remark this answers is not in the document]
raviks789: Same here

public/css/list/redundancy-group-list-item.less
```diff
@@ -0,0 +1,20 @@
+.redundancy-group-list-item {
+  .caption {
+    display: flex;
+    justify-content: space-between;
+
+    .plugin-output {
+      .line-clamp(2);
```
nilmerg, line 7: please add a comment, explaining that this is necessary due to the caption not being a -webkit-box anymore

public/css/list/redundancy-group-list-item.less
```diff
@@ -0,0 +1,20 @@
+.redundancy-group-list-item {
+  .caption {
+    display: flex;
+    justify-content: space-between;
```
nilmerg, line 4: Instead of this, and because you already have a rule for plugin output, why not making the latter `flex: 1 1 auto`?

public/css/list/redundancy-group-list-item.less
```diff
@@ -0,0 +1,20 @@
...
+
+    .plugin-output {
+      .line-clamp(2);
+    }
+
+    .object-statistics {
+      ul {
+        display: flex;
```
nilmerg, line 12: Why?

public/css/list/redundancy-group-list-item.less
```diff
@@ -0,0 +1,20 @@
...
+      }
+
+      .state-badges {
+        font-size: 0.75em;
```
nilmerg, line 16: This makes badges illegible.

nilmerg (changes requested):

public/css/list/redundancy-group-list-item.less
raviks789: The parent element `main` has `display: block` property, `flex: 1 1 auto` does not work. And changing the `display` property in  `main` to `flex` could have other consequences.

public/css/list/redundancy-group-list-item.less
raviks789: TBH, this should be applied for `dependency-node-state-badges` to make all the state badges appear in a single row.

nilmerg: They are in a single row, all state badges are. For me this rule didn't have any effect, which is why I'm asking.

raviks789: Ok, I should have been clear, this happens when you have both handled and unhandled state badges. And now that I have tested it again, it should be applied to the `dependency-node-state-badges` and all `ul` elements in the `dependency-node-state-badges`.

nilmerg: Then I'm wondering why this needs to be specifically handled for dependency node state badges. They are not the only badges with handled/unhandled states which are shown in a single line.

raviks789: This is caused by `display: flex` on caption.

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -602,4 +604,48 @@ protected function fetchCustomVars()
...
+                'service.state.last_comment',
+                'service.host',
+                'service.host.state'
+            ])->orderBy([
+                'host.state.severity'                       => SORT_DESC,
+                'host.state.last_state_change'              => SORT_DESC,
+                'service.state.severity'                    => SORT_DESC,
+                'service.state.last_state_change'           => SORT_DESC,
+                'service.host.state.severity'               => SORT_DESC,
+                'service.host.state.last_state_change'      => SORT_DESC,
```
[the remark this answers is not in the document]
raviks789: Yes, that was my oversight.

public/css/list/redundancy-group-list-item.less
nilmerg: Okay, seems plausible. But making something else flex because a parent is flex, cannot be the correct solution. Prevent the statistics from wrapping, that's all what's needed.

library/Icingadb/Model/Behavior/HasProblematicParent.php
```diff
@@ -0,0 +1,94 @@
...
+        return $this;
+    }
+
+    public function rewriteColumn($column, ?string $relation = null): ?AliasedExpression
+    {
+        if (! $this->isSelectableColumn($column)) {
+            return null;
+        }
+
+        $path = 'from.';
```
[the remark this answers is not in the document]
raviks789: As discussed offline, instead of using the `Query::createSubQuery` method to form the basic sub query and then join the relevant tables. The entire sub query is manually created. Hence, this line of code becomes obsolete.

library/Icingadb/Model/Behavior/HasProblematicParent.php
```diff
@@ -0,0 +1,94 @@
...
+        $subQueryRelation = $relation !== null ? $relation . $path : $path;
+        $subQuery = $this->query->createSubQuery(new DependencyEdge(), $subQueryRelation, null, false)
```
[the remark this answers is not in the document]
raviks789: Same here.

library/Icingadb/Model/Behavior/HasProblematicParent.php
```diff
@@ -0,0 +1,94 @@
...
+                ['root_dependency_state' => 'dependency_state'],
+                ['root_dependency.id = root_dependency_state.dependency_id']
+            )->joinLeft(
+                ['root_group' => 'redundancy_group'],
+                ['root_group.id = to_dependency_node.redundancy_group_id']
+            )->joinLeft(
+                ['root_group_state' => 'redundancy_group_state'],
+                ['root_group_state.redundancy_group_id = root_group.id']
+            )->where(
+                new Expression("root_dependency_state.failed = 'y' OR root_group_state.failed = 'y'")
```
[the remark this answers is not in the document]
raviks789: The filter has been used as suggested on the new manually created sub query.

library/Icingadb/Model/RedundancyGroupSummary.php
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    public static function on(Connection $db): Query
+    {
+        $q = parent::on($db)->with([
+            'from',
+            'from.to.host',
+            'from.to.host.state',
+            'from.to.service',
+            'from.to.service.state'
```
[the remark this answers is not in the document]
raviks789: You are right , this is redundant.

sukhwinder33445 (changes requested):

library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -90,16 +93,26 @@ protected function assembleIconImage(BaseHtmlElement $iconImage): void
...
 
-            if ((int) $total > 1000) {
+            if ($total > 1000) {
                 $total = '1000+';
+                $tooltip = $this->translate('Up to 1000+ affected objects');
+            } else {
+                $tooltip = sprintf(
+                    $this->translatePlural(
+                        '%d affected object',
+                        'Up to %d affected objects',
```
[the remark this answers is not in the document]
raviks789: We are not sure whether all children are affected. The number in the indicator only says, that up to this number of objects could be affected.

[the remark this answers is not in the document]
sukhwinder33445: I see.

library/Icingadb/Redis/VolatileStateResults.php
```diff
@@ -108,7 +109,19 @@ protected function applyRedisUpdates($rows)
...
+                    $type = 'service';
+                    $row = $row->service;
+                } else {
+                    $type = 'host';
+                    $row = $row->host;
+                }
+
+                $behaviors = $this->resolver->getBehaviors($row->state);
+            } elseif ($type === null) {
                 $behaviors = $this->resolver->getBehaviors($row->state);
```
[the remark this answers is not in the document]
raviks789: No, the type should be checked for each object in the UnreachedParent. You are not sure if it is a host or service or redundancy group.

library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -90,16 +93,26 @@ protected function assembleIconImage(BaseHtmlElement $iconImage): void
...
         ));
 
         if ($this->state->affects_children) {
-            $total = $this->item->affected_children;
+            $total = (int) $this->item->affected_children;
 
-            if ((int) $total > 1000) {
+            if ($total > 1000) {
                 $total = '1000+';
+                $tooltip = $this->translate('Up to 1000+ affected objects');
```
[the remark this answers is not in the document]
raviks789: For the same reason as mentioned below it is `up to`.

sukhwinder33445:
Please check the `Phpstan` errors if there are any related to your changes.

library/Icingadb/Model/Behavior/HasProblematicParent.php
```diff
@@ -0,0 +1,101 @@
...
+
+    public function rewriteColumnDefinition(ColumnDefinition $def, string $relation): void
+    {
+    }
+
+    public function rewriteCondition(Filter\Condition $condition, $relation = null): void
+    {
+        $column = substr($condition->getColumn(), strlen($relation ?? ''));
+
+        if ($column === 'has_problematic_parent') {
```
sukhwinder33445, line 97: Why this change (here and above), the method did exactly the same check?

sukhwinder33445 (changes requested):

library/Icingadb/Model/Behavior/HasProblematicParent.php
raviks789: I want to discuss this offline

library/Icingadb/Model/Behavior/HasProblematicParent.php
```diff
@@ -0,0 +1,101 @@
...
+
+    public function rewriteColumn($column, ?string $relation = null): ?AliasedExpression
+    {
+        if (! $this->isSelectableColumn($column)) {
+            return null;
+        }
+
+        $resolver = $this->query->getResolver();
+        if ($relation !== null) {
+            $serviceTableAlias = str_replace('.', '_', $relation);
```
nilmerg, line 40:
```suggestion
            $serviceTableAlias = $resolver->getAlias($resolver->resolveRelation($relation)->getTarget());
```

library/Icingadb/Model/Behavior/HasProblematicParent.php
```diff
@@ -0,0 +1,101 @@
...
+            $column = $serviceTableAlias . "_$column";
```
nilmerg, line 41:
```suggestion
            $column = $resolver->qualifiyColumnAlias($column, $serviceTableAlias);
```

library/Icingadb/Model/Behavior/HasProblematicParent.php
```diff
@@ -0,0 +1,101 @@
...
+            ->columns([new Expression('1')])
+            ->with([
+                'from',
+                'to',
+                'dependency'
+            ])
+            ->limit(1)
+            ->filter(Filter::any(
+                Filter::equal('to.redundancy_group.state.failed', 'y'),
+                Filter::equal('dependency.state.failed', 'y')
```
nilmerg, line 59: I'm sure the dependency has failed, in case the redundancy group has. So the latter is redundant in the filter.

library/Icingadb/Widget/DependencyNodeStatistics.php
```diff
@@ -0,0 +1,51 @@
...
+use ipl\Html\Text;
+use ipl\Html\ValidHtml;
+use ipl\Html\HtmlString;
+
+/**
+ * Dependency node statistics
+ */
+class DependencyNodeStatistics extends ObjectStatistics
+{
+    /** @var RedundancyGroupSummary Dependency node summary */
```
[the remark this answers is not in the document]
nilmerg: Eh, no?

public/css/list/redundancy-group-list-item.less
```diff
@@ -0,0 +1,7 @@
+.redundancy-group-list-item {
+  .caption {
+    .object-statistics {
+      justify-content: end;
```
nilmerg, line 4: You don't have it under control whether the object statistics are a flex container. The caption is which is deciding that the statistics should be aligned right, so it's responsible anyway.

nilmerg (changes requested):
In case of `StateBadges::getUrl()`, phpstan is right, it's a public interface where we cannot guarantee noone will use it unless a url has been set.

library/Icingadb/Model/Behavior/HasProblematicParent.php
raviks789: You have a point.

nilmerg (changes requested):
Still need to test restrictions, however

nilmerg (approved):
LGTM!

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -602,4 +605,55 @@ protected function fetchCustomVars()
...
+                'service.host',
+                'service.host.state'
+            ])
+            ->setResultSetClass(VolatileStateResults::class)
+            ->orderBy([
+                'host.state.severity',
+                'host.state.last_state_change',
+                'service.state.severity',
+                'service.state.last_state_change',
+                'redundancy_group.state.last_state_change'
```
nilmerg, line 647: one small thing: just ordering after a group's state change date doesn't suffice. A group doesn't have a severity, but still a state, which should be used instead.

nilmerg (changes requested):

nilmerg (approved):

PR #1060, opened 2024-09-18
Add missing support for automated removal of downtimes and comments

### LINKED ISSUE #1059: Downtime: delete all downtimes of a host and its service via icingadb api withouth the need to address individual downtime guids
Today we use the monitoring-module's api to schedule and remove downtimes for hosts and their services on demand in automated ways. The old monitoring module handles the removal in the following way:

`../icingaweb2/monitoring/downtimes/delete-all?host_name=<hostname>`

If I want to do this with icingadb, it's api call references multiple downtimes, combined with their individual guids:
image

Is it possible to remove all downtimes for a host an it's services with icingadb's api without the need to adress these guids?
Something like:

`../icingaweb2/icingadb/downtimes/delete-all?name=<hostname>`

## Additional context
This issues forces us to keep the old monitoring module activated in the background.

nilmerg:
fixes #1059

### DIFF
--- a/application/controllers/CommentsController.php
+++ b/application/controllers/CommentsController.php
@@ -5,6 +5,7 @@
 namespace Icinga\Module\Icingadb\Controllers;
 
 use GuzzleHttp\Psr7\ServerRequest;
+use Icinga\Module\Icingadb\Common\CommandActions;
 use Icinga\Module\Icingadb\Common\Links;
 use Icinga\Module\Icingadb\Forms\Command\Object\DeleteCommentForm;
 use Icinga\Module\Icingadb\Model\Comment;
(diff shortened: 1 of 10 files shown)


application/controllers/CommentsController.php
```diff
@@ -202,4 +182,36 @@ public function searchEditorAction()
         $this->getDocument()->add($editor);
         $this->setTitle(t('Adjust Filter'));
     }
+
+    protected function getCommandTargetsUrl(): Url
+    {
+        return Url::fromPath('__CLOSE__');
+    }
+
+    protected function fetchCommandTargets()
```
sukhwinder33445, line 191: Please add return type `Query`.

application/controllers/DowntimesController.php
```diff
@@ -208,4 +188,40 @@ public function searchEditorAction()
         $this->getDocument()->add($editor);
         $this->setTitle(t('Adjust Filter'));
     }
+
+    protected function getCommandTargetsUrl(): Url
+    {
+        return Url::fromPath('__CLOSE__');
+    }
+
+    protected function fetchCommandTargets()
```
sukhwinder33445, line 197: same here.

sukhwinder33445:
The code works fine.

sukhwinder33445 (approved):

Issue #1058, opened 2024-09-17
Indicate the number of affected children by a particular parent


nilmerg:
We want to show the number of affected children in case a parent has a problem.

**It should look like this:**
![image](https://github.com/user-attachments/assets/122dc833-8a4a-4d00-9feb-b304edc8e376)

## Notes

The monitoring module had such an indicator for services of a host in the past. It was dropped entirely in recent versions, due to technical limitations.

To bypass these, the actual number will be calculated [by Icinga 2](https://github.com/Icinga/icinga2/issues/10158) in advance. This makes it a static value, as calculating it to represent the *actual* affected children is too expensive.

This also excludes the possibility for Icinga DB Web to apply role restrictions on it.

The description for this indicator then can be phrased like this: *Up to \<n\> affected children*

But it should only be shown in case there are actually *any* children affected, `host_state.affects_children` (column name not final yet) should allow to conditionally show it.

This indication then has the only effect of making the element stand out in lists, so that is visible at first glance and the user can click on it, to then analyze the incident more closely.

PR #1062, opened 2024-09-24
Indicate the number of affected children by a particular parent

raviks789:
ref #1058

### DIFF
--- a/library/Icingadb/Model/Host.php
+++ b/library/Icingadb/Model/Host.php
@@ -56,6 +56,7 @@ use ipl\Orm\ResultSet;
  * @property ?string $zone_id
  * @property string $command_endpoint_name
  * @property ?string $command_endpoint_id
+ * @property ?int $affected_children
  */
 class Host extends Model
 {
@@ -111,7 +112,8 @@ class Host extends Model
             'zone_name',
             'zone_id',
             'command_endpoint_name',
-            'command_endpoint_id'
+            'command_endpoint_id',
+            'affected_children'
         ];
     }
 
@@ -155,7 +157,8 @@ class Host extends Model
             'zone_name'                 => t('Zone Name'),
             'zone_id'                   => t('Zone Id'),
             'command_endpoint_name'     => t('Endpoint Name'),
-            'command_endpoint_id'       => t('Endpoint Id')
+            'command_endpoint_id'       => t('Endpoint Id'),
+            'affected_children'         => t('Affected Children'),
         ];
     }
 
--- a/library/Icingadb/Model/Service.php
+++ b/library/Icingadb/Model/Service.php
@@ -51,6 +51,7 @@ use ipl\Orm\ResultSet;
  * @property ?string $zone_id
  * @property string $command_endpoint_name
  * @property ?string $command_endpoint_id
+ * @property ?int $affected_children
  */
 class Service extends Model
 {
@@ -103,7 +104,8 @@ class Service extends Model
             'zone_name',
             'zone_id',
             'command_endpoint_name',
-            'command_endpoint_id'
+            'command_endpoint_id',
+            'affected_children'
         ];
     }
 
@@ -144,7 +146,8 @@ class Service extends Model
             'zone_name'                 => t('Zone Name'),
             'zone_id'                   => t('Zone Id'),
             'command_endpoint_name'     => t('Endpoint Name'),
-            'command_endpoint_id'       => t('Endpoint Id')
+            'command_endpoint_id'       => t('Endpoint Id'),
+            'affected_children'         => t('Affected Children')
         ];
     }
 
--- a/library/Icingadb/Model/State.php
+++ b/library/Icingadb/Model/State.php
@@ -48,6 +48,7 @@ use ipl\Web\Widget\Icon;
  * @property DateTime $last_state_change The time when the node last got a status change
  * @property DateTime $next_check The time when the node will execute the next check
  * @property DateTime $next_update The time when the next check of the node is expected to end
+ * @property bool $affects_children Whether any of the children is affected if there is a problem
  */
 abstract class State extends Model
 {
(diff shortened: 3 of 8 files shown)


library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -137,4 +156,29 @@ protected function assemble(): void
...
     }
+
+    /**
+     * Shorten the given amount to 4 characters max
+     *
+     * @param int $amount
+     *
+     * @return string
+     */
+    protected function shortenAmount(int $amount): string
```
nilmerg: You don't need this. As I discussed with @julianbrost already, Icinga 2 will not count more than 1000 (or so) objects. Just show the raw number and a plus afterwards if there are `1001`, so `1001 -> 1000+`, as there may actually be exactly 1000 objects. :roll_eyes:

library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -93,6 +94,29 @@ protected function assembleTitle(BaseHtmlElement $title): void
...
         ));
+
+        if ($this->state->affects_children && $this->item->affected_children) {
+            $total = $this->item->affected_children;
+
+            if ((int) $total > 1000) {
+                $total = '1000+';
+            }
+
+            $icon = new Icon(Icons::HOST_DOWN);
```
nilmerg, line 105: I'd rather want to see an icon for the unreachable state, which just happens to be equal to the host's icon.

public/css/common.less
```diff
@@ -403,3 +403,12 @@ div.show-more {
 form[name="form_confirm_removal"] {
   text-align: center;
 }
+
+.affected-objects {
+  display: inline-flex;
+  align-items: center;
```
nilmerg, line 409: baseline please

public/css/common.less
```diff
@@ -403,3 +403,12 @@ div.show-more {
...
+  margin-left: 0.5em;
+  color: @text-color-light;
+  padding: 0 0.5em;
+  text-align: center;
```
nilmerg, line 413: this shouldn't make a difference, the badge is only that big as it needs to be

nilmerg (changes requested):
* [ ] The background color of the indicator should always be red
* [ ] The foreground color of the indicator must be light, as in the mockup

library/Icingadb/Widget/ItemList/StateListItem.php
```diff
@@ -93,6 +94,23 @@ protected function assembleTitle(BaseHtmlElement $title): void
...
+            if ((int) $total > 1000) {
+                $total = '1000+';
+            }
+
+            $icon = new Icon(Icons::UNREACHABLE);
+
+            $title->addHtml((new StateBadge([$icon, Text::create($total)], ''))
+                ->addAttributes([
+                'class' => 'affected-objects',
+                'title' => sprintf('%s %s', $total, t('affected objects'))
```
nilmerg, line 110: fix the indentation

nilmerg: and: `sprintf(t('%s affected objects'), $total)`
but: `sprintf(t('Up to %s affected objects'), $total)` is probably really better. I'd like to prevent issues (*there are just 999 objects affected!!11*) already in the first place...

public/css/common.less
```diff
@@ -403,3 +403,10 @@ div.show-more {
 form[name="form_confirm_removal"] {
   text-align: center;
 }
+
+.affected-objects {
```
nilmerg, line 407: this shouldn't affect a list item's height, currently they grow by about 5px in case there is this indicator. Once not all of the items have it (no fake data) this leads to differently sized items, which is bad.

nilmerg (changes requested):

nilmerg (approved):
LGTM! Though, please split this into two PR's one of which includes what we want to merge and the other the adjustments for the fake data. Just in case we need such again for further tests without having the real data.

library/Icingadb/Model/Host.php
```diff
@@ -13,6 +13,7 @@
 use ipl\Orm\Model;
 use ipl\Orm\Relations;
 use ipl\Orm\ResultSet;
+use ipl\Sql\Expression;
```
nilmerg, line 16: unused

library/Icingadb/Model/Service.php
```diff
@@ -13,6 +13,7 @@
 use ipl\Orm\Model;
 use ipl\Orm\Relations;
 use ipl\Orm\ResultSet;
+use ipl\Sql\Expression;
```
nilmerg, line 16: unused

library/Icingadb/Model/State.php
```diff
@@ -11,6 +11,7 @@
 use ipl\Orm\Behavior\MillisecondTimestamp;
 use ipl\Orm\Behaviors;
 use ipl\Orm\Model;
+use ipl\Sql\Expression;
```
nilmerg, line 14: unused

public/css/common.less
```diff
@@ -403,3 +403,11 @@ div.show-more {
...
+  display: inline-flex;
+  align-items: baseline;
+  background-color: @state-critical;
+  color: @default-text-color-inverted;
```
nilmerg, line 411: Please don't use default less variables if there is an alternative non-default: `@text-color-inverted`

library/Icingadb/Common/IcingaRedis.php
nilmerg, line 1: this is only required for the fake data

nilmerg (changes requested):

public/css/common.less
nilmerg: this is a candidate for a draft issue

library/Icingadb/Model/Host.php
nilmerg: this is a candidate for a draft issue

PR #1073, opened 2024-10-08
Hide unreachable parents in root problems

nilmerg:


raviks789 (commented):
This change causes false negatives.

For example, if there is an explicit dependency between two services with same host, and the host is down, then both the services become unreachable but also the dependency fails. And this change would filter out the unreachable parent service, which should not happen.

See the screenshot below for one such case

Screenshot 2024-10-14 at 15 55 39

raviks789:
> This change causes false negatives.
> 
> For example, if there is an explicit dependency between two services with same host, and the host is down, then both the services become unreachable but also the dependency fails. And this change would filter out the unreachable parent service, which should not happen.

This seems to apply to any service-to-service dependency. See the screenshot below.

Screenshot 2024-10-15 at 10 39 08

So put it simply, if the host of the parent service is down, the parent service becomes unreachable. This in turn results in the dependency being failed. But according to the filter the parent service should be reachable.

raviks789 (approved):

raviks789:
If the root itself is unreachable, then an incorrect empty message `You are not authorized to view these objects.` as shown below  is displayed.

Screenshot 2025-01-09 at 12 15 30

This was introduced previously to make sure that the users with permission to only particular objects, know that there are root problems but not authorized to see them.

But now there is also a possibility that the roots are not displayed as they are not reachable. In this scenario, we could either make the empty message to be `No items found.` like earlier  and remove the empty message `You are not authorized to view these objects.` entirely. Or we could only show the message `You are not authorized to view these objects.` only if the roots are reachable and the user does not have permission to see those objects.

nilmerg:
This is an *edge* case (pun intended) only as long as we don't follow implicit dependencies in this widget. There are plans to do so however. I'll ignore this for now.

Issue #1075, opened 2024-10-14
Hostgroup: Searcheditor does not work


sukhwinder33445:
## Describe the bug
![Screenshot 2024-10-14 at 11 55 50](https://github.com/user-attachments/assets/d782d8f5-f880-4aa4-b9dd-74bd24502a3f)

## To Reproduce:

1. Go to `/icingaweb2/icingadb/hostgroup?name=ANY_AVAILABLE`
2. Open the search editor and press the space bar to get suggestions.

nilmerg:
I cannot reproduce this. With which revision of Icinga DB Web and ipl-web does this happen?

PR #1076, opened 2024-10-14
Hostgroup: Add missing `setSuggestionUrl()` call

sukhwinder33445:
fixes https://github.com/Icinga/icingadb-web/issues/1075


nilmerg (changes requested):
Since this also affects the servicegroup detail (which you should have noticed yourself, to be honest) and maybe several others (didn't further check), please find a more general solution. `createSearchEditor` already knows about parameters to preserve, so it is potentially a good idea to also pass them automatically to the suggestion url. As long as this doesn't have any negative side-effects (as not only `name` is passed then), I'd prefer this.

nilmerg (approved):

PR #1077, opened 2024-10-15
Affected objects indicator should not use `StateBadge` class

raviks789:


sukhwinder33445 (approved):

public/css/common.less
```diff
@@ -407,6 +407,7 @@ form[name="form_confirm_removal"] {
 .affected-objects {
   display: inline-flex;
   align-items: baseline;
+  border-radius: 0.4em;
```
nilmerg, line 410: We have a mixin for this, why not use it?

PR #1091, opened 2024-11-12
Introduce child options for service downtimes

### LINKED ISSUE #1090: Proper dependency options while scheduling a downtime
## Describe the bug

The downtime form for services doesn't allow to define `child_options` while Icinga 2's documentation doesn't limit them to host downtimes

## Expected behavior

Service downtimes must be allowed to be scheduled with `child_options`.

nilmerg:
`ScheduleDowntimeCommand` is basically what `ScheduleServiceDowntimeCommand` previously was. Though, with the addition of `get|setForAllServices()` which was only part of `ScheduleHostDowntimeCommand`.

`ScheduleDowntimeCommand::get|setChildOption()` is completely new and replaces `ScheduleHostDowntimeCommand::get|setTriggered()`.

`ScheduleHostDowntimeCommand`, `PropagateHostDowntimeCommand` and `ScheduleServiceDowntimeCommand` are now deprecated.

fixes #1090


nilmerg:
@sukhwinder33445 You should now be able to force the `scheduleDowntimeAction` for redundancy groups to always assume a `ScheduleHostDowntimeForm` as source. But I didn't test, what will happen if a service downtime is scheduled with the `all_services` flag set. Either Icinga 2 ignores it, or complains about it. Please check this.

sukhwinder33445 (approved):
Works fine.

PR #1093, opened 2024-11-18
VolatileStateResults: Properly apply state changes in case of DependecyNode

sukhwinder33445:


library/Icingadb/Redis/VolatileStateResults.php
```diff
@@ -99,68 +106,92 @@ public function rewind(): void
...
-
         $showSourceGranted = $this->getAuth()->hasPermission('icingadb/object/show-source');
 
+        $getKeysAndBehaviors = function (State $state): array {
+            return [$state->getColumns(), $this->resolver->getBehaviors($state)];
+        };
+
         $states = [];
-        $hostStates = [];
+        $serviceHostStates = [];
```
nilmerg, line 116: Why isn't everything in `$states`? I see no reason to differ between *hosts* and *service hosts*.

nilmerg (changes requested):
Why private const and function?

PR #1101, opened 2024-11-26
ObjectSuggestions: Don't apply base filter to $customVars query

sukhwinder33445:
Blocked by: https://github.com/Icinga/icingadb-web/pull/1100

 - The basefilter may contain relations, that are not know to `CustomvarFlat`, which can leads to errors. 
      For example: `RedundancygroupController:completeAction`'s base filter (`(parent|child).redundancy_group.id)`) is not known to `CustomvarFlat`.

### DIFF
--- a/library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
+++ b/library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
@@ -301,7 +301,7 @@ class ObjectSuggestions extends Suggestions
         }
 
         $customVars->columns('flatname');
-        $this->applyBaseFilter($customVars);
+        $this->applyRestrictions($customVars);
         $customVars->filter(Filter::like('flatname', $searchTerm));
         $idColumn = $resolver->qualifyColumn('id', $resolver->getAlias($customVars->getModel()));
         $customVars = $customVars->assembleSelect();


library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -301,7 +301,7 @@ protected function queryCustomvarConfig(string $searchTerm): Select
         }
 
         $customVars->columns('flatname');
-        $this->applyBaseFilter($customVars);
+        $this->applyRestrictions($customVars);
```
nilmerg, line 304: Please explain why this is required in such a global way. At the moment, this disables base filter appliance for all routes that have one and show a searchbar.

nilmerg: e.g. hostgroup, servicegroup

nilmerg (changes requested):

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
sukhwinder33445: As discussed offline, the base filter is not necessary here as it is already applied to the inner query (`$scalarQueries`) with `applyBaseFIlter()`.
Yes, the `applyBaseFIlter()` also applies restrictions to the inner query, but the `denylist/protect` can only be applied correctly if the `flatname` column is fetched.
Since the outer query retrieves the `flatname` column, the restriction must also be applied here (introduced with https://github.com/Icinga/icingadb-web/commit/5a808ae6d2063fba8fd667b214ad4550eb7f08fe).

nilmerg (approved):

PR #1102, opened 2024-11-27
ObjectSuggestions: Fix exotic columns match

sukhwinder33445:
These columns should only be displayed if the full column name is explicitly specified by the user.
   For example: `name_checksum` should show matches, but (`name|checksum|_checksum|ame_checksum`) should not.
Previously, exotic columns were also suggested, even if only the column name suffix matched


nilmerg:
I am not sure about the necessity of this change. Are there chances for _false positives_? I mean sure, `id` is sufficiently short that it may match with custom variables, but since #1215 that seems to be mitigated somewhat. Also, this change doesn't help here anyway, since `id` is most of the time the full name.

PR #1104, opened 2024-12-03
Use new BaseObjectHeader class to create detail view header

sukhwinder33445:


nilmerg (changes requested):
Haven't looked at everything yet, CSS not at all even.

Though, I'm already quite puzzled anyway. My expectation, and I told you this many times, was that nothing (or only with good reasoning) must change in the UI. As can be seen on the screenshots in my comments, this isn't the case.

nilmerg:
Superseded by #1141

PR #1114, opened 2024-12-17
Keep compatibility with Icinga DB v5

nilmerg:
blocked by #1112

### DIFF
--- a/library/Icingadb/Model/Host.php
+++ b/library/Icingadb/Model/Host.php
@@ -5,6 +5,7 @@
 namespace Icinga\Module\Icingadb\Model;
 
 use Icinga\Module\Icingadb\Common\Auth;
+use Icinga\Module\Icingadb\Common\Backend;
 use Icinga\Module\Icingadb\Model\Behavior\BoolCast;
 use Icinga\Module\Icingadb\Model\Behavior\ReRoute;
 use ipl\Orm\Behavior\Binary;
@@ -74,7 +75,7 @@ class Host extends Model
 
     public function getColumns()
     {
-        return [
+        $columns = [
             'environment_id',
             'name_checksum',
             'properties_checksum',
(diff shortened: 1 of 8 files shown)


sukhwinder33445 (approved):
Works fine.

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -617,6 +619,13 @@ protected function fetchCustomVars()
      */
     protected function createRootProblems(): ?array
     {
+        if (Backend::getDbSchemaVersion() < 6 && ! $this->object->state->is_reachable) {
+            return [
+                HtmlElement::create('h2', null, Text::create(t('Root Problems'))),
+                new EmptyStateBar(t('Please update Icinga DB to view root problems.'))
```
nilmerg, line 625: *You're missing out! Uprade Icinga DB and see the actual root cause here!*

library/Icingadb/Model/Host.php
```diff
@@ -112,14 +113,19 @@ public function getColumns()
             'zone_name',
             'zone_id',
             'command_endpoint_name',
-            'command_endpoint_id',
-            'affected_children'
+            'command_endpoint_id'
         ];
+
+        if (Backend::getDbSchemaVersion() >= 6) {
```
yhabteab, line 119: Since [`Icinga DB 1.2.1`](https://github.com/Icinga/icingadb/releases/tag/v1.2.1) was released last year without any changes to dependencies, this should be updated to schema version `7` accordingly.

nilmerg: @sukhwinder33445 Please do so.

yhabteab: Btw. The used schema version isn't the same for both supported DBMs, i.e. current MySQL schema version is `6`, while for PostgreSQL, it is `4`. Is this distinction not relevant in this context?

sukhwinder33445: Ups, yes, it is relevant.

PR #1115, opened 2024-12-17
Adjust `UnreachableParent` model

sukhwinder33445:


nilmerg (changes requested):
It's rather a child id, not a parent id :wink:

nilmerg (approved):

PR #1118, opened 2025-01-09
Macros: Add missing support for `DateTime` type macro

### LINKED ISSUE #1111: Some macros can not be resolved in action URLs
## Describe the bug

Using `$host.state.last_state_change$` macro for action URLs breaks the detail view of affected hosts with the following error message:

```
Icinga\Module\Icingadb\Web\Navigation\Action::resolveMacro(): Return value must be of type string, DateTime returned

```

## To Reproduce

1. Creat a new Navigation Item of the type "Host Action" or "Service Action"
2. Use the macro `$host.state.last_state_change$` in the `URL` field
3. Open detail view of affected Host or Service

## Your Environment

* Icinga Web version: 2.12.1
* Icinga DB Web version: 1.1.3

## Additional context

ref/IP/56978

sukhwinder33445:
fixes #1111

### DIFF
--- a/library/Icingadb/Common/Macros.php
+++ b/library/Icingadb/Common/Macros.php
@@ -127,6 +127,6 @@ trait Macros
             $value = $value ? 'y' : 'n';
         }
 
-        return $value ?? $macro;
+        return (string) ($value ?? $macro);
     }
 }
--- a/library/Icingadb/Common/Macros.php
+++ b/library/Icingadb/Common/Macros.php
@@ -123,6 +123,8 @@ trait Macros
 
         if ($value instanceof DateTime) {
             $value = $value->format(DateTime::ATOM);
+        } elseif (is_bool($value)) {
+            $value = $value ? 'y' : 'n';
         }
 
         return $value ?? $macro;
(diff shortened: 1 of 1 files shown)


library/Icingadb/Common/Macros.php
```diff
@@ -120,6 +121,12 @@ public function resolveMacro(string $macro, $object): string
...
         }
 
-        return $value !== null ? $value : $macro;
+        if ($value instanceof DateTime) {
+            $value = $value->getTimestamp();
+        } elseif (is_bool($value)) {
+            $value = $value ? 'y' : 'n';
+        }
+
+        return $value ?? $macro;
```
nilmerg, line 130: Cast to string here, it might still be an int or float I assume.

sukhwinder33445: The specified return type of the function is responsible for the cast.

nilmerg: Hm, yeah, as if [implicit type conversion did never turn out to be an issue](https://php.watch/versions/8.1/deprecate-implicit-conversion-incompatible-float-string).

Explicit is better than implicit.

nilmerg (approved):

Issue #1105, opened 2024-12-04
Show "process check result" action link only when it is a passive service


log1-c:
Currently the "process check result" link is always shown.
But, unless the service is a passive check or the "Passive checks" feature command is enabled, it is without function.
Therefore it could/should not be displayed as long as the requirements are not set.
![image](https://github.com/user-attachments/assets/8486a14c-1343-47e1-ac8e-bf3173872b38)

PR #1120, opened 2025-01-09
QuickActions: Hide `Process check` and `Notification` button when feature is disabled

sukhwinder33445:
fixes https://github.com/Icinga/icingadb-web/issues/1105

### DIFF
--- a/library/Icingadb/Widget/Detail/QuickActions.php
+++ b/library/Icingadb/Widget/Detail/QuickActions.php
@@ -104,7 +104,10 @@ class QuickActions extends BaseHtmlElement
             );
         }
 
-        if ($this->isGrantedOn('icingadb/command/process-check-result', $this->object)) {
+        if (
+            $this->object->passive_checks_enabled
+            && $this->isGrantedOn('icingadb/command/process-check-result', $this->object)
+        ) {
             $this->assembleAction(
                 'processCheckresult',
                 t('Process check result'),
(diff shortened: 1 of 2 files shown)


library/Icingadb/Widget/Detail/QuickActions.php
```diff
@@ -71,7 +71,10 @@ protected function assemble()
             );
         }
 
-        if ($this->isGrantedOn('icingadb/command/send-custom-notification', $this->object)) {
+        if (
+            $this->isGrantedOn('icingadb/command/send-custom-notification', $this->object)
+            && $this->object->notifications_enabled //TODO: not sure, Model contains a forced switch
```
nilmerg, line 76: Yeah, and because of that switch this is nonsense. The tooltip of the option clearly states this.

library/Icingadb/Widget/Detail/QuickActions.php
```diff
@@ -104,7 +107,10 @@ protected function assemble()
             );
         }
 
-        if ($this->isGrantedOn('icingadb/command/process-check-result', $this->object)) {
+        if (
+            $this->isGrantedOn('icingadb/command/process-check-result', $this->object)
+            && $this->object->passive_checks_enabled
```
nilmerg, line 112: Check the enable flag first, it increases efficiency as it's less costly to check

nilmerg (changes requested):
The multi select quick actions also need adjustment

nilmerg (approved):

PR #1121, opened 2025-01-09
Use severity as default sort

### LINKED ISSUE #1096: Icinga DB Hosts/Service view sort by Severity as default
When viewing hosts/services status in Icinga DB, the object is always sorted by name.

When viewing hosts/services status in Icinga DB, the object is always sorted by Severity as the default. This is infinitely more useful.

The alternative is to use the old 'Host Problems' / 'Service Problems' status as it's sorted by Severity by default.

## Additional context
Apparently you can modify code to make this change. This is a change that should be available in the UI, or just make it the default, as sorting by name is not helpful at all.

sukhwinder33445:
resolves #1096

requires https://github.com/Icinga/ipl-web/pull/241

### DIFF
--- a/application/controllers/HostsController.php
+++ b/application/controllers/HostsController.php
@@ -57,7 +57,8 @@ class HostsController extends Controller
                 'host.state.severity desc,host.state.last_state_change desc' => t('Severity'),
                 'host.state.soft_state'                                      => t('Current State'),
                 'host.state.last_state_change desc'                          => t('Last State Change')
-            ]
+            ],
+            ['host.state.severity DESC', 'host.state.last_state_change DESC']
         );
         $viewModeSwitcher = $this->createViewModeSwitcher($paginationControl, $limitControl);
         $columns = $this->createColumnControl($hosts, $viewModeSwitcher);
--- a/application/controllers/ServicesController.php
+++ b/application/controllers/ServicesController.php
@@ -68,7 +68,8 @@ class ServicesController extends Controller
                 'service.state.soft_state'                                         => t('Current State'),
                 'service.state.last_state_change desc'                             => t('Last State Change'),
                 'host.display_name'                                                => t('Host')
-            ]
+            ],
+            ['service.state.severity DESC', 'service.state.last_state_change DESC']
         );
         $viewModeSwitcher = $this->createViewModeSwitcher($paginationControl, $limitControl);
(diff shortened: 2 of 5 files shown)


application/controllers/HostsController.php
```diff
@@ -50,15 +50,19 @@ public function indexAction()
 
         $limitControl = $this->createLimitControl();
         $paginationControl = $this->createPaginationControl($hosts);
+
+        $defaultSort = 'host.state.severity desc,host.state.last_state_change desc';
```
raviks789, line 54:
```suggestion
        $defaultSort = ['host.state.severity DESC', 'host.state.last_state_change DESC'];
```

This looks more readable to me.

application/controllers/HostsController.php
```diff
@@ -50,15 +50,19 @@ public function indexAction()
...
         $sortControl = $this->createSortControl(
             $hosts,
             [
                 'host.display_name'                                          => t('Name'),
                 'host.state.severity desc,host.state.last_state_change desc' => t('Severity'),
```
raviks789, line 59:
```suggestion
                'host.state.severity desc,host.state.last_state_change desc' => t('Severity'),
                'host.display_name'                                          => t('Name'),
```

You need to also change the order of this array to reflect that the hosts were ordered by severity by default.

application/controllers/ServicesController.php
```diff
@@ -60,6 +60,8 @@ public function indexAction()
 
         $limitControl = $this->createLimitControl();
         $paginationControl = $this->createPaginationControl($services);
+
+        $defaultSort = 'service.state.severity desc,service.state.last_state_change desc';
```
raviks789, line 64:
```suggestion
        $defaultSort = ['service.state.severity DESC', 'service.state.last_state_change Desc'];
```

application/controllers/ServicesController.php
```diff
@@ -60,6 +60,8 @@ public function indexAction()
...
         $sortControl = $this->createSortControl(
             $services,
             [
```
raviks789, line 67: Same here, change the order of the array to reflect the default sort.

raviks789 (approved):

raviks789:
I believe the services under a host should also be sorted according to their severity by default. I mean in `HostController::servicesAction`.

PR #1122, opened 2025-01-09
ObjectInspectionDetail: Show host/service templates

### LINKED ISSUE #1095: Show templates of host/service in source tab
No problem but sometimes it would be very helpful

Show the templates which were imported by the host or service in the source tab

sukhwinder33445:
resolves #1095

### DIFF
--- a/library/Icingadb/Common/ObjectInspectionDetail.php
+++ b/library/Icingadb/Common/ObjectInspectionDetail.php
@@ -169,7 +169,6 @@ abstract class ObjectInspectionDetail extends BaseHtmlElement
             'display_name',
             'last_check_result',
             'source_location',
-            'templates',
             'package',
             'version',
             'type',


raviks789:
The object name is also shown along with the templates imported as shown in the screenshot below, which IMO opinion must be removed.

Screenshot 2025-01-10 at 13 46 55

nilmerg:
> which IMO opinion must be removed.

Why? It's what the API returns, isn't it? The source tab is meant to provide raw information.

nilmerg:
It's totally out of scope to adjust what the API returns. And we should not hide anything provided by it. The information might be confusing to you. For me this is a template tree in which case a root is required and that's the object itself.

Anyway, the source tab is meant to provide raw information. Some of it might be represented differently (timestamps -> datetime, bool -> yes|no), but it never should hide anything.

nilmerg (approved):

PR #1124, opened 2025-01-13
Keep compatibility with Icinga DB v6

sukhwinder33445:
ref https://github.com/Icinga/icingadb-web/pull/1114

### DIFF
--- a/library/Icingadb/Common/Backend.php
+++ b/library/Icingadb/Common/Backend.php
@@ -29,6 +29,9 @@ final class Backend
     /** @var ?IcingaRedis */
     private static $redis;
 
+    /** @var ?bool Whether the current Icinga DB version supports dependencies */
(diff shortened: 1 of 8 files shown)


library/Icingadb/Common/Backend.php
```diff
@@ -165,4 +165,18 @@ public static function getRedis(): IcingaRedis
 
         return self::$redis;
     }
+
+    /**
+     * Check if the Icinga DB supports dependencies
+     *
+     * @return bool
+     */
+    public static function supportDependencies(): bool
```
nilmerg, line 174: * rename to *support**s**Dependencies*
* cache the return value

nilmerg (changes requested):

nilmerg (approved):

PR #1128, opened 2025-01-17
RedundancygroupController: Separate `(searchEditor/complete)Action` for `members & children tab`

sukhwinder33445:


application/controllers/RedundancygroupController.php
```diff
@@ -174,19 +174,22 @@ public function childrenAction(): Generator
...
+        ];
+
+        $requestParams = Url::fromRequest()->onlyWith($preserveParams)->getParams();
+        $searchBar = $this->createSearchBar($nodesQuery, $preserveParams)
+            ->setEditorUrl(
+                Url::fromPath('icingadb/redundancygroup/children-search-editor')
+                    ->setParams($requestParams)
+            )->setSuggestionUrl(
+                Url::fromPath('icingadb/redundancygroup/children-complete')
+                    ->setParams($requestParams)
```
nilmerg, line 191: You share one params object with two separate and very distinct url objects. I don't recommend this, clone it.

nilmerg (changes requested):

nilmerg (approved):

PR #1130, opened 2025-01-20
RedundancyGroup: Don't wrap state msg into `span`

sukhwinder33445:


sukhwinder33445:
Added to https://github.com/Icinga/icingadb-web/pull/1098.

Issue #1138, opened 2025-02-19
There is no timout for actual APICommandTransport calls


SebastianOpeni:
## Describe the bug
An unreachable command Transport endpoint will block the execution for a very long time until a second CommandTransportEndpoint will be tried.
When configuring the Endpoints the probe call utilizes such a timeout but the actual calls are run without such a configuration.

Probe with timeout: https://github.com/Icinga/icingadb-web/blob/main/library/Icingadb/Command/Transport/ApiCommandTransport.php#L314
Actual call without: https://github.com/Icinga/icingadb-web/blob/main/library/Icingadb/Command/Transport/ApiCommandTransport.php#L240

## To Reproduce
Provide a link to a live example, or an unambiguous set of steps to reproduce this issue. Include configuration, logs, etc. to reproduce, if relevant.

1. Set up a fake non working CommandTransport Endpoint ( for example with "ncat -lvnp 5665 --keep-open") to simulate an outage
2. Configure said fake commandTranspoirt Endpoint (force its config, since the ui-config will correctly fail after 15 seconds)
3. Configure a working CommandTransport Endpoint as second option
5. Try to send a Command (for example a "recheck")
6. The "recheck command" will not be send to the second Endpoint for multiple minutes and the ui will freeze.
This makes working with icingaweb impossible without changing the commandTransport Endpoint order.

## Expected behavior
After the first commandEndpoint is not reacting for some time (for example if a network error happened), I would expect it to switch over the the second endpoint after some seconds and not after multiple minutes.

## Logs
2025-02-19T10:46:10+01:00 - DEBUG - Sending Icinga command "actions/reschedule-check" to the API "testserver:5665"
2025-02-19T10:51:10+01:00 - ERROR - Icinga\Module\Icingadb\Command\Transport\CommandTransportException in /usr/share/icingaweb2/modules/icingadb/library/Icingadb/Command/Transport/ApiCommandTransport.php:249 with message: Can't connect to the Icinga 2 API: 0 cURL error 28: SSL connection timeout (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://testserver.de:5665/v1/actions/reschedule-check
2025-02-19T10:51:10+01:00 - DEBUG - Sending Icinga command "actions/reschedule-check" to the API "correcthost:5665"

## Your Environment
Include as many relevant details about the environment you experienced the problem in.

* Icinga DB Web version (System - About): 1.1.3
[...]

PR #1139, opened 2025-02-19
Fix/set api timeout also for actual api call

SebastianOpeni:
fixes: https://github.com/Icinga/icingadb-web/issues/1138


nilmerg (approved):

nilmerg:
Will not merge right away. But I will, eventually. :)

PR #1155, opened 2025-04-04
Fix that host / servicegroup bulk processing options affect objects not in said group

### LINKED ISSUE #1154: Host-/Servicegroup bulk processing options affect objects not in said group
## Describe the bug
Using any of the bulk processing options of a host or service group detail, might affect objects that are not in the same group. The statistics also show too much objects.

## To Reproduce

1. Open a host or servicegroup detail
2. Use a filter that will match some objects in the group **and not in the group** (The list will not include them, though)
3. Open the bulk processing details
4. Note that the statistics widget shows more objects and that any action affects them as well

## Expected behavior
Only the actual group members are affected.

## Your Environment

* Icinga DB Web version (System - About): main

nilmerg:
fixes #1154
requires https://github.com/Icinga/ipl-web/issues/263

### DIFF
--- a/application/controllers/HostgroupController.php
+++ b/application/controllers/HostgroupController.php
@@ -118,7 +118,11 @@ class HostgroupController extends Controller
         $this->addControl($limitControl);
         $this->addControl($viewModeSwitcher);
         $this->addControl($searchBar);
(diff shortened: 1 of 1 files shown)


raviks789 (approved):
LGTM!

PR #1158, opened 2025-04-07
CSS: Invert affected children element

flourish86:


public/css/common.less
```diff
@@ -303,8 +303,11 @@ form[name="form_confirm_removal"] {
 .affected-objects {
   display: inline-flex;
   align-items: baseline;
-  background-color: @state-critical;
-  color: @text-color-inverted;
+  color: @state-critical;
   padding: 0 0.25em;
   .rounded-corners();
```
sukhwinder33445, line 308: Please remove this as there is now no background and this has no visible effect.

public/css/common.less
```diff
@@ -303,8 +303,11 @@ form[name="form_confirm_removal"] {
...
+
+  .icon {
+    opacity: .9;
```
sukhwinder33445, line 311: Now the count is slightly darker than the icon, is that intention?

sukhwinder33445 (changes requested):

public/css/common.less
flourish86: Yes, since icons are usually bolder than fonts it's common practive to tone them down with opacity. We already do this in different places, e.g. menu nav items.

sukhwinder33445 (approved):

PR #1173, opened 2025-04-25
ObjectSuggestion: Dynamically adjust columns and custom var suggestion limits

### LINKED ISSUE #1094: Custom vars not displayed in searchbar in case it matches the object type.
## Describe the bug
The searchbar has an bug with "short" custom vars, that matches the object type.

It seems "built-in" variables are prioritized inside the SearchBar. Custom vars are display at the bottom. Also I think there is an hardcoded limit of suggestions that will be displayed.

"os" is part of the string "host". So if you try to search for "os" inside of `icingadb/hosts`  it will always show the "built-in" variables first. The custom var will not be shown.

## To Reproduce
1. Create a custom var "os"
2. Try to use it inside of `icingadb/hosts`
3. You wont find it

## Expected behavior
The custom var should be displayed when it is matching 100%.

## Your Environment

* Icinga DB Web version (System - About):  1.1.3
* Icinga Web 2 version (System - About): 2.12.1
* icinga/icinga-php-library: 0.14.1
* icinga/icinga-php-thirdparty: 0.12.1

ref/NC/836693

raviks789:
Currently, the limit is globally set to 50, hence the short custom variable names which overlaps with most of the column names will not be suggested and instead only column names will be suggested. To prevent this the default limit is split in half and adjusted dynamically based on available suggestions.

resolves #1094

### DIFF
--- a/library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
+++ b/library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
@@ -226,26 +226,30 @@ class ObjectSuggestions extends Suggestions
         $varSuggestions = [];
         $exactVarMatches = [];
         $halfLimit = (int) (static::DEFAULT_LIMIT / 2);
-        foreach ($this->getDb()->select($this->queryCustomvarConfig($searchTerm)) as $customVar) {
-            $search = $name = $customVar->flatname;
-            if (preg_match('/\w+(?:\[(\d*)])+$/', $search, $matches)) {
-                $name = substr($search, 0, -(strlen($matches[1]) + 2));
-                if (isset($parsedArrayVars[$name])) {
-                    continue;
+
+        $exactSearchTerm = trim($searchTerm, ' *');
+        if ($exactSearchTerm !== '') {
+            foreach ($this->getDb()->select($this->queryCustomvarConfig($searchTerm)) as $customVar) {
+                $search = $name = $customVar->flatname;
+                if (preg_match('/\w+(?:\[(\d*)])+$/', $search, $matches)) {
+                    $name = substr($search, 0, -(strlen($matches[1]) + 2));
+                    if (isset($parsedArrayVars[$name])) {
+                        continue;
+                    }
+
+                    $parsedArrayVars[$name] = true;
+                    $search = $name . '[*]';
                 }
 
-                $parsedArrayVars[$name] = true;
-                $search = $name . '[*]';
-            }
-
-            foreach ($this->customVarSources as $relation => $label) {
-                if (isset($customVar->$relation)) {
-                    $varRelation = $relation . '.vars.' . $search;
-                    $varLabel = sprintf($label, $name);
-                    if ($search === trim($searchTerm, '*')) {
-                        $exactVarMatches[$varRelation] = $varLabel;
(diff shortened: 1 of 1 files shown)


nilmerg (commented):
My first impressions:

* I think *Best Matches* or *Best Candidates* or *Best Suggestions* is a better suited phrase
* The columns title is shown without any suggestions
* It's way slower than before. Sure, now are custom variables always fetched, but still it feels slower
  * I also think showing 25 custom variables by default (i.e. with no input) is of no advantage. They're not ordered, somewhat random, unlike columns which include the base table columns first. Custom variables should only be fetched first if the user provides input.

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -221,41 +221,98 @@ protected function fetchColumnSuggestions($searchTerm)
...
+
+            foreach ($exactVarMatches as $relation => $label) {
+                yield $relation => $label;
+            }
+        }
+
+        // Ordinary columns comes after exact matches,
+        // or if there ar no exact matches they come first
+        $ordinaryColumns = self::collectFilterColumns($model, $query->getResolver());
+        if (! empty($ordinaryColumns)) {
```
nilmerg, line 281: collectFilterColumns returns a generator, which is never empty

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -221,41 +221,98 @@ protected function fetchColumnSuggestions($searchTerm)
...
+                    $search = $name . '[*]';
                 }
 
-                $parsedArrayVars[$name] = true;
-                $search = $name . '[*]';
+                foreach ($this->customVarSources as $relation => $label) {
+                    if (isset($customVar->$relation)) {
+                        $varRelation = $relation . '.vars.' . $search;
+                        $varLabel = sprintf($label, $name);
+                        if ($search === $exactSearchTerm) {
```
nilmerg, line 248: `$search` is changed in case of arrays and such won't appear as best suggestion as a result. Use `$name` instead.

nilmerg (changes requested):

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -221,39 +221,93 @@ protected function fetchColumnSuggestions($searchTerm)
...
-                $name = substr($search, 0, -(strlen($matches[1]) + 2));
-                if (isset($parsedArrayVars[$name])) {
-                    continue;
+        $varSuggestions = [];
+        $exactVarMatches = [];
+        $halfLimit = (int) (static::DEFAULT_LIMIT / 2);
+
+        $exactSearchTerm = trim($searchTerm, ' *');
+        if ($exactSearchTerm !== '') {
+            foreach ($this->getDb()->select($this->queryCustomvarConfig($searchTerm)) as $customVar) {
```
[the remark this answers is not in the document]
raviks789: The generator enforces that the total suggestions is 50. I need the limit of the fetched custom variables to be 50 because if the column count is less than 25 then remaining suggestion slots would be occupied by custom variables.

nilmerg (approved):

PR #1175, opened 2025-04-28
Fix that detail extensions produce gaps

### LINKED ISSUE #1133: Object detail extensions may produce unnecessary gaps
## Describe the bug
Image

## To Reproduce
Have a hook that renders conditionally and have it render nothing.

## Expected behavior
No gap. No unnecessary reserved space.

nilmerg:
fixes #1133


raviks789 (commented):
LGTM! Just a question though. Should we also not do the same when rendering the native extensions in the object detail view? I mean in `ObjectDetail::createExtensions` method

nilmerg:
Yes!

raviks789 (approved):

PR #1178, opened 2025-05-05
Support setting a default comment for acknowledgements

marianrh:
I propose to add a new option for setting a default comment for acknowledging problems (`acknowledge_comment_text`), similar to the existing `hostdowntime_comment_text` and `servicedowntime_comment_text` options.

This originally was a pull request for Icinga Web 2 (https://github.com/Icinga/icingaweb2/pull/5365). I was asked to implement this in Icinga DB Web as well. *However, since I'm not using Icinga DB Web myself, this is untested.*

Beside adding the new feature, I also fixed a copy-and-paste error for the `servicedowntime_comment_text` option.


nilmerg (approved):
Yeeha! Thank you!

Issue #1192, opened 2025-05-21
Object source is missing all API results


nilmerg:
## Describe the bug
The source tab of an object doesn't include any results from the Icinga 2 API anymore.

## To Reproduce
Use latest default branch of Icinga 2.

## Expected behavior
Attributes, source location and check results fetched from the api should appear.

## Screenshots
![Image](https://github.com/user-attachments/assets/7a89e1d6-f7a9-4ea7-b0cc-f135bf353424)

## Additional context
Caused by https://github.com/Icinga/icinga2/issues/8167

PR #1200, opened 2025-05-27
Fix missing object source results

sukhwinder33445:
fixes https://github.com/Icinga/icingadb-web/issues/1192


nilmerg (changes requested):
Please leave the renderer alone. We should actually require each attribute. Just like we do with database queries. Let's actually learn of this issue and do better, shall we?

nilmerg:
Huh? Please elaborate

nilmerg (approved):
Whatever :man_shrugging:

PR #1201, opened 2025-05-28
Prevent `.icinga-info` contents from colliding

flourish86:
The width of the key-value elements varies, so in some cases there's no gap between them. This makes is hard to grasp the information.

Screenshot 2025-05-28 at 15 39 14

I added some space between the elements to prevent that.

Screenshot 2025-05-28 at 15 41 38

### DIFF
--- a/public/css/widget/monitoring-health.less
+++ b/public/css/widget/monitoring-health.less
@@ -43,11 +43,7 @@
 
 
   .icinga-info {
-    margin-bottom: -2em;
-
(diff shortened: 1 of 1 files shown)


nilmerg (changes requested):
icinga-info is a flex container, so use `gap` please

flourish86:
> icinga-info is a flex container, so use `gap` please

Yessssss! That hovered totally under my radar for flex-box

PR #1205, opened 2025-06-04
Fix long running command actions

### LINKED ISSUE #1204: Make the downtimes-window responsive, in case a large number of downtimes is set
I had to set a downtime for all objects of a zone (90 hosts, 450 services). After selecting all objects and activating the downtime, I was eagerly waiting for any kind of progress or even an indication of success after everything is completed. After 5 min I closed the downtime-window, because there was no response.

It whould be helpful to get some ui responsiveness, in case an action takes a longer time.

nilmerg:
Reduces risk of timeouts in case Icinga takes its time to respond.
The timeout of 15 seconds introduced earlier has been kept.

Chunk sizes were chosen as follows:
- 50: Expected disk writes (comments)
- 250: Only used for process checkresult
- 500: Cheap calculations (object features, check scheduling)

fixes #1204

@yhabteab Please comment my chosen chunk sizes. I'm not quite sure how realistic they are.


yhabteab:
> Are you sure? Processing check results is limited to objects where passive checks are enabled. And the implementation is the same as for the other commands. It'd be broken for them as well.

No, due the dependencies I have in my config, some of my services get automatically rescheduled and instantly changed their states. So if I select all services in a WARNING state, and submit a process check result, there are still some services left in a WARNING state, and look like some leftovers from the request. Works fine now!

> Especially if the _all_services_ and _trigger_children_ options are enabled. I'll change it to this then:

I guess, so! Wouldn't power of two numbers be more appropriate though? :)

nilmerg:
> Wouldn't power of two numbers be more appropriate though? :)

That's just the coder mind of yours :rofl: 
Mine is using SI units most of the time, and since it's my code … :stuck_out_tongue: 

> Your commit description for https://github.com/Icinga/icingadb-web/commit/24fac04488adb7791f8758fcc06c20b7ad1ee722 is now outdated!

Good catch!

raviks789 (dismissed):
LGTM!

raviks789 (commented):
LGTM! 

Just one thing though. IMO, since the timeout value 15 is used in 3 places in the same code, it would be better to define a constant for it.

raviks789 (approved):

PR #1209, opened 2025-06-05
StateBadges: Actually render links for single badges

### LINKED ISSUE #1160: Link ok/up badges in Host-/Service Group List Items
While all other status item badges are linked in host-/service group list items, the ok/up ones are not.

Badges are spread throughout our Icinga Webs user interface and are expected to be clickable and although the ok/down ones, might be the least relevant to be listed, users probably expect that they can click them. On the other hand, I don't see a reason, why they shouldn't be.

### Host Group Badges
Image

### List Item Footer
Image

### Sidebar Menu
Image

### Host Group Ok Badges
Image

nilmerg:
fixes #1160


raviks789 (commented):
LGTM!

PR #1212, opened 2025-06-06
Show Icinga DB and Redis version in health

### LINKED ISSUE #1188: Health: Show Icinga DB and Redis Version

Show both the Icinga DB and Redis version in the Health page.
The latter may be required for a quick check against the latest security patch.

Show both versions somewhere under Health.

One could monitor this with data from the `icingadb` and some Redis-check or from the packages. However, this information would not be available next to the Icinga 2 version in the Health output.

## Additional context
This feature request was brought forward by a customer, Icinga RT 58420.

nilmerg:
resolves #1188

### DIFF
--- a/library/Icingadb/ProvidedHook/RedisHealth.php
+++ b/library/Icingadb/ProvidedHook/RedisHealth.php
@@ -6,6 +6,7 @@ namespace Icinga\Module\Icingadb\ProvidedHook;
 
 use Exception;
 use Icinga\Application\Hook\HealthHook;
+use Icinga\Module\Icingadb\Common\Backend;
 use Icinga\Module\Icingadb\Common\Database;
 use Icinga\Module\Icingadb\Common\IcingaRedis;
 use Icinga\Module\Icingadb\Model\Instance;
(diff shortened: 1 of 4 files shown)


library/Icingadb/ProvidedHook/RedisHealth.php
```diff
@@ -39,10 +41,15 @@ public function checkHealth()
...
 
+            $this->setMetrics($server);
+
             $outdatedDbHeartbeat = $instance->heartbeat->getTimestamp() < time() - 60;
             if (! $outdatedDbHeartbeat || $instance->heartbeat->getTimestamp() <= $lastIcingaHeartbeat) {
                 $this->setState(self::STATE_OK);
-                $this->setMessage(t('Redis available and up to date.'));
+                $this->setMessage(sprintf(
+                    t('Redis available and up to date. (Version: %s)'),
+                    $server['redis_version'] ?? t('N/A')
```
oxzi, line 51: Since this message now contains a specific version, I would interpret this as the Redis server version is up to date. However, the guarding `if` checks the heartbeat age, not the package version. Thus, this can be misleading.

```suggestion
                    t('Data from Redis is up to date. (Version: %s)'),
                    $server['redis_version'] ?? t('N/A')
```

oxzi (commented):
Thanks for this PR! Works as intended.

On the current Icinga DB `main` branch, it complains about an outdated Icinga DB version, which makes somehow sense. However, this is a situation an end user will not face.
Outdated Redis Version

On the current `release-1.4.0` branch, everything is green.
Updated Redis Version

The new information column results in a second row. Please compare it with this screenshot without this patch. Dunno, I just stumbled about this.

Single row without this PR

PR #1215, opened 2025-06-11
ObjectSuggestions: Prioritze Column suggestions and best suggestions

### LINKED ISSUE #1206: Restore suggestion responsiveness
Since https://github.com/Icinga/icingadb-web/pull/1173 suggestions are delayed until custom variables were identified. This leads to a noticeable delay of column suggestions. Previously they were prioritized and as long as there were 50 of them identified, custom variables were not even fetched.

I propose to still prioritize columns over custom variables. Though, we can keep the new *Best Suggestions* feature.

Reason for this is, a best suggestion is technically an exact match. An exact match is very easy to identify for a database. So, let's query the custom variables still first, but only look for exact matches. (i.e. remove all trailing and leading wildcards from the search term but still make sure there's no wildcard left)

Then again, columns next. They are allowed to consume the limit of 50. So that custom variables that don't exactly match only appear if there are not enough columns.

raviks789:
Show custom variable suggestions only if the column suggestions have not consumed all the slots (50).

fixes #1206

### DIFF
--- a/library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
+++ b/library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
@@ -221,13 +221,21 @@ class ObjectSuggestions extends Suggestions
         $model = $this->getModel();
         $query = $model::on($this->getDb());
 
-        // Collect custom variables
         $parsedArrayVars = [];
-        $varSuggestions = [];
         $exactSearchTerm = trim($searchTerm, ' *');
+        $exactVarMatches = [];
         $titleAdded = false;
+
+        // Suggest exact custom variable matches first
         if ($exactSearchTerm !== '') {
-            foreach ($this->getDb()->select($this->queryCustomvarConfig($searchTerm)) as $customVar) {
+            foreach (
+                $this->getDb()->select($this->queryCustomvarConfig(
+                    Filter::any(
+                        Filter::equal('flatname', $exactSearchTerm),
+                        Filter::like('flatname', $exactSearchTerm . '[*]') // Filter for array type custom variables
+                    )
+                )) as $customVar
+            ) {
                 $search = $name = $customVar->flatname;
                 if (preg_match('/\w+(?:\[(\d*)])+$/', $search, $matches)) {
                     $name = substr($search, 0, -(strlen($matches[1]) + 2));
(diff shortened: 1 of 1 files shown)


library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -221,13 +221,11 @@ protected function fetchColumnSuggestions($searchTerm)
...
+        // Collect custom variables
         $parsedArrayVars = [];
         $varSuggestions = [];
-        $exactVarMatches = [];
-        $halfLimit = (int) (static::DEFAULT_LIMIT / 2);
-
         $exactSearchTerm = trim($searchTerm, ' *');
+        $titleAdded = false;
         if ($exactSearchTerm !== '') {
             foreach ($this->getDb()->select($this->queryCustomvarConfig($searchTerm)) as $customVar) {
```
nilmerg, line 230: > So, let's query the custom variables still first, but only look for exact matches. (i.e. remove all trailing and leading wildcards from the search term but still make sure there's no wildcard left)

nilmerg (changes requested):

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -244,70 +250,81 @@ protected function fetchColumnSuggestions($searchTerm)
...
-        }
 
-        // Exact custom variable matches first
-        if (! empty($exactVarMatches)) {
-            $this->addHtml(HtmlElement::create(
-                'li',
-                ['class' => static::SUGGESTION_TITLE_CLASS],
-                t('Best Suggestions')
-            ));
+                        $exactVarMatches[] = $varRelation;
```
nilmerg, line 263: Can't you register `$search` here to avoid checking this over and over while iterating through the sources below?

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -244,70 +250,81 @@ protected function fetchColumnSuggestions($searchTerm)
...
+            }
+
+            foreach ($this->customVarSources as $relation => $label) {
+                if (isset($customVar->$relation)) {
+                    $varRelation = $relation . '.vars.' . $search;
+                    $varLabel = sprintf($label, $name);
+                    // Suggest exact custom variable matches first
+                    if (
+                        ! in_array($varRelation, $exactVarMatches)
+                        && $this->matchSuggestion($varRelation, $varLabel, $searchTerm)
```
nilmerg, line 313: Please restore `filterColumnSuggestions`. Didn't test, but aren't columns not filtered right now?

nilmerg (changes requested):

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
raviks789: As discussed offline, the custom variables need not be filtered again. But `filterColumnSuggestions` should not be restored as we are filtering columns directly in `fetchColumnSuggestions`.

nilmerg (approved):
Good idea to replace the in_array search with a database filter!

Issue #1187, opened 2025-05-14
IcingaDB history can cause system crashes


tbauriedel:
## Describe the bug
While using the `icingadb/history`, the MySQL needs to cache all selected data in tmp-tables.
Based on the amount of data and the executed search filter, a lot of JOINS needs to be performed and saved inside of these tmp tables.

As a result of this, the MySQL tmp-dir grows massive and reaches the disk limit. In some scenarios this causes system outages and kills the monitoring infrastructure

## To Reproduce
1. Prepare a big setup with a lot of data
2. Execute queries (with custom vars for example)
3. watch the behavior of the backend

## Your Environment
Include as many relevant details about the environment you experienced the problem in.

* Icinga DB Web version (System - About):  1.1.3

PR #1219, opened 2025-06-13
Allow disabling the MySQL/MariaDB query optimizer for history queries

lippserd:
Some versions of these RDBMS perform poorly with history queries, particularly when the optimizer changes join order or uses block nested loop joins. Ideally, testing across all RDBMS versions to identify when the optimizer fails and adjusting queries or using optimizer switches would be preferable, but this level of effort is not justified at the moment.

Optimizer is disabled via config:

`/etc/icingaweb2/modules/icingadb/config.ini`:

```
[icingadb]
...
disable_optimizer_for_history_queries=1
```

refs #1187


nilmerg:
I think the config option should be documented.

library/Icingadb/Util/OptimizerHints.php
```diff
@@ -0,0 +1,86 @@
+<?php
```
nilmerg, line 1: missing license header

application/controllers/HistoryController.php
```diff
@@ -87,6 +88,10 @@ public function indexAction()
             Filter::like('service.id', '*')
         ));
 
+        if (OptimizerHints::shouldDisableOptimizerForHistoryQueries()) {
```
nilmerg, line 91: any reason why this is not done by `disableOptimizer`? (It's only a temporary fix anyway, isn't it? So I don't care much about that `disableOptimizer` does not disable anything if disabled. :rofl: )

library/Icingadb/Util/OptimizerHints.php
```diff
@@ -0,0 +1,86 @@
...
+     *
+     * @param Query $q
+     *
+     * @return void
+     */
+    public static function disableOptimizer(Query $q): void
+    {
+        if ($q->getDb()->getAdapter() instanceof Mysql) {
+            // Locates the first string column, prepends the optimizer hint,
+            // and resets columns to ensure it appears first in the SELECT statement:
```
nilmerg, line 60: why the *first* string column? Does the query hint not work if placed before an (aliased) expression?

nilmerg (changes requested):

application/controllers/HistoryController.php
lippserd: Yeah, it's a temporary fix. I'm fine with combining the functions.

library/Icingadb/Util/OptimizerHints.php
lippserd: Sure it works, but it will also increase complexity. My thought was that the hint should work (or rather not add crap) for every query scenario, even if it can't be added at all as with `SELECT 1` (unaliased expression). Considering that all other queries will select real columns, I decided to use this implementation.

nilmerg (approved):

PR #1224, opened 2025-06-16
Change column defination labels of user and usergroup

sukhwinder33445:


library/Icingadb/Model/Behavior/FlattenedObjectVars.php
```diff
@@ -93,6 +93,15 @@ public function rewriteColumnDefinition(ColumnDefinition $def, string $relation)
             $name = substr($name, 0, -3);
         }
 
+        $labelMap = [
+            'user'      => 'contact',
+            'usergroup' => 'contactgroup'
+        ];
```
nilmerg, line 99: Please use a constant for this. (Which you can document properly ;) )

nilmerg (changes requested):

PR #1229, opened 2025-06-18
IcingaHealth: Fix version comparison with dev versions

oxzi:
Icinga DB's development versions mimic git-describe(1), including a commit hash separated by a dash after the semantic version.

The current version comparison using PHP's builtin version_compare() treats this as an "any string"[^0], which is smaller than dev, alpha, beta, and so on. Thus, any git-describe(1) version of Icinga DB 1.4.0 would be considered smaller as 1.4.0.

Since the Icinga DB team uses development versions on a daily basis and sometimes gives patched versions to end users, this faulty CRITICAL would be annoying.

[^0]: https://www.php.net/manual/en/function.version-compare.php

---

FTR, the last time I wrote more than a few lines of PHP was overa decade ago. Thus, please take the diff with a grain of salt. If you dislike this approach, feel free to implement this in a better fashion.


lippserd (changes requested):
I would rather introduce a function called `normalizeVersion` and do the following:
- `ltrim($version, 'v')`
- `explode('-', $version, 2)[0]`
Both calls could get a comment explaining their necessity.

oxzi:
> I want to fix this quickly so that we can release a bugfix today:

Full ack!

> What I don't understand is that there is still a regex.

I just went with a regex because it is easy to read for me. But this is no hill for me to die on. Thus, I am going to alter it as you have asked.

> Edit: We could also add some tests for this function.

Full ack, again. I wanted to do this in the first place, but have not found any good docs how to set this up locally quickly - currently, I am just using vim for this.

lippserd (approved):

PR #1231, opened 2025-06-19
Fix the required version string

### LINKED ISSUE #1230: IcingaHealth: Icinga DB Version Check Is Broken On Debian Package Build
## Describe the bug
After upgrading Icinga 2, Icinga DB, Icinga DB Web and so far on a Debian 12 based on the official Icinga repositories, the newly introduced version check states that Icinga DB must be updated.

Icinga Web Health View

After making sure that this was no caching issue, I took a closer look. The Icinga DB version says `v1.4.0` - with a leading `v`.

Seems like the package build pipeline sets the git tag, which contains this leading `v`. Compare this with the screenshot I shared earlier, <https://github.com/Icinga/icingadb-web/pull/1228/files#r2154091726>.

The usage of PHP's `version_compare` thus fails.

```
php > var_dump(version_compare('v1.4.0', '1.4.0', '<'));
bool(true)
```

As a workaround, I monkey-patched the following.

```diff
--- library/Icingadb/ProvidedHook/IcingaHealth.php.bkp_20250618 2025-06-18 20:36:59.091597024 +0000
+++ library/Icingadb/ProvidedHook/IcingaHealth.php      2025-06-18 20:37:11.495790108 +0000
@@ -14,7 +14,7 @@
 {
     use Database;
 
-    public const REQUIRED_ICINGADB_VERSION = '1.4.0';
+    public const REQUIRED_ICINGADB_VERSION = 'v1.4.0';
 
     /** @var Instance */
     protected $instance;
```

## To Reproduce
Install Icinga DB and Icinga DB Web on a Debian 12 from the official packages.

## Your Environment

* Icinga DB Web version (System - About): 1.2.0
* Web browser: Firefox
* Icinga 2 version (`icinga2 --version`): 2.15.0-1
* Icinga DB version (`icingadb --version`): 1.4.0
* PHP version used (`php --version`): 8.2.28
* Server operating system and version: Debian GNU/Linux 12 (bookworm)

RincewindsHat:
The required version of _Icinga DB_ in _Icinga DB Web_ was given as `1.4.0`, but _Icinga DB_ seems to write `v1.4.0` to the Database.
This commit changes the string in _Icinga DB Web_ and adds the missing `v`.

This should fix #1230.
Alternatively one could change that in _Icinga DB_ I guess.

### DIFF
--- a/library/Icingadb/ProvidedHook/IcingaHealth.php
+++ b/library/Icingadb/ProvidedHook/IcingaHealth.php
@@ -14,7 +14,7 @@ class IcingaHealth extends HealthHook
 {
     use Database;
 
-    public const REQUIRED_ICINGADB_VERSION = '1.4.0';
+    public const REQUIRED_ICINGADB_VERSION = 'v1.4.0';
 
     /** @var Instance */
     protected $instance;


oxzi (changes requested):
First, thanks for creating this PR!

While this is the exact quick fix I posted in #1230, this will break other setups. The issue's core lies in the fact that the packages are built with the git tag, containing the `v`. However, if one simply checks out the `v1.4.0` git tag on the Icinga DB repo and builds the software for themself, the current version would even work and this PR would break the behavior. I am going to alter #1229 in this regard.

PR #1243, opened 2025-07-07
db: Use an alias for `*_customvar` through tables

### LINKED ISSUE #1162: Object's customvars property causes error when Role filters objects on custom var value
## Describe the bug
Came across this developing custom modules that utilize the Icingadb Web classes.  For the ..Model\Service or ..Model\Host objects, if I attempt to retrieve the customvars for that object an error will be thrown IF the user is in a role that restricts objects based on a custom variable.  

## To Reproduce
Setup a role with a restriction like:
icingadb/​filter/​objects: host.vars.FOO=bar

Implement in a module something (like a Service Action hook), that accesses custom vars

$vars = $service->customvars.

Login with a user with the role restriction and attempt to use the module.

PHP throws an error like:
PDOException in /usr/share/icinga-php/ipl/vendor/ipl/sql/src/Connection.php .... SQLSTATE[42000]: Syntax error or access violation: 1066 Not unique table/alias: 'sub_customvar_flat_host_customvar'

## Expected behavior
return a PHP array with the customvars

## Your Environment

* Icinga DB Web version (System - About): 1.1.3

## Additional context
Can workaround by using $service->vars  or assigning the vars with 
(new CustomvarFlat())->unFlattenVars($service->customvar_flat);

(to get the nested array values)

Seems like, somewhere, when the query for icingadb is pieced together that table alias is used more than once.

nilmerg:
fixes #1162


sukhwinder33445 (approved):
Works fine.

sukhwinder33445 (approved):

PR #1244, opened 2025-07-08
Move the `HostController::joinFix()` method to `DependencyNode`

sukhwinder33445:


sukhwinder33445:
> Is it possible to use the new method in `ObjectDetail` as well? Then please do so.

The `ObjectDetail::createAffectedObjects()` only fetch failed edges.

library/Icingadb/Model/DependencyNode.php
```diff
@@ -130,4 +133,51 @@ public function createRelations(Relations $relations): void
...
+     *
+     * @internal Bug: This fix is required for host based queries. Otherwise, ipl-orm makes two subqueries
+     * from the filter, which leads to incorrect results.
+     *
+     * @todo This is a workaround, remove it once https://github.com/Icinga/ipl-orm/issues/119 is fixed
+     *
+     * @param string $hostId Host id to fetch parents or children for
+     * @param Connection $db The database connection
+     * @param bool $fetchParents Fetch parents if true, children otherwise
+     */
```
nilmerg, line 148: Missing return annotation

nilmerg (changes requested):

nilmerg (approved):

PR #1246, opened 2025-07-09
Comment|DowntimeDetail: Use correct filter for removal

### LINKED ISSUE #1245: Unable to delete downtimes despite existing permissions
## Describe the bug
The user receives the message `No authorization for icingadb/command/downtime/delete` if he tries to delete a downtime although he has the `icingadb/command/downtime/*` permission.

## To Reproduce
1. User creates downtime
2. Downtime was created successfully
3. User tries to remove the downtime but Icinga DB Web refuses to mentioned error.

## Expected behavior
Downtime gets removed.

## Screenshots

Image

Permission check via audit:

Image

## Your Environment

* Icinga DB Web version (System - About): 1.2.1
* Web browser: Google Chrome 138.0.7204.93
* Icinga 2 version (`icinga2 --version`): 2.15.0
* Icinga DB version (`icingadb --version`): 1.4.0
* PHP version used (`php --version`): 8.3
* Server operating system and version: Ubuntu 24.04

## Additional context
n/a

nilmerg:
Since #1060, comments/delete and downtimes/delete try to use `ObjectAuthorization`'s cache properly and override `CommandAction`'s `isGrantedOnType()`. Though, the filter is applied to the host and service model as a result, and not to the downtime model. This way, downtime filters MUST be absolute, just like filters provided by the search bar. Otherwise `name=downtime-name` will be translated to e.g. `host.name=downtime-name` which obviously cannot match.

fixes #1245

### DIFF
--- a/library/Icingadb/Widget/Detail/CommentDetail.php
+++ b/library/Icingadb/Widget/Detail/CommentDetail.php
@@ -110,7 +110,7 @@ class CommentDetail extends BaseHtmlElement
         }
 
         $action = Links::commentsDelete();
-        $action->setFilter(Filter::equal('name', $this->comment->name));
+        $action->setFilter(Filter::equal('comment.name', $this->comment->name));
 
         return (new DeleteCommentForm())
             ->setObjects([$this->comment])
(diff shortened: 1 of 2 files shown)


nilmerg:
@sukhwinder33445 Note that you need to be restricted to reproduce the issue.

PR #1253, opened 2025-07-11
History: Fix incorrect relation name of target `NotificationHistory`

sukhwinder33445:
Screenshot 2025-07-11 at 13 09 26


sukhwinder33445:
This is not the right fix. There is a new issue https://github.com/Icinga/icingadb-web/issues/1254 for this.

PR #1264, opened 2025-08-05
Break the group names and display names

jrauh01:
Break the names and display names of host and service groups instead of cutting them off.


jrauh01:
There will be a minimal view in the future that fixes this problem.

PR #1277, opened 2025-09-08
Add 'scroll to top' button for `History` list

### LINKED ISSUE #760: Add a `scroll to top` button for `History` list
In the `History` page, it would be nice to have a `scroll to top` button to go back to top of the list.

jrauh01:
The button links to a span with the id `#page-1` at the top of the list. Additionally every 'Load More' click the button's href gets updated to keep the already opened pages open when scrolling to top via the button.

Resolve #760

### DIFF
--- a/public/css/common.less
+++ b/public/css/common.less
@@ -321,3 +321,9 @@ form[name="form_confirm_removal"] {
   background: @control-hover-bg;
   text-decoration: none;
 }
+
+#page-1 {
+  display: block;
+  width: 1px;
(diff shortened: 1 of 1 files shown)


jrauh01:
As discussed with @flourish86 and @lippserd I marked this PR as draft. Before continuing here, we agreed to consider making the `Home/Pos 1` and `End` keys work properly instead of adding a `scroll-to-top` button. That would make the button unnecessary. Regarding container focus, we could handle it the same way as with the item list arrow key navigation.

PR #1298, opened 2025-11-10
Hide list selection-count and action list hint in minimal layout

sukhwinder33445:
ref https://github.com/Icinga/icingaweb2/issues/5386


flourish86 (dismissed):
lgtm

nilmerg (changes requested):

PR #1300, opened 2025-11-12
ContinueWith: Disable button if the query with filter returns no results

### LINKED ISSUE #1299: Filter bulk action `Continue with` shows an empty detail page
Image

## To Reproduce
1. Add a non-matching search bar filter so that the Continue with button becomes enabled.
2. Ensure the filtered list is empty.
3. Click the Continue with button to open the detail view.
4. All action buttons in the detail view now throw an error.

sukhwinder33445:
fixes #1299 

requires https://github.com/Icinga/ipl-web/pull/334

### DIFF
--- a/library/Icingadb/Common/SearchControls.php
+++ b/library/Icingadb/Common/SearchControls.php
@@ -55,7 +55,7 @@ trait SearchControls
      *
      * @return ContinueWith
      */
-    public function createContinueWith(Url $detailsUrl, SearchBar $searchBar, bool $hasResults): ContinueWith
+    public function createContinueWith(Url $detailsUrl, SearchBar $searchBar, bool $hasResults = true): ContinueWith
     {
         if ($hasResults) {
             $continueWith = ContinueWith::create(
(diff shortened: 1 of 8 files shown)


library/Icingadb/Common/SearchControls.php
```diff
@@ -51,14 +51,24 @@ public function createSearchBar(Query $query, ...$params): SearchBar
      *
      * @param Url $detailsUrl
      * @param SearchBar $searchBar
+     * @param bool $hasResults Whether the current query has results
      *
      * @return ContinueWith
      */
-    public function createContinueWith(Url $detailsUrl, SearchBar $searchBar): ContinueWith
+    public function createContinueWith(Url $detailsUrl, SearchBar $searchBar, bool $hasResults): ContinueWith
```
nilmerg: Default the new parameter to `true` here. This trait is used by the graphite module and dependencies, but while they don't call `createContinueWith`, it's a perfect example why backwards compatibility matters: Because they *could have* called it.

nilmerg (changes requested):

nilmerg (approved):

Issue #1011, opened 2024-04-25
Feature request : print only requested columns for CSV/PDF/JSON Export


gbin2265:
## Proposal for export's CSV/PDF/JSON

If you now request a CSV export, it provides all the columns that exist.
Wouldn't it be possible to include an additional parameter where you specify which columns you want to see only in the CSV/PDF/EXPORT export?

## Describe the solution you'd like

add parameter like **printcolums**=...

You can then use this parameter as a filter when printing the columns.

### example
**printcolumns**=host.id,host.display_name,host.vars.os,host.address,host.name,host.vars.local_var

### Result
id,display_name,address,name,vars.os,vars.local_var
"1370166c5ea3a961e81ecf3ace578e19dac2184e","host.com","1.1.1.1","obj_host","rhel9","city1"

nilmerg:
Hi,

for PDF this is already possible by using `?columns=` in the hosts and services list. This just needs to be supported for the CSV and JSON export.

gbin2265:
Hi,

Added value was from printcolumns that you can give additional columns like the IDs in the export but not on the screen.

But tell you that the columns option is the fastest/easiest way, I will be very happy already.

Nathaniel-Donahue:
ref/NC/816326

nilmerg:
> This just needs to be supported for the CSV and JSON export.

Actually, I'm not sure what I was talking about here. Exporting e.g. `hosts?columns=…` to CSV/JSON works fine, it just includes not only the declared columns but still all the default columns.

If I export `hosts?columns=hostgroup.name` to JSON, there is a `hostgroup` key with an object `{ "name": … }` in each result.

gbin2265:
If you do an export now, you get all columns in the CSV/JSON.
What I asked was whether it would be possible that, when using the column option, only the specified columns are included in the CSV/JSON export.
Similar to how it already works for the PDF.



what you wrote :
for PDF this is already possible by using ?columns= in the hosts and services list. This just needs to be supported for the CSV and JSON export.

nilmerg:
> only the specified columns are included in the CSV/JSON export. Similar to how it already works for the PDF.

Of course, I just wanted to mention that the basic functionality of this parameter already works.

PR #1301, opened 2025-11-13
Only include requested columns in CSV/JSON exports when using `?columns=…`

BastianLedererIcinga:
resolves: #1011

### DIFF
--- a/library/Icingadb/Web/Controller.php
+++ b/library/Icingadb/Web/Controller.php
@@ -22,8 +22,6 @@ use Icinga\Module\Icingadb\Common\Database;
 use Icinga\Module\Icingadb\Common\SearchControls;
 use Icinga\Module\Icingadb\Data\CsvResultSet;
 use Icinga\Module\Icingadb\Data\JsonResultSet;
-use Icinga\Module\Icingadb\Model\Host;
-use Icinga\Module\Icingadb\Model\Service;
 use Icinga\Module\Icingadb\Web\Control\GridViewModeSwitcher;
 use Icinga\Module\Icingadb\Web\Control\ViewModeSwitcher;
 use Icinga\Module\Icingadb\Widget\ItemTable\StateItemTable;


library/Icingadb/Data/JsonResultSetUtils.php
```diff
@@ -51,7 +51,10 @@ protected function createObject(Model $model): array
         $keysAndValues = [];
         foreach ($model as $key => $value) {
             if ($value instanceof Model) {
-                $keysAndValues[$key] = $this->createObject($value);
+                $object = $this->createObject($value);
+                if ($object !== []) {
```
nilmerg:
```suggestion
                if (! empty($object)) {
```

library/Icingadb/Redis/VolatileStateResults.php
```diff
@@ -36,11 +36,17 @@ class VolatileStateResults extends ResultSet
...
+    protected ?array $columns = null;
+
     public static function fromQuery(Query $query)
     {
         $self = parent::fromQuery($query);
         $self->resolver = $query->getResolver();
         $self->redisUnavailable = Backend::getRedis()->isUnavailable();
+        if (! empty($query->getColumns())) {
+            $self->columns = $query->getColumns();
+        }
```
nilmerg: I suppose it's fine to just assign the property here. Later on you can still differentiate, as whether null or an empty array does not make a difference, does it?

nilmerg (changes requested):

BastianLedererIcinga:
I made the change to `CsvResultSetUtils::stream`, because it previously added the CSV-header not only at the beginning of the file, but repeated printing it after every 1000 lines.

doc/05-Upgrading.md
```diff
@@ -11,9 +11,9 @@ If you are upgrading across multiple versions, make sure to follow the steps for
 
 **Breaking Changes**
 
-The JSON/CSV-export for Hosts and Services now supports filtering for specific columns with the `?columns` parameter.
-The exported file will only contain the listed columns and the id of the Host/Service, if the parameter is not set, all columns will be included.
-If a relation is entirely empty and would result in an empty JSON-object, the JSON-export will not create an object for it at all.
+* The JSON/CSV-export for Hosts and Services now supports filtering for specific columns with the `?columns` parameter.
+  * The exported file will only contain the listed columns of the Host/Service, if the parameter is not set, all columns will be included.
+  * If a relation is entirely empty and would result in an empty JSON-object, the JSON-export will not create an object for it at all.
```
nilmerg: That should be its own first level list item. It's not dependent on `?columns` after all. Also, please give an example, I'm not sure everyone is familiar with the term *relation*.

library/Icingadb/Web/Controller.php
```diff
@@ -22,6 +22,8 @@
 use Icinga\Module\Icingadb\Common\SearchControls;
 use Icinga\Module\Icingadb\Data\CsvResultSet;
 use Icinga\Module\Icingadb\Data\JsonResultSet;
+use Icinga\Module\Icingadb\Model\Host;
+use Icinga\Module\Icingadb\Model\Service;
```
nilmerg: :sweat_smile:

nilmerg (changes requested):

nilmerg (approved):

PR #1322, opened 2026-01-19
Allow switching between relative and absolute timestamps

### LINKED ISSUE #1065: Feature Request: Please give us back the timestamps in history view
In the monitoring module webinterface (the old web interface) you had timestamps in the history view.
In the icingadb webinterface, It's really hard to debug thinks because of the relative time values:
grafik

Please return to the good old timestamps:
thumbnail_image001

I have some ideas:
idea

BastianLedererIcinga:
resolves #1065

For the views:
* Host History
* Service History
* Notifications
* Event Overview

A new control is added that allows to switch between relative and absolute timestamps.
Screenshot from 2026-01-16 10-08-57
The mode can also be switched by clicking one of timestamps.
Whenever the mode is changed it is added to the url and saved in the user's session and preferences.

Absolute timestamps now always show the time of day, to display the dates a seperate line is added.
Screenshot from 2026-01-16 10-31-58
To ensure this line is also added when loading more entries the timestamp of the latest entry is added to the `load-more` url.

The relative timestamps rely on the new implementation in https://github.com/Icinga/ipl-web/pull/350.


nilmerg (changes requested):
Please also prevent text selection in case the time tags are interactive, that's common practice for such widgets.

And I had a bug I did't track down yet: After some time or interactions, switching between relative/absolute didn't work anymore. The toggle updated accordingly, but neither pressing the toggle nor the time tags did update the time. A refresh fixes it. But I didn't look at the JS in detail yet, maybe I can easily spot the problem once I do this, so please focus on the other comments first.

application/controllers/HistoryController.php
```diff
@@ -98,10 +103,17 @@ public function indexAction()
             ->onlyWith($preserveParams)
             ->setFilter($filter);
 
-        $historyList = (new LoadMoreObjectList($history->execute()))
+        $historyList = (new LoadMoreObjectList(
+            $history->execute(),
+            $previousTimestamp,
+            $this->useRelativeTimestamps,
```
nilmerg: Please let the toggle provide this. Though, my main concern with this is actually the controller's property . We usually avoid allowing a control (i.e. url parameter) to have an effect on the state of a controller. In addition, I don't see a reason in your case for doing so either.

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,36 @@
...
+    public function __construct(bool $useRelativeTimestamps = false)
+    {
+        $this->useRelativeTimestamps = $useRelativeTimestamps;
+    }
+
+    protected function assemble()
+    {
+        $this->addElement('checkbox', 'timestamp-toggle', [
+            'class'             => 'timestamp-toggle autosubmit',
+            'id'                => Icinga::app()->getRequest()->protectId('timestamp-toggle'),
```
nilmerg: Is there a reason why you need the id here?

Though, accessing a singleton here is why I'm asking actually. This should be avoided if possible, or rather not be required to begin with.

library/Icingadb/Web/Controller.php
```diff
@@ -268,6 +272,71 @@ function (ViewModeSwitcher $viewModeSwitcher) use (
         return $viewModeSwitcher;
     }
 
+    /**
+     * Create a control to switch between relative and absolute timestamps
+     *
+     * @param string $path The path that is used as key to load the current preference e.g. icingadb/history
```
nilmerg: This can be derived from the request.

nilmerg (changes requested):
again, php only, css + js after my break :)

nilmerg (changes requested):

public/js/timestamp-mode.js
```diff
@@ -0,0 +1,87 @@
...
+         * @param event The change event of the timestamp-toggle
+         */
+        onTimestampModeToggle(event)
+        {
+            const preference = (event.target.checked) ? 'relative' : 'absolute';
+            const container = event.target.closest('.container');
+            const url = window.icinga.utils.addUrlParams($(container).data('icingaUrl'), {timestamps: preference});
+            $(container).data('icingaUrl', url);
+
+            container.querySelectorAll('.load-more .action-link, .refresh-container-control')
```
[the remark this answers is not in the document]
BastianLedererIcinga: Well that's odd, works on my machine...
I tested this again  in Firefox and Chrome both for my recent changes and while checked out to 71fdd1f0158f38ab1e293ed2b7ae8f898c3c82e8, which should be the version you tested. The load more url was updated in all four views, it also seemed to work fine when using two columns.

I have no clue why this would work in my environment but not in yours.

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,50 @@
...
+    /** @var bool Whether relative or absolute timestamps are to be used */
+    protected bool $useRelativeTimestamps;
+
+    protected $defaultAttributes = [
+        'name'    => 'Use relative timestamps',
+        'class'   => ['icinga-form', 'icinga-controls', 'inline']
+    ];
+
+    /**
+     * A toggle that allows to switch between absolute and relative timestamps
```
sukhwinder33445: I would suggest: `Create a toggle to switch between absolute and relative timestamps `

public/js/timestamp-mode.js
```diff
@@ -0,0 +1,87 @@
...
+                .forEach((el) => {
+                    let loadMoreUrl = el.getAttribute('href');
+                    loadMoreUrl = window.icinga.utils.addUrlParams(loadMoreUrl, {timestamps: preference});
+                    el.setAttribute('href', loadMoreUrl);
+                });
+
+            icinga.history.replaceCurrentState();
```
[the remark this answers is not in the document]
sukhwinder33445: This is still not moved.

library/Icingadb/View/EventRenderer.php
```diff
@@ -41,9 +43,15 @@ class EventRenderer implements ItemRenderer
     /** @var NotificationRenderer To render NotificationHistory event */
     protected $notificationRenderer;
 
-    public function __construct()
+    protected bool $useRelativeTimestamps;
+
+    protected bool $interactiveTimestamps;
+
+    public function __construct(bool $useRelativeTimestamps = false, bool $interactiveTimestamps = false)
```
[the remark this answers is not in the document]
sukhwinder33445: As discussed offline with @nilmerg, please add a proper php-docblock with the subject line.

library/Icingadb/View/NotificationRenderer.php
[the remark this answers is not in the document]
sukhwinder33445: Please add proper DocBlock for constructor.

library/Icingadb/Widget/ItemList/HistoryObjectList.php
```diff
@@ -0,0 +1,94 @@
...
+ */
+class HistoryObjectList extends ObjectList
+{
+    use LoadMore;
+
+    /** @var ?int The timestamp of the previous element in the list */
+    protected ?int $previousTimeStamp = null;
+
+    /**
+     * A list of History or NotificationHistory objects with a Load more link and add a separator between days.
```
sukhwinder33445: Create ...

library/Icingadb/Widget/ItemList/HistoryObjectList.php
```diff
@@ -0,0 +1,94 @@
...
+                    new Attributes(['class' => ['day-separator']]),
+                    new Text($formatter->format($timestamp))
+                ));
+            }
+
+            $this->previousTimeStamp = $timestamp;
+        });
+
+        $this->on(self::ON_ASSEMBLED, fn () => $this->loadMoreUrl->setParam('last-entry', $this->previousTimeStamp));
+        parent::init();
```
sukhwinder33445: Is there a reason to call the `parent::init()` at the end? If not, please move it to the top.

library/Icingadb/Widget/ItemList/HistoryObjectList.php
```diff
@@ -0,0 +1,94 @@
...
+
+    protected function init(): void
+    {
+        $formatter = new IntlDateFormatter(
+            Locale::getDefault(),
+            IntlDateFormatter::MEDIUM,
+            IntlDateFormatter::NONE
+        );
+
+        $this->on(self::BEFORE_ITEM_ADD, function ($item, $data) use ($formatter) {
```
sukhwinder33445: Please use `ItemList::BEFORE_ITEM_ADD` instead.

library/Icingadb/Widget/ItemList/HistoryObjectList.php
```diff
@@ -0,0 +1,94 @@
...
+                    'li',
+                    new Attributes(['class' => ['day-separator']]),
+                    new Text($formatter->format($timestamp))
+                ));
+            }
+
+            $this->previousTimeStamp = $timestamp;
+        });
+
+        $this->on(self::ON_ASSEMBLED, fn () => $this->loadMoreUrl->setParam('last-entry', $this->previousTimeStamp));
```
sukhwinder33445: Please use `HtmlDocument::ON_ASSEMBLED`.

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,50 @@
...
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Module\Icingadb\Web\Control;
+
+use ipl\Web\Compat\CompatForm;
+
+class TimestampToggle extends CompatForm
+{
+    /** @var string Default timestamp mode param */
+    public const string DEFAULT_TIMESTAMP_MODE_PARAM = 'timestamps';
```
sukhwinder33445: Typed class constants are only allowed since PHP 8.3, please remove.

public/js/timestamp-mode.js
```diff
@@ -0,0 +1,107 @@
...
+                container.querySelectorAll('.content .interactive-time').forEach(el => {
+                    el.removeAttribute('data-relative-time');
+                    el.textContent = this.timeFormatter.format(new Date(el.getAttribute('datetime')));
+                });
+            }
+        }
+    }
+
+    Icinga.Behaviors.TimestampToggle = TimestampToggle;
+}(Icinga));
```
sukhwinder33445: Please add a new line at the end of the file.

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,50 @@
...
+    /** @var bool Whether relative or absolute timestamps are to be used */
+    protected bool $useRelativeTimestamps;
+
+    protected $defaultAttributes = [
+        'name'    => 'Use relative timestamps',
```
sukhwinder33445: The attribute `name` is not used anywhere and can be removed.

sukhwinder33445 (commented):
I have noticed the following issues:
- The searchbar is not working properly.
- The refresh button does not preserve the `load more` params.

This is just a quick overview. I’ll take a closer look at it tomorrow.

library/Icingadb/Web/Controller.php
```diff
@@ -270,6 +271,63 @@ function (ViewModeSwitcher $viewModeSwitcher) use (
...
+                    if (($preferences = $user->getAdditional('icingadb.timestamps')) === null) {
+                        if (
+                            array_key_exists('icingadb', $storedPreferences)
+                            && array_key_exists('timestamps', $storedPreferences['icingadb'])
+                        ) {
+                            $preferences = Json::decode($storedPreferences['icingadb']['timestamps'], true);
+                        } else {
+                            $preferences = [];
+                        }
+                    }
```
sukhwinder33445: This can be simplified as follows:
```suggestion
$preferences = $user->getAdditional('icingadb.timestamps') 
    ?? Json::decode($storedPreferences['icingadb']['timestamps'] ?? '[]', true);
```

library/Icingadb/Web/Controller.php
```diff
@@ -270,6 +271,63 @@ function (ViewModeSwitcher $viewModeSwitcher) use (
...
+                    $this->_helper->viewRenderer->setNoRender(true);
+                    $this->_helper->layout->disableLayout();
+                    $this->ignoreXhrBody();
+
+                    $value = $form->getValue('timestamp-toggle') === 'y' ? 'relative' : 'absolute';
+
+                    $preferencesStore = PreferencesStore::create(new ConfigObject([
+                        'resource'  => Config::app()->get('global', 'config_resource')
+                    ]), $user);
+                    $storedPreferences = $preferencesStore->load();
```
sukhwinder33445: Please wrap this in try-catch just like in `createViewModeSwitcher()`. Calling `PreferencesStore::create()` and `$preferencesStore->load()` may throw an exception.

public/js/timestamp-mode.js
```diff
@@ -0,0 +1,107 @@
...
+        {
+            const preference = relative ? 'relative' : 'absolute';
+            const url = this.icinga.utils.addUrlParams($(container).data('icingaUrl'), {timestamps: preference});
+            $(container).data('icingaUrl', url);
+
+            [
+                ['.load-more .action-link, .refresh-container-control, .primary-nav li.active a', 'href'],
+                ['.search-editor-opener', 'data-search-editor-url'],
+                ['.search-bar', 'action'],
+            ].forEach(([selector, attr]) => {
```
sukhwinder33445: I would suggest using the following for better readability of selector and attr:
```suggestion
Object.entries({
                '.load-more .action-link, .refresh-container-control, .primary-nav li.active a' : 'href',
                '.search-editor-opener' : 'data-search-editor-url',
                '.search-bar' : 'action'
            }).forEach(([selector, attr]) => { ...
```

public/js/timestamp-mode.js
```diff
@@ -0,0 +1,107 @@
...
+        {
+            event.stopPropagation();
+
+            const container = event.target.closest('.container');
+            const toggle = container.querySelector('.timestamp-toggle');
+
+            if (toggle) {
+                toggle.click();
+            } else {
+                const isRelative = event.target.hasAttribute('data-relative-time');
```
sukhwinder33445: Please avoid variables that are only used once.

public/js/timestamp-mode.js
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        /**
+         * Handle changed timestamp mode
+         *
+         * @param event The change event of the timestamp-toggle
+         */
+        onTimestampModeToggle(event)
+        {
+            const container = event.target.closest('.container');
```
sukhwinder33445: Same here.

library/Icingadb/Web/Controller.php
```diff
@@ -270,6 +271,63 @@ function (ViewModeSwitcher $viewModeSwitcher) use (
...
+        if ($timestampMode === null) {
+            if (($preferences = $user->getAdditional('icingadb.timestamps')) === null) {
+                $preferences = Json::decode(
+                    $user->getPreferences()->getValue('icingadb', 'timestamps', '[]'),
+                    true
+                );
+                $user->setAdditional('icingadb.timestamps', $preferences);
+            }
+
+            $timestampMode = array_key_exists($path, $preferences) ? $preferences[$path] : 'absolute';
```
sukhwinder33445:
```suggestion
$timestampMode =  $preferences[$path] ?? 'absolute';
```

library/Icingadb/Web/Controller.php
```diff
@@ -270,6 +271,63 @@ function (ViewModeSwitcher $viewModeSwitcher) use (
...
+     *
+     * @return TimestampToggle
+     */
+    public function createTimestampControl()
+    {
+        $path = $this->getRequest()->getUrl()->getPath();
+        $timestampMode = $this->params->shift('timestamps');
+        $user = $this->Auth()->getUser();
+        if ($timestampMode === null) {
+            if (($preferences = $user->getAdditional('icingadb.timestamps')) === null) {
```
sukhwinder33445: Please don't initialize variables in the if condition.

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,50 @@
...
+
+namespace Icinga\Module\Icingadb\Web\Control;
+
+use ipl\Web\Compat\CompatForm;
+
+class TimestampToggle extends CompatForm
+{
+    /** @var string Default timestamp mode param */
+    public const DEFAULT_TIMESTAMP_MODE_PARAM = 'timestamps';
+    /** @var bool Whether relative or absolute timestamps are to be used */
```
sukhwinder33445, line 20: Please add a newline after the property defination.

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @param bool $useRelativeTimestamps Whether to use relative timestamps
+     */
+    public function __construct(bool $useRelativeTimestamps = false)
+    {
+        $this->useRelativeTimestamps = $useRelativeTimestamps;
+    }
+
+    /**
+     * Get whether relative or absolute timestamps are to be used
```
sukhwinder33445: **Get whether relative timestamps should be used**
This method does not return `true` for a relative timestamp.

sukhwinder33445 (changes requested):

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+namespace Icinga\Module\Icingadb\Web\Control;
+
+use ipl\Web\Compat\CompatForm;
+
+class TimestampToggle extends CompatForm
```
sukhwinder33445: This is a new form, so we want to use the new decorator implementation. To impelement this, please do the following:
1. Extend the `ipl\Html\Form` instead of `CompatForm`.
2. In constructor, call the following method to load the decorators of `ipl\web`:
```php
$this->addElementDecoratorLoaderPaths([['ipl\\Web\\Compat\\FormDecorator', 'Decorator']]);
```  
4. Apply the decorators to the element in `assemble()`:
```php
$this->addElement('checkbox', 'timestamp-toggle', [
            'class'             => ['timestamp-toggle', 'autosubmit'],
            'label'             => $this->translate('Use relative timestamps'),
            'value'             => $this->useRelativeTimestamps,
            'decorators'        => [
                'Label',
                'LabelGroup' => [
                    'name' => 'HtmlTag',
                    'options' => [
                        'tag' => 'div',
                        'class' => 'control-label-group',
                    ]
                ],
                'Checkbox',
                'RenderElement',
                'ControlGroup' => [
                    'name' => 'HtmlTag',
                    'options' => [
                        'tag' => 'div',
                        'class' => 'control-group'
                    ]
                ]
            ]
        ]);
```

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,50 @@
...
+     * @return bool
+     */
+    public function getUseRelativeTimestamps(): bool
+    {
+        return $this->useRelativeTimestamps;
+    }
+
+    protected function assemble()
+    {
+        $this->addElement('checkbox', 'timestamp-toggle', [
```
sukhwinder33445, line 65: The search bar is currently not working because its submit triggers the submit of this form, which ignores the response. To fix this, create a UID element using the `FormUid` trait. This trait now requires the form's `name` attribute, so please add it back if you removed it due to my previous comment. Then add the element as follows:
```php
// assemble():
$this->addElement($this->createUidElement());
```

sukhwinder33445:
> * The refresh button does not preserve the `load more` params.

I checked again and unfortunately this is a bug in the main branch, so it must be fixed in a separate PR.

library/Icingadb/Web/Control/TimestampToggle.php
sukhwinder33445: It is required now.

run.php
```diff
@@ -30,4 +30,16 @@
...
+        new Zend_Controller_Router_Route(
+            'icingadb/history/timestamp-preference',
+            [
+                'module'     => 'icingadb',
+                'controller' => 'history',
+                'action'     => 'timeStamp'
+            ]
+        )
+    );
 }
```
sukhwinder33445, line 33: This action no longer exists, please remove it.

public/js/timestamp-toggle.js
```diff
@@ -0,0 +1,107 @@
+// SPDX-FileCopyrightText: 2026 Icinga GmbH <https://icinga.com>
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+(function (Icinga) {
```
sukhwinder33445, line 4: I would suggest naming this file `timestamp-toggle.js` so that it matches the PHP implementation.

public/js/timestamp-mode.js
```diff
@@ -0,0 +1,107 @@
...
+                { dateStyle: 'medium', timeStyle: 'medium', timeZone: icinga.config.timezone }
+            );
+            this.timeFormatter = new Intl.DateTimeFormat(
+                icinga.config.locale,
+                { timeStyle: 'medium' }
+            );
+        }
+
+        /**
+         * Stop event propagation, so that the list-item underneath is not clicked,
```
sukhwinder33445: That should be the description, as this is not the method’s primary function.

Something like:
```
/**
*  Toggle the timestamp mode of the container
*
*  This method Stop event propagation, so that the list-item underneath is not clicked.
* 
* @param event The click event of the timestamp
*/
```

library/Icingadb/Widget/ItemList/HistoryObjectList.php
```diff
@@ -0,0 +1,95 @@
...
+use ipl\Orm\ResultSet;
+use ipl\Web\Widget\ItemList;
+use Locale;
+
+/**
+ * HistoryObjectList
+ *
+ * Create a list of History or NotificationHistory objects with a Load more link
+ * and add a separator when a new day begins.
+ *
```
sukhwinder33445: Please remove this, as the constructor has the same description.

library/Icingadb/View/NotificationRenderer.php
```diff
@@ -154,7 +165,27 @@ public function assembleCaption($item, HtmlDocument $caption, string $layout): v
 
     public function assembleExtendedInfo($item, HtmlDocument $info, string $layout): void
     {
-        $info->addHtml(new TimeAgo($item->send_time->getTimestamp()));
+        if (
+            ($this->useRelativeTimestamps && time() - $item->send_time->getTimestamp() < 3600)
+            || $layout === 'header'
```
sukhwinder33445: Please swap the conditions as this condition execution costs less than the previous one.

library/Icingadb/View/EventRenderer.php
```diff
@@ -414,7 +422,27 @@ public function assembleCaption($item, HtmlDocument $caption, string $layout): v
 
     public function assembleExtendedInfo($item, HtmlDocument $info, string $layout): void
     {
-        $info->addHtml(new TimeAgo($item->event_time->getTimestamp()));
+        if (
+            ($this->useRelativeTimestamps && time() - $item->event_time->getTimestamp() < 3600)
+            || $layout === 'header'
```
sukhwinder33445: Please swap the conditions as this condition execution costs less than the previous one.

sukhwinder33445 (changes requested):

library/Icingadb/Widget/ItemList/HistoryObjectList.php
```diff
@@ -0,0 +1,96 @@
...
+ * @template Item of NotificationHistory|History
+ *
+ * @extends ObjectList<Item>
+ */
+class HistoryObjectList extends ObjectList
+{
+    use LoadMore;
+
+    /** @var ?int The timestamp of the previous element in the list */
```
sukhwinder33445: Please describe, why this property is required?
```suggestion
 /** @var ?int The timestamp of the recently added list item, used to detect day changes */
```

library/Icingadb/Widget/ItemList/HistoryObjectList.php
```diff
@@ -0,0 +1,96 @@
...
+    protected ?int $previousTimeStamp = null;
+
+    /**
+     * Create a list of History or NotificationHistory objects with a Load more link and add a separator between days.
+     *
+     * @param ResultSet $data
+     * @param ?int $previousTimeStamp
```
sukhwinder33445: Same here.

public/css/list/item-list.less
```diff
@@ -42,6 +44,21 @@
...
     border-top: none;
   }
+
+  .interactive-time {
+    &:hover {
+      .rounded-corners(1em);
+      background: @interactive-time-hover-bg;
+    }
+
+    user-select: none;
```
sukhwinder33445: Please use the mixin instead:
```suggestion
.user-select(none);
```

public/js/timestamp-toggle.js
```diff
@@ -0,0 +1,106 @@
...
+                '.load-more .action-link, .refresh-container-control, .primary-nav li.active a': 'href',
+                '.search-editor-opener': 'data-search-editor-url',
+                '.search-bar': 'action'
+            }).forEach(([selector, attr]) => {
+                container.querySelectorAll(selector).forEach((el) => {
+                    el.setAttribute(attr, this.icinga.utils.addUrlParams(el.getAttribute(attr), {timestamps: preference}));
+                });
+            });
+
+            this.icinga.history.replaceCurrentState();
```
sukhwinder33445: Please move this immediately after `$(container).data(“icingaUrl”, url);`, as the two calls belong together.
@nilmerg mentioned this [here](https://github.com/Icinga/icingadb-web/pull/1322#discussion_r2940585095).

public/js/timestamp-toggle.js
```diff
@@ -0,0 +1,106 @@
...
+
+            if (relative) {
+                container.querySelectorAll('.content .interactive-time').forEach(el => {
```
sukhwinder33445: `.content` is very general, please use a more specific selector such as `.item-list`

public/js/timestamp-toggle.js
```diff
@@ -0,0 +1,106 @@
...
+
+                    if (diff < 3600) {
+                        const parts = el.getAttribute('data-ago-label').split(/\d+/);
+                        el.textContent = parts[0]
+                            + Math.floor(diff / 60) + parts[1]
+                            + Math.floor(diff % 60) + parts[2];
+                    }
+                });
+            } else {
+                container.querySelectorAll('.content .interactive-time').forEach(el => {
```
sukhwinder33445: Same here.

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,77 @@
...
+        'name'    => 'timestamp-toggle',
+        'class'   => ['icinga-form', 'icinga-controls', 'inline']
+    ];
+
+    /**
+     * Create a toggle to switch between absolute and relative timestamps
+     *
+     * @param bool $useRelativeTimestamps Whether to use relative timestamps
+     */
+    public function __construct(bool $useRelativeTimestamps = false)
```
sukhwinder33445: As this is a toggle, it can have only two states, so `$enabled` defines whether to use relative (`true`) or absolute (`false`) timestamps.
```suggestion
/**
     * Create a toggle to switch between absolute and relative timestamps
     * 
     * When enabled, timestamps are displayed as relative, absolute otherwise
     *
     * @param bool $enabled True to use relative timestamps, false for absolute
     */
     public function __construct(bool $enabled = false)
```

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Get whether relative timestamps should be used
+     *
+     * @return bool
+     */
+    public function getUseRelativeTimestamps(): bool
+    {
+        return $this->useRelativeTimestamps;
+    }
```
sukhwinder33445, line 61:
```suggestion
/**
     * Get whether the toggle is enabled
     * 
     * When enabled, timestamps are displayed as relative, absolute otherwise
     *
     * @return bool
     */
    public function isEnabled(): bool
    {
        return $this->getElement('timestamp-toggle')->isChecked();
    }
```

library/Icingadb/Web/Control/TimestampToggle.php
```diff
@@ -0,0 +1,77 @@
...
+
+    protected function assemble()
+    {
+        $this->addElement('checkbox', 'timestamp-toggle', [
+            'class'             => ['timestamp-toggle', 'autosubmit'],
+            'label'             => $this->translate('Use relative timestamps'),
+            'value'             => $this->useRelativeTimestamps,
```
sukhwinder33445: `$this->enabled`

library/Icingadb/Web/Controller.php
```diff
@@ -270,6 +271,60 @@ function (ViewModeSwitcher $viewModeSwitcher) use (
...
+            ->on(
+                TimestampToggle::ON_SUBMIT,
+                function (TimestampToggle $form) use ($path, $user) {
+                    // The js changes the representation in the frontend, this request only saves the preference,
+                    // so we can skip all unnecessary work and signal the frontend not to replace the DOM
+                    $this->_helper->viewRenderer->setNoRender(true);
+                    $this->_helper->layout->disableLayout();
+                    $this->ignoreXhrBody();
+
+                    $value = $form->getValue('timestamp-toggle') === 'y' ? 'relative' : 'absolute';
```
sukhwinder33445: Please omit the variable and use the new method `isEnabled()`.

library/Icingadb/Web/Controller.php
```diff
@@ -270,6 +271,60 @@ function (ViewModeSwitcher $viewModeSwitcher) use (
...
+                $user->setAdditional('icingadb.timestamps', $preferences);
+            }
+
+            $timestampMode = $preferences[$path] ?? 'absolute';
+        }
+
+        return (new TimestampToggle($timestampMode === 'relative'))
+            ->on(
+                TimestampToggle::ON_SUBMIT,
+                function (TimestampToggle $form) use ($path, $user) {
```
sukhwinder33445: You can use the variable `$toggle` instead of `$form` so that calling `isEnabled()` makes more sense.

sukhwinder33445 (changes requested):

sukhwinder33445 (dismissed):
🥳  Looks good to me.

library/Icingadb/View/EventRenderer.php
```diff
@@ -414,7 +426,27 @@ public function assembleCaption($item, HtmlDocument $caption, string $layout): v
 
     public function assembleExtendedInfo($item, HtmlDocument $info, string $layout): void
     {
-        $info->addHtml(new TimeAgo($item->event_time->getTimestamp()));
+        if (
+            $layout === 'header'
+            || ($this->useRelativeTimestamps && time() - $item->event_time->getTimestamp() < 3600)
```
nilmerg: Why does this show absolute times once an hour has passed? This renders most of the cases `TimeAgo` handles useless.

library/Icingadb/View/NotificationRenderer.php
```diff
@@ -154,7 +169,27 @@ public function assembleCaption($item, HtmlDocument $caption, string $layout): v
 
     public function assembleExtendedInfo($item, HtmlDocument $info, string $layout): void
     {
-        $info->addHtml(new TimeAgo($item->send_time->getTimestamp()));
+        if (
+            $layout === 'header'
+            || ($this->useRelativeTimestamps && time() - $item->send_time->getTimestamp() < 3600)
```
nilmerg: same here, why?

public/css/list/item-list.less
```diff
@@ -4,6 +4,8 @@
 // Style
 
 .item-list {
+  @interactive-time-hover-bg: @base-gray-light;
```
nilmerg: Not sure why this is a new variable. Didn't track my past comments back, but now, why not just using `@gray-light`?

public/js/timestamp-toggle.js
```diff
@@ -0,0 +1,105 @@
...
+                    if (diff < 3600) {
+                        const parts = el.dataset.agoLabel.split(/\d+/);
+                        el.textContent = parts[0]
+                            + Math.floor(diff / 60) + parts[1]
+                            + Math.floor(diff % 60) + parts[2];
+                    }
+                });
+            } else {
+                container.querySelectorAll('.item-list .interactive-time').forEach(el => {
+                    el.removeAttribute('data-relative-time');
```
nilmerg: `del el.dataset.relativeTime;`

public/js/timestamp-toggle.js
```diff
@@ -0,0 +1,105 @@
...
+                '.search-bar': 'action'
+            }).forEach(([selector, attr]) => {
+                container.querySelectorAll(selector).forEach((el) => {
+                    el.setAttribute(attr, this.icinga.utils.addUrlParams(el.getAttribute(attr), {timestamps: preference}));
+                });
+            });
+
+            if (relative) {
+                container.querySelectorAll('.item-list .interactive-time').forEach(el => {
+                    el.setAttribute('data-relative-time', 'ago');
```
nilmerg: `el.dataset.relativeTime = 'ago';`

nilmerg (changes requested):
Please clean up your commits, squashing refactor commits to their respective ancestor etc.

library/Icingadb/View/EventRenderer.php
BastianLedererIcinga: I could've sworn this is what it looked like in the mockup, but after checking again I guess I mixed something up.
Would you prefer to always use the representation of `TimeAgo` when relative is enabled?

That will require `timestamp-toggle.js` to support the full range of representations that `TimeAgo` can render, which will be a bigger change so I'd like to confirm the intended behavior before implementing it.

nilmerg: I guess so! Though, of course it only needs to animate formats where seconds are relevant. *Most* (except history, but anyway) views have auto refresh enabled, so a change from `1d 23h ago` to `2d ago` (if that's the format, didn't verify) isn't required.

BastianLedererIcinga: Ok I adjusted the renderers and the js, but it requires `TimeAgo::format()` to be public so the js can easily swap between the representations.

public/js/timestamp-toggle.js
```diff
@@ -0,0 +1,107 @@
...
+                    } else {
+                        el.textContent = el.dataset.relativeText;
+                    }
+                });
+            } else {
+                container.querySelectorAll('.item-list .interactive-time').forEach(el => {
+                    delete el.dataset.relativeTime;
+                    el.textContent = this.timeFormatter.format(new Date(el.getAttribute('datetime')));
+                });
+            }
```
nilmerg: https://github.com/Icinga/ipl-web/pull/371 just *widened* my view of this proposal, to be honest.

Let me start with `\ipl\Web\Widget\Time` and why `\ipl\Web\Widget\Time::assemble` should not declare `data-ago-label`. `js/RelativeTime` only requires this in case `data-relative-time` is set to `"until"`. The only case where this is true, is when `\ipl\Web\Widget\TimeUntil` is used. So it's the job of `\ipl\Web\Widget\TimeUntil` to declare `data-ago-label`.

If this had been already the case prior, you couldn't have accessed it here as well, potentially leading to the questioning of its need or at least the consideration of another approach. (In which case we would end up with https://github.com/Icinga/ipl-web/pull/371 again I guess)

What this tells me, is that there are two implementations which seemingly need the same behavior:

- The backend renders absolute and relative times
- The frontend renders absolute and relative times

The backend relies on ipl-web's widgets now, in case of relative times. Absolute times are rendered using a custom formatter.

The frontend also renders absolute times using a custom formatter, but only supports a specific relative format (`ago`) as it's part of the history views only. Now that I pointed out that this is also only the case for durations lower than an hour, you needed to synchronize this with the backend.

So, let's take a step back.

It seems we went into the wrong direction already way beforehand. To me the frontend doesn't need to know anything about the various relative formats. It just allows the user to switch between absolute and relative after all.

Let me propose a fundamentally different, but simple approach:

* The backend always renders **both** representations, two `<time>` tags, but one with the `hidden` attribute set, according to the request parameter
* The frontend only toggles the `hidden` attribute

Oh, and, please change what I mentioned regarding `data-ago-label` at the beginning. 😉

nilmerg (changes requested):

library/Icingadb/View/EventRenderer.php
```diff
@@ -414,7 +426,29 @@ public function assembleCaption($item, HtmlDocument $caption, string $layout): v
 
     public function assembleExtendedInfo($item, HtmlDocument $info, string $layout): void
     {
-        $info->addHtml(new TimeAgo($item->event_time->getTimestamp()));
+        $timeRelative = new TimeAgo($item->event_time->getTimestamp());
```
nilmerg: Please pass the datetime object directly

library/Icingadb/View/NotificationRenderer.php
```diff
@@ -154,7 +169,29 @@ public function assembleCaption($item, HtmlDocument $caption, string $layout): v
 
     public function assembleExtendedInfo($item, HtmlDocument $info, string $layout): void
     {
-        $info->addHtml(new TimeAgo($item->send_time->getTimestamp()));
+        $timeRelative = new TimeAgo($item->send_time->getTimestamp());
```
nilmerg: same here

public/js/timestamp-toggle.js
```diff
@@ -0,0 +1,78 @@
...
+                '.load-more .action-link, .refresh-container-control, .primary-nav li.active a': 'href',
+                '.search-editor-opener': 'data-search-editor-url',
+                '.search-bar': 'action'
+            }).forEach(([selector, attr]) => {
+                container.querySelectorAll(selector).forEach((el) => {
+                    el.setAttribute(attr, this.icinga.utils.addUrlParams(el.getAttribute(attr), {timestamps: preference}));
+                });
+            });
+
+            container.querySelectorAll('.item-list .interactive-time').forEach(el => {
```
nilmerg: The `.item-list` selector brings up an important point: I was about to ask why you use it here but nowhere else. Though, this here isn't the problem, it's that a rather generic selector is used anywhere else. Just take a look at `.timestamp-toggle`, don't you think that's a rather broad selector for an event listener? Remember, they listen for all events on the current document. So please choose more concrete class names.

nilmerg (changes requested):
ipl-web will animate the widgets even if hidden by the way. Please open a separate PR over there to let it ignore those.

library/Icingadb/View/NotificationRenderer.php
```diff
@@ -35,6 +38,18 @@ class NotificationRenderer implements ItemRenderer
...
+
+    /**
+     * Create a NotificationRenderer
+     *
+     * @param bool $useRelativeTimestamps Whether to use relative timestamps
+     */
+    public function __construct(bool $useRelativeTimestamps = false)
+    {
+        $this->useRelativeTimestamps = $useRelativeTimestamps;
+    }
```
nilmerg, line 52: empty line afterwards please

nilmerg (changes requested):

nilmerg (approved):
LGTM!

Issue #823, opened 2023-07-28
List views with `tabular` view mode are broken


yhabteab:
## Describe the bug

Using the `tabular` view mode without providing custom columns causes PHP fatal error.
<img width="1248" alt="Bildschirmfoto 2023-07-28 um 13 53 55" src="https://github.com/Icinga/icingadb-web/assets/57616252/b355d037-a639-4e57-99bc-c7cf6da15bca">

PR #1333, opened 2026-02-13
Add a control to choose columns in tabular view mode

### LINKED ISSUE #82: Column chooser
The monitoring module allows to show additional columns in the host and service list via the `addColumns` parameter. There's no widget though with which this can be easily adjusted and neither does the result look good.

We should provide a widget, like the one used (planned..) for sorting. It should probably also be possible to move and resize the columns.

## Additional context
https://github.com/Icinga/icingaweb2/issues/3977

BastianLedererIcinga:
Add a new form to choose the displayed columns in hosts and services view.

resolves: #82

requires: https://github.com/Icinga/ipl-web/pull/277 for the `TabularViewModeSwitcher` icon
requires: https://github.com/Icinga/ipl-web/pull/357 to display error messages for invalid columns

### DIFF
--- a/application/controllers/HostsController.php
+++ b/application/controllers/HostsController.php
@@ -69,7 +69,6 @@ class HostsController extends Controller
         );
         $columns = $this->createColumnControl(
             $hosts,
-            $viewModeSwitcher,
             Url::fromPath('icingadb/hosts/suggestColumns'),
             ['host.name', 'host.state.output'],
             Url::fromPath('icingadb/hosts')
@@ -234,11 +233,6 @@ class HostsController extends Controller
         $this->addTitleTab($this->translate('Select Columns'));
         $columnChooser = $this->createColumnControl(
             Host::on($this->getDb()),
-            $this->createViewModeSwitcher(
-                $this->createPaginationControl(Host::on($this->getDb())),
-                $this->createLimitControl(),
-                viewModeSwitcherClass: TabularViewModeSwitcher::class,
-            ),
             Url::fromPath('icingadb/hosts/suggestColumns'),
             ['host.name', 'host.state.output'],
             Url::fromPath('icingadb/hosts')
--- a/application/controllers/HostsController.php
+++ b/application/controllers/HostsController.php
@@ -72,7 +72,14 @@ class HostsController extends Controller
             $limitControl,
             viewModeSwitcherClass: TabularViewModeSwitcher::class
         );
-        $columns = $this->createColumnControl($hosts, $viewModeSwitcher, ['host.name', 'host.state.output']);
+        $columns = $this->createColumnControl(
+            $hosts,
+            $viewModeSwitcher,
+            Url::fromPath('icingadb/hosts/suggestColumns'),
+            Host::on($this->getDb())->getResolver(),
+            ['host.name', 'host.state.output']
+        )
+            ->getColumns();
 
         $searchBar = $this->createSearchBar($hosts, [
             $limitControl->getLimitParam(),
@@ -230,25 +237,18 @@ class HostsController extends Controller
     public function columnControlAction()
     {
         $this->addTitleTab($this->translate('Select Columns'));
-        $this->addContent(
-            (new ColumnChooser(Url::fromPath('icingadb/hosts/suggestColumns'), Host::on($this->getDb())->getResolver()))
-                ->setAction((string) Url::fromRequest())
-                ->on(ColumnChooser::ON_SENT, function (ColumnChooser $form) {
-                    if ($form->hasBeenSubmitted()) {
-                        $url = Url::fromPath('icingadb/hosts');
-                        $url->setParam('columns', $form->getValue('columns', ''));
-                        $this->redirectNow($url);
-                    } else {
-                        foreach ($form->getPartUpdates() as $update) {
-                            if (! is_array($update)) {
-                                $update = [$update];
-                            }
-
-                            $this->addPart(...$update);
-                        }
-                    }
-                })->handleRequest($this->getServerRequest())
-        );
+        $columnChooser = $this->createColumnControl(
+            Host::on($this->getDb()),
+            $this->createViewModeSwitcher(
+                $this->createPaginationControl(Host::on($this->getDb())),
+                $this->createLimitControl(),
+                viewModeSwitcherClass: TabularViewModeSwitcher::class
+            ),
+            Url::fromPath('icingadb/hosts/suggestColumns'),
+            Host::on($this->getDb())->getResolver(),
+            ['host.name', 'host.state.output']
+        )->handleRequest($this->getServerRequest());
+        $this->addContent($columnChooser);
     }
 
     protected function fetchCommandTargets(): Query
--- a/application/controllers/HostsController.php
+++ b/application/controllers/HostsController.php
@@ -70,7 +70,7 @@ class HostsController extends Controller
             $limitControl,
             viewModeSwitcherClass: TabularViewModeSwitcher::class
         );
-        $columns = $this->createColumnControl($hosts, $viewModeSwitcher);
+        $columns = $this->createColumnControl($hosts, $viewModeSwitcher, ['host.name', 'host.state.output']);
 
         $searchBar = $this->createSearchBar($hosts, [
             $limitControl->getLimitParam(),
@@ -295,10 +295,5 @@ class HostsController extends Controller
         $this->filter($summary);
 
         return new FeatureStatus('host', $summary->first());
-    }
-
-    protected function getDefaultColumns(): string
-    {
-        return 'host.name, host.state.output';
     }
 }
--- a/application/controllers/HostsController.php
+++ b/application/controllers/HostsController.php
@@ -25,6 +25,7 @@ use Icinga\Module\Icingadb\Widget\ItemTable\HostItemTable;
 use Icinga\Module\Icingadb\Web\Control\ViewModeSwitcher;
 use Icinga\Module\Icingadb\Widget\ShowMore;
 use ipl\Orm\Query;
+use ipl\Orm\Resolver;
 use ipl\Stdlib\Filter;
 use ipl\Web\Control\LimitControl;
 use ipl\Web\Control\SortControl;
@@ -201,6 +202,20 @@ class HostsController extends Controller
         $requestData = json_decode($request->getBody()->read(8192), true);
         if (! array_key_exists('type', $requestData['term'])) {
             $requestData['term']['type'] = 'column';
+            $host = new Host();
+            $resolver = new Resolver($host::on($this->getDb()));
+            $columns = iterator_to_array(ObjectSuggestions::collectFilterColumns($host, $resolver));
+
+            $excluded = array_flip($requestData['exclude']);
+            $excludedCustomVars = array_filter(
+                $excluded,
+                fn(string $col) => str_contains($col, '.vars.'),
+                ARRAY_FILTER_USE_KEY
+            );
+
+            $suggestions
+                ->withFixedColumns(array_diff_key($columns, $excluded))
+                ->excludeCustomVars($excludedCustomVars);
         }
 
         $request = $request->withBody(Utils::streamFor(json_encode($requestData)));
--- a/application/controllers/HostsController.php
+++ b/application/controllers/HostsController.php
@@ -6,13 +6,16 @@
 namespace Icinga\Module\Icingadb\Controllers;
 
 use GuzzleHttp\Psr7\ServerRequest;
+use GuzzleHttp\Psr7\Utils;
 use Icinga\Module\Icingadb\Common\CommandActions;
 use Icinga\Module\Icingadb\Common\Links;
 use Icinga\Module\Icingadb\Model\Host;
 use Icinga\Module\Icingadb\Model\HoststateSummary;
 use Icinga\Module\Icingadb\Redis\VolatileStateResults;
 use Icinga\Module\Icingadb\Util\FeatureStatus;
+use Icinga\Module\Icingadb\Web\Control\ColumnChooser;
 use Icinga\Module\Icingadb\Web\Control\SearchBar\ObjectSuggestions;
+use Icinga\Module\Icingadb\Web\Control\TabularViewModeSwitcher;
 use Icinga\Module\Icingadb\Web\Controller;
 use Icinga\Module\Icingadb\Widget\Detail\MultiselectQuickActions;
 use Icinga\Module\Icingadb\Widget\Detail\ObjectsDetail;
@@ -194,7 +197,14 @@ class HostsController extends Controller
     {
         $suggestions = new ObjectSuggestions();
         $suggestions->setModel(Host::class);
-        $suggestions->forRequest(ServerRequest::fromGlobals());
+        $request = clone ServerRequest::fromGlobals();
+        $requestData = json_decode($request->getBody()->read(8192), true);
+        if (! array_key_exists('type', $requestData['term'])) {
+            $requestData['term']['type'] = 'column';
+        }
+
+        $request = $request->withBody(Utils::streamFor(json_encode($requestData)));
+        $suggestions->forRequest($request);
         $this->getDocument()->add($suggestions);
     }
 
@@ -209,6 +219,30 @@ class HostsController extends Controller
 
         $this->getDocument()->add($editor);
         $this->setTitle(t('Adjust Filter'));
+    }
+
+    public function columnControlAction()
+    {
+        $this->addTitleTab($this->translate('Select Columns'));
+        $this->addContent(
+            (new ColumnChooser(Url::fromPath('icingadb/hosts/complete'), Host::on($this->getDb())->getResolver()))
+                ->setAction((string) Url::fromRequest())
+                ->on(ColumnChooser::ON_SENT, function (ColumnChooser $form) {
+                    if ($form->hasBeenSubmitted()) {
+                        $url = Url::fromPath('icingadb/hosts');
+                        $url->setParam('columns', $form->getValue('columns', ''));
+                        $this->redirectNow($url);
+                    } else {
+                        foreach ($form->getPartUpdates() as $update) {
+                            if (! is_array($update)) {
+                                $update = [$update];
+                            }
+
+                            $this->addPart(...$update);
+                        }
+                    }
+                })->handleRequest($this->getServerRequest())
+        );
     }
 
     protected function fetchCommandTargets(): Query
@@ -242,5 +276,15 @@ class HostsController extends Controller
         $this->filter($summary);
 
         return new FeatureStatus('host', $summary->first());
+    }
+
+    protected function getViewModeSwitcherInstance(): ViewModeSwitcher
+    {
+        return new TabularViewModeSwitcher();
+    }
+
+    protected function getDefaultColumns(): string
+    {
+        return 'host.name, host.state.output';
     }
 }
--- a/application/controllers/HostsController.php
+++ b/application/controllers/HostsController.php
@@ -6,14 +6,12 @@
 namespace Icinga\Module\Icingadb\Controllers;
 
 use GuzzleHttp\Psr7\ServerRequest;
-use GuzzleHttp\Psr7\Utils;
 use Icinga\Module\Icingadb\Common\CommandActions;
 use Icinga\Module\Icingadb\Common\Links;
 use Icinga\Module\Icingadb\Model\Host;
 use Icinga\Module\Icingadb\Model\HoststateSummary;
 use Icinga\Module\Icingadb\Redis\VolatileStateResults;
 use Icinga\Module\Icingadb\Util\FeatureStatus;
-use Icinga\Module\Icingadb\Web\Control\ColumnChooser;
 use Icinga\Module\Icingadb\Web\Control\SearchBar\ObjectSuggestions;
 use Icinga\Module\Icingadb\Web\Control\TabularViewModeSwitcher;
 use Icinga\Module\Icingadb\Web\Controller;
@@ -25,12 +23,9 @@ use Icinga\Module\Icingadb\Widget\ItemTable\HostItemTable;
 use Icinga\Module\Icingadb\Web\Control\ViewModeSwitcher;
 use Icinga\Module\Icingadb\Widget\ShowMore;
 use ipl\Orm\Query;
-use ipl\Orm\Relations;
-use ipl\Orm\Resolver;
 use ipl\Stdlib\Filter;
 use ipl\Web\Control\LimitControl;
 use ipl\Web\Control\SortControl;
-use ipl\Web\FormElement\SearchSuggestions;
 use ipl\Web\Url;
 
 class HostsController extends Controller
--- a/application/controllers/HostsController.php
+++ b/application/controllers/HostsController.php
@@ -65,7 +65,11 @@ class HostsController extends Controller
             ],
             ['host.state.severity DESC', 'host.state.last_state_change DESC']
         );
-        $viewModeSwitcher = $this->createViewModeSwitcher($paginationControl, $limitControl);
+        $viewModeSwitcher = $this->createViewModeSwitcher(
+            $paginationControl,
+            $limitControl,
+            viewModeSwitcherClass: TabularViewModeSwitcher::class
+        );
         $columns = $this->createColumnControl($hosts, $viewModeSwitcher);
 
         $searchBar = $this->createSearchBar($hosts, [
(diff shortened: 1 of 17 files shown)


library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -43,9 +43,12 @@ class ObjectSuggestions extends Suggestions
     /** @var array */
     protected $customVarSources;
 
-    /** @var ?array<string, string> */
+    /** @var ?array<string, mixed> */
     protected ?array $fixedColumns = null;
 
+    /** @var array Names of excluded custom variables as {model}.vars.{flatname} */
+    protected array $excludedCustomVars = [];
```
nilmerg: Please tell me how I'd notice that this is required and for what. I can't imagine this right now and why custom variables are special in that regard. The `TermInput` already excludes already entered terms in the suggestion list.

nilmerg (changes requested):

nilmerg (changes requested):
I can see why you're not satisfied with the usage of `SearchSuggestions`. To get to a desirable result, a larger change is required. `ObjectSuggestions` were never designed to be more than a provider for the `SearchBar`. Now you're borrowing a part of its implementation either by making a protected method public or by copying parts of the code and reduce them to what you need.

I'd suggest to put the logic to retrieve custom variables and model columns for suggestions in a dedicated location that can be used by `ObjectSuggestions` and `SearchSuggestions` equally.

```php
namespace Icinga\Module\Icingadb\Data;

/** Not necessarily complete, but I hope you get the idea */
class QueryColumnsProvider implements IteratorAggregate
{
    public function __construct(Query $query, string $searchTerm);
    
    public function setFixedColumns(array $columns);

    public function setCustomVarSources(array $sources);

    /** Basically a copy from `ObjectSuggestions` but without the need for a base implementation */
    protected function matchSuggestion($path, $label, $searchTerm);

    public function getIterator(): Generator;

    /** public because `ObjectSuggestions::fetchValueSuggestions()` need it as well */
    public static function collectFilterColumns(…): Generator;
}
```

This should serve functionality currently only implemented in `ObjectSuggestions` to retrieve columns and custom variable names. It should be mostly a drop-in replacement for what `ObjectSuggestions::fetchColumnSuggestions()` currently does. Though, there is one thing it cannot provide:

`addHtml` – It's not a html element nor a container. The reason this is required comes from what the `SearchSuggestions` define as _grouping_. Though, in case of the column chooser this is required as well I think. `QueryColumnsProvider` should group by default. i.e. It yields in two levels, first a string key and a generator as value. `SearchSuggestions` should be able to handle this right now already, `ObjectSuggestions` need to be adjusted for this.

I hope this is sufficient to get you on track again.

BastianLedererIcinga:
> `addHtml` – It's not a html element nor a container. The reason this is required comes from what the `SearchSuggestions` define as _grouping_. Though, in case of the column chooser this is required as well I think. `QueryColumnsProvider` should group by default. i.e. It yields in two levels, first a string key and a generator as value. `SearchSuggestions` should be able to handle this right now already, `ObjectSuggestions` need to be adjusted for this.


From my understanding `SearchSuggestions` can't handle grouping the way you suggested.
When no `groupingCallback` is set the output is not grouped at all, when it is set `yield_groups()` is used to create the groups, which expects a flat `Traversable`. 

I added a group key to the values yielded by QueryColumnsProvider instead.

library/Icingadb/Web/Controller.php
```diff
@@ -76,46 +80,43 @@ public function getFilter(): Filter\Rule
...
-                $columns[] = $column;
-            }
-        }
+    public function createColumnControl(
+        Query $query,
+        Url $suggestionUrl,
+        array $defaultColumns,
+        Url $redirectUrl
+    ): ColumnChooser {
+        $columnsDef = $this->params->shift('columns', $defaultColumns);
```
nilmerg: `array` is a rather complex type to expect `UrlParams::shift()` to return.

```suggestion
        $columnsDef = $this->params->shift('columns') ?? $defaultColumns;
```

library/Icingadb/Web/Controller.php
```diff
@@ -76,46 +80,43 @@ public function getFilter(): Filter\Rule
...
+        $chooser = (new ColumnChooser($suggestionUrl, $query->getResolver(), $columnsDef))
+            ->setAction((string) Url::fromRequest())
+            ->on(ColumnChooser::ON_SENT, function (ColumnChooser $form) use ($redirectUrl) {
+                if ($form->hasBeenSubmitted()) {
+                    $redirectUrl->setParam('columns', $form->getValue('columns', ''));
```
nilmerg: Please consider adding either a separate getter for the serialized value or extend `getColumns()` with a parameter that allows to return it instead of an array. Accessing form internals is unnecessary as long we have full control over the form implementation.

library/Icingadb/Web/Controller.php
```diff
@@ -76,46 +80,43 @@ public function getFilter(): Filter\Rule
...
      * @param Query $query
-     * @param ViewModeSwitcher $viewModeSwitcher
+     * @param Url $suggestionUrl
+     * @param array $defaultColumns
+     * @param Url $redirectUrl
      *
-     * @return array provided columns
-     *
-     * @throws HttpBadRequestException
+     * @return ColumnChooser provided columns
```
nilmerg: Either remove *provided columns* or provide a better fitting alternative. It's an object after all.

library/Icingadb/Web/Controller.php
```diff
@@ -500,4 +496,30 @@ protected function moduleInit()
...
     }
+
+    /**
+     * Add column suggestions for the given model
+     *
+     * @param Model $model
+     *
+     * @return void
+     */
+    protected function suggestColumns(Model $model): void
```
nilmerg: Why an instance and not a class-string? I mean, the instance isn't required 🤷‍♂️

library/Icingadb/Web/Controller.php
```diff
@@ -500,4 +496,30 @@ protected function moduleInit()
...
+            (function () use (&$suggestions, $provider) {
+                $provider->setSearchTerm($suggestions->getSearchTerm());
+                foreach ($provider as $suggestion) {
+                    if (! in_array($suggestion['search'], $suggestions->getExcludeTerms())) {
+                        yield $suggestion;
+                    }
+                }
+            })()
+        );
+        $suggestions->forRequest(ServerRequest::fromGlobals());
```
nilmerg: Woah! How long did it take for you to understand this?

1. provider is initialized
2. The closure is called
3. suggestions are initialized
4. request data is loaded into suggestions

But the provider still gets request data? Yeah, because the closure is a generator which isn't evaluated until the iteration starts…

But in case the closure is not a generator, it has no way to access the search term, as the suggestions cannot be initialized without a provider and there is no way to load request data without them.

This needs a refactoring of `SearchSuggestions`, seriously. I'll open an issue for ipl-web.

library/Icingadb/Web/Control/ColumnChooser.php
```diff
@@ -0,0 +1,104 @@
...
+            }
+        } else {
+            $this->columns = $columns;
+        }
+
+        $this->suggestionUrl = $suggestionUrl;
+        $this->resolver = $resolver;
+    }
+
+    public function getColumns(): array
```
nilmerg, line 54: phpdoc missing

library/Icingadb/Web/Control/ColumnChooser.php
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+    }
+
+    protected function assemble()
+    {
+        $termInput = (new TermInput(
+            'columns',
+            [
+                'type' => 'text',
```
nilmerg: What's this for?

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -224,291 +202,38 @@ protected function fetchColumnSuggestions($searchTerm)
...
-                            $titleAdded = true;
-                        }
-
-                        yield $relation . '.vars.' . $search => sprintf($label, $name);
-                    }
-                }
-            }
-        }
+        $provider = (new QueryColumnsProvider($query, $searchTerm))
+            ->setCustomVarSources($this->customVarSources);
```
nilmerg, line 201: Isn't it strange that the provider uses the same custom var sources by default?

My suggestion didn't include a default customvar source array as there isn't on required, is it? `ObjectSuggestions` use their specific set and the hosts/services column control its own. Both call the setter. In case we need the provider to yield only columns, as might be the case anyway in case of fixed columns, there's no need for customvar sources.

library/Icingadb/Web/Controller.php
```diff
@@ -500,4 +496,30 @@ protected function moduleInit()
...
+        $suggestions->setGroupingCallback(fn($x) => $x['group']);
```
nilmerg: Woah! How long did it take for you to understand this?

1. provider is initialized
2. The closure is called
3. suggestions are initialized
4. request data is loaded into suggestions

But the provider still gets request data? Yeah, because the closure is a generator which isn't evaluated until the iteration starts…

But in case the closure is not a generator, it has no way to access the search term, as the suggestions cannot be initialized without a provider and there is no way to load request data without them.

`SearchSuggestions` were designed as dumb as possible, which is why the provider just needs to be a `Traversable`. But I think we can avoid the closure at all:

```suggestion
        $provider = new QueryColumnsProvider($query);
        $suggestions = new SearchSuggestions($provider);
        $suggestions->forRequest(ServerRequest::fromGlobals());
        $provider->forSuggestions($suggestions);
```

`forSuggestions` needs to be added. It's task would be to configure itself based on the suggestions. i.e. registering the search term, the exclusion list and setting up the grouping callback. The exclusion list should potentially be used to filter the queries already if possible.

nilmerg (changes requested):
Didn't look in detail at `QueryColumnsProvider` and `ObjectSuggestions` yet.

application/controllers/HostsController.php
```diff
@@ -211,7 +211,19 @@ public function completeAction()
...
+            'eventcommand'          => t('Eventcommand %s', '..<customvar-name>'),
+            'host'                  => t('Host %s', '..<customvar-name>'),
+            'hostgroup'             => t('Hostgroup %s', '..<customvar-name>'),
+            'notification'          => t('Notification %s', '..<customvar-name>'),
+            'notificationcommand'   => t('Notificationcommand %s', '..<customvar-name>'),
+            'service'               => t('Service %s', '..<customvar-name>'),
+            'servicegroup'          => t('Servicegroup %s', '..<customvar-name>'),
+            'timeperiod'            => t('Timeperiod %s', '..<customvar-name>'),
+            'user'                  => t('Contact %s', '..<customvar-name>'),
+            'usergroup'             => t('Contactgroup %s', '..<customvar-name>')
```
nilmerg: Not sure whether it's a good idea to suggest this many custom variables. Let's reduce this to host and service variables. This is about the display of these values and as such about joining them into the hosts or services table and I'm not unsure what the performance impact of joining servicegroup variables into hosts is or whether I'm willing to wait for user's complaining about it ;)

library/Icingadb/Data/QueryColumnsProvider.php
```diff
@@ -171,7 +174,10 @@ protected function fetchColumns(): Generator
             $this->query->getResolver()
         );
         foreach ($columns as $columnName => $columnMeta) {
-            if ($this->matchSuggestion($columnName, $columnMeta, $this->searchTerm)) {
+            if (
+                ! in_array($columnName, $this->excludedColumns)
```
nilmerg: `in_array` always with the `strict` parameter set. This has often bit us in the past.

library/Icingadb/Data/QueryColumnsProvider.php
```diff
@@ -356,7 +372,17 @@ public function queryCustomvarConfig(Filter\Rule $filter): Select
...
                 $this->applyBaseFilter($query);
 
-                $aggregates[$name] = new Expression("MAX($name)");
+                if (isset($excludedByRelation[$name])) {
+                    $placeholders = implode(', ', array_fill(0, count($excludedByRelation[$name]), '?'));
+                    $aggregates[$name] = new Expression(
+                        "CASE WHEN flatname NOT IN ($placeholders) THEN MAX($name) ELSE NULL END",
+                        null,
+                        ...$excludedByRelation[$name]
+                    );
```
nilmerg: Did you consider adding this as filter to the query instead? A `Filter::unequal('flatname', $excludedByRelation[$name])` should work I suppose.

library/Icingadb/Data/QueryColumnsProvider.php
```diff
@@ -388,25 +414,33 @@ public function queryCustomvarConfig(Filter\Rule $filter): Select
...
+    /**
+     * Set columns to exclude from the suggestions
+     *
+     * @param array $columns
      *
-     * @return array<string, string>
+     * @return $this
      */
-    public static function getDefaultCustomVarSources()
+    public function setExcludedColumns(array $columns): static
```
nilmerg: Please put setters/getters near the top of a class

library/Icingadb/Web/Controller.php
```diff
@@ -500,4 +496,22 @@ protected function moduleInit()
...
+     * @param array<string, string> $customVarSources
+     *
+     * @return void
+     */
+    protected function suggestColumns(string $model, array $customVarSources = []): void
+    {
+        $query = $model::on($this->getDb());
+        $provider = (new QueryColumnsProvider($query))->setCustomVarSources($customVarSources);
+        $suggestions = new SearchSuggestions($provider);
+        $suggestions->forRequest(ServerRequest::fromGlobals());
```
nilmerg: Please use `getServerRequest` instead

library/Icingadb/Data/QueryColumnsProvider.php
BastianLedererIcinga: Yes, I have actually tried that. The problem is that the filter excludes entire hosts/services that have this custom var set from the result of the query. 
In our dev environment you can easily try this by selecting a host custom var that is set on all the dummy hosts like `location`, then try to enter an other of these custom vars like `department`. `host.vars.department` is not suggested because the filter removed all dummy hosts from the result.

application/controllers/HostsController.php
```diff
@@ -197,6 +209,14 @@ public function completeAction()
         $this->getDocument()->add($suggestions);
     }
 
+    public function suggestColumnsAction()
+    {
+        $this->suggestColumns(Host::class, [
+            'host'                  => t('Host %s', '..<customvar-name>'),
+            'service'               => t('Service %s', '..<customvar-name>')
```
nilmerg: Reduce indentation please

application/controllers/ServicesController.php
```diff
@@ -214,6 +226,14 @@ public function completeAction()
         $this->getDocument()->add($suggestions);
     }
 
+    public function suggestColumnsAction()
+    {
+        $this->suggestColumns(Service::class, [
+            'host'                  => t('Host %s', '..<customvar-name>'),
+            'service'               => t('Service %s', '..<customvar-name>')
```
nilmerg: Reduce indentation please

library/Icingadb/Data/QueryColumnsProvider.php
```diff
@@ -0,0 +1,480 @@
...
+                $parsedArrayVars[$name] = true;
+                $search = $name . '[*]';
+            }
+
+            foreach ($this->customVarSources as $relation => $label) {
+                if (isset($customVar->$relation)) {
+                    yield [
+                        'search' => $relation . '.vars.' . $search,
+                        'label'  => sprintf($label, $name),
+                        'group'  => t('Best Suggestions')
```
nilmerg: Please use ipl-i18n's `Translation` trait instead of the global functions.

library/Icingadb/Data/QueryColumnsProvider.php
```diff
@@ -0,0 +1,480 @@
...
+                    $resolver->qualifyPath($name, $tableName)
+                );
+
+                $this->applyBaseFilter($query);
+                $select = $query->assembleSelect();
+
+                if (isset($excludedByRelation[$name])) {
+                    $flatname = $resolver->qualifyColumn('flatname', $resolver->getAlias($customVars->getModel()));
+                    $select->where(
+                        "$flatname Not In (?)",
```
nilmerg: `NOT IN` please

library/Icingadb/Data/QueryColumnsProvider.php
```diff
@@ -0,0 +1,480 @@
...
+            $this->query->getResolver()
+        );
+        foreach ($columns as $columnName => $columnMeta) {
+            if (
+                ! in_array($columnName, $this->excludedColumns, true)
+                && $this->matchSuggestion($columnName, $columnMeta, $this->searchTerm)
+            ) {
+                yield [
+                    'search' => $columnName,
+                    'label'  => $columnMeta,
```
nilmerg, line 207: I noticed that a feature of `ObjectSuggestions` is missing here: Relation names. (Controlled by `shouldShowRelationFor`)

We have several duplicate column labels such as *Endpoint Name* that need further distinction in the suggestions. The search bar's `Suggestions` have this builtin, the `SearchSuggestions` on the other hand do not. But as said, they should be dumb, what I'd suggest is to teach them only about that *labels* might be `ipl\Html\Contract\ValidHtml` and in that case, they don't render an `<input type="button">` but a `<button type="button"></button>` (That's what `Suggestions` already do but always.)

So please add support for this, I will create an issue for ip-web, and replicate here the markup the `Suggestions` use to show relations.

nilmerg (changes requested):

library/Icingadb/Data/QueryColumnsProvider.php
```diff
@@ -0,0 +1,549 @@
...
+
+        return fnmatch($searchTerm, $label, FNM_CASEFOLD) || fnmatch($searchTerm, $path, FNM_CASEFOLD);
+    }
+
+    /**
+     * Apply restrictions and the base filter to the given query
+     *
+     * @param Query $query
+     */
+    private function applyBaseFilter(Query $query): void
```
nilmerg: This is the only private method here, why? If there's no reason except *why not*, make it protected.

nilmerg (changes requested):

nilmerg:
Oh, and please rebase with main.

library/Icingadb/Web/Controller.php
```diff
@@ -78,46 +82,43 @@ public function getFilter(): Filter\Rule
...
-                $columns[] = $column;
-            }
-        }
+    public function createColumnControl(
+        Query $query,
+        Url $suggestionUrl,
+        array $defaultColumns,
+        Url $redirectUrl
+    ): ColumnChooser {
+        $columnsDef = $this->params->shift('columns') ?? $defaultColumns;
```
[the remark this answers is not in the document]
nilmerg: Honestly, bullshit in, bullshit out. It's a rather helpful error since #823 anyway, so I don't see why we should fallback now considering we add a proper UI control here that doesn't allow an empty set.

library/Icingadb/Web/Control/ColumnChooser.php
```diff
@@ -0,0 +1,116 @@
...
+        foreach ($terms as $term) {
+            try {
+                $columnDefinition = $this->resolver->getColumnDefinition($term->getSearchValue());
+                $label = $columnDefinition->getLabel();
+                if ($label !== null) {
+                    $term->setLabel($label);
+                }
+            } catch (InvalidRelationException) {
+                $term->setMessage($this->translate('Is not a valid column'));
+            }
```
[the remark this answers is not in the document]
nilmerg: Not sure where this comes from. `getColumnDefinition` tries to resolve an invalid column as relation and the expected exception is thrown.

library/Icingadb/Data/QueryColumnsProvider.php
```diff
@@ -0,0 +1,549 @@
...
+    protected function matchSuggestion($path, $label, $searchTerm)
+    {
+        if (preg_match('/[_.](id|bin|checksum)$/', $path)) {
+            // Only suggest exotic columns if the user knows about them
+            $trimmedSearch = trim($searchTerm, ' *');
+            return substr($path, -strlen($trimmedSearch)) === $trimmedSearch;
+        }
+
+        return fnmatch($searchTerm, $label, FNM_CASEFOLD) || fnmatch($searchTerm, $path, FNM_CASEFOLD);
+    }
```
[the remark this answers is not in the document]
nilmerg: The question is rather, why does this method not use a strict interface?

library/Icingadb/Data/QueryColumnsProvider.php
```diff
@@ -0,0 +1,549 @@
...
+                );
+
+                $this->applyBaseFilter($query);
+                $select = $query->assembleSelect();
+
+                if (isset($excludedByRelation[$name])) {
+                    $flatname = $resolver->qualifyColumn('flatname', $resolver->getAlias($customVars->getModel()));
+                    $select->where(
+                        "$flatname NOT IN (?)",
+                        ...$excludedByRelation[$name]
```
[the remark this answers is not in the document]
nilmerg: This makes no difference :laughing:. The driver is intelligent enough here.

nilmerg (changes requested):
Plus what the AI commented and isn't resolved yet.

nilmerg (changes requested):
Code works and looks fine now. The CSS might need another look as the modal opener looks somehow skewed to me: 

Image

Missing padding at the bottom and the … is off to the left.

nilmerg:
Oh wait, it's not the padding that bothers me, it's that the background is always visible. The preview by @flourish86 also shows this only upon hover!

nilmerg:
Make the opener a `.control-button` instead, this should solve the problem.

BastianLedererIcinga:
> Make the opener a `.control-button` instead, this should solve the problem.

Using a `.control-button` seems to create some new problems.
Screenshot from 2026-04-14 16-36-13

But keeping the `.action-link` and only showing the background on hover should look fine asweell right?

Screenshot from 2026-04-14 16-37-45
Screenshot from 2026-04-14 16-37-48

nilmerg:
:rofl: This may be due to the fact that you render the *balls* by yourself. Try using an `\ipl\Web\Widget\Icon` using [`ellipsis`](https://fontawesome.com/v6/icons/ellipsis?f=classic&s=solid) instead.

library/Icingadb/Widget/ItemTable/StateItemTable.php
```diff
@@ -17,8 +17,12 @@
 use ipl\Orm\Common\SortUtil;
 use ipl\Orm\Query;
 use ipl\Web\Control\SortControl;
+use ipl\Web\Url;
+use ipl\Web\Widget\ActionLink;
+use ipl\Web\Widget\Ball;
 use ipl\Web\Widget\EmptyStateBar;
 use ipl\Web\Widget\Icon;
+use ipl\Web\Widget\Link;
```
nilmerg, line 23: Cleanup unused usages

public/css/list/state-item-table.less
```diff
@@ -86,6 +90,28 @@ table.state-item-table {
...
+          white-space: nowrap;
+          padding-left: 0.2em;
+          width: 3em;
+          display: flex;
+          align-items: stretch;
+
+          .control-button {
+            display: inline-flex;
+            align-items: center;
+          }
```
nilmerg, line 110:
```suggestion
        form {
          flex: 1 1 auto;
          padding-right: 1px;
        }

        .column-chooser-opener {
          width: 3em;
          line-height: 1;
          padding-left: 0.2em;
            
          i.icon {
            font-size: 2em;
            line-height: 0.5;
          }
```

library/Icingadb/Widget/ItemTable/StateItemTable.php
```diff
@@ -124,10 +131,46 @@ public function setSort(?string $sort): self
...
 
+    /**
+     * Create an action-link that opens a ColumnChooser form in a modal
+     *
+     * @return ValidHtml
+     */
+    protected function createColumnChooserOpener(): ValidHtml
+    {
+        return new HtmlElement(
+            'div',
```
nilmerg: Please use a `span`

nilmerg (changes requested):

nilmerg (changes requested):
Please update the requirements and require icinga-php-library >= 0.20.0. Note that these are defined in several places so it's best to search for the current requirement and update all occurrences.

nilmerg (approved):
LGTM!

PR #1348, opened 2026-03-18
Replace `[INFO]`/`(INFO)` with a `.state-info` ball in plugin output

jrauh01:
Like the existing OK/WARNING/CRITICAL/etc. markers, `[INFO]` and `(INFO)` in plugin output are now replaced with a state ball element so they render as a visual indicator instead of plain text.


lippserd:
@LordHepipud Could you please check whether the result of this change meets your requirements?

PR #1355, opened 2026-04-02
Fix incorrect start times in the timelines of fixed downtimes

### LINKED ISSUE #1353: Incorrect start time displayed for fixed downtimes
## Describe the bug

Fixed downtimes show time ago instead of time until for the start in the progress section

## To Reproduce

1. create a fixed downtime, ideally with a start time within the next hour
2. check the `DowntimeCard`

## Expected behavior
The start time shown in the progress section should match the configured start time.

## Screenshots
Image

## Additional context
`DowntimeCard` seems to always use `TimeAgo` for the start of fixed downtimes.

BastianLedererIcinga:
Downtimecards for fixed downtimes currently always use `TimeAgo` to display the start time in the timeline, which is incorrect if the start is in the future, for such cases `TimeUntil` should be used instead.

I tested the change with the current main of ipl-web and with https://github.com/Icinga/ipl-web/pull/350 and it's dependencies, both worked fine.

resolves: #1353

### DIFF
--- a/library/Icingadb/Widget/Detail/DowntimeCard.php
+++ b/library/Icingadb/Widget/Detail/DowntimeCard.php
@@ -217,7 +217,12 @@ class DowntimeCard extends BaseHtmlElement
                     Html::tag(
                         'div',
                         ['class' => 'bubble upwards'],
-                        new VerticalKeyValue(t('Start'), new TimeAgo($this->start))
+                        new VerticalKeyValue(
+                            t('Start'),
+                            time() >= $this->start
+                                ? new TimeAgo($this->start)
+                                : new TimeUntil($this->start)
+                        )
                     )
                 ),
                 Html::tag(


library/Icingadb/Widget/Detail/DowntimeCard.php
```diff
@@ -217,7 +217,12 @@ protected function assemble()
                     Html::tag(
                         'div',
                         ['class' => 'bubble upwards'],
-                        new VerticalKeyValue(t('Start'), new TimeAgo($this->start))
+                        new VerticalKeyValue(
+                            t('Start'),
+                            time() > $this->start
```
nilmerg: It's a minor thing, but above `>=` is used so should this.

nilmerg (changes requested):

nilmerg (approved):
Thank you!

PR #1364, opened 2026-04-21
Fix `LoadMore` in the service history

### LINKED ISSUE #1363: `LoadMore` in service history ignores `sort` parameter
## Describe the bug
The `LoadMore` action in the service history always uses ascending order.

## To Reproduce
1. Go to the history tab of a service
2. Select descending order
3. Click `Load More`

## Expected behavior
New elements should be added in the same order that was used previously.

## Screenshots

Image

## Your Environment

* Icinga DB Web version (System - About): main

BastianLedererIcinga:
`ServiceController::historyAction` previously used `Url::setParams` to set the parameters `name` and `host.name` in the URL for the load-more link.
This removed the `sort` parameter from the URL, so load-more always added elements in ascending order.

Using `UrlParams::addValues` does not affect the `sort` parameter and load-more in descending order works fine.

For consistency `HostController::historyAction` is also adjusted.

fix https://github.com/Icinga/icingadb-web/issues/1363

### DIFF
--- a/application/controllers/HostController.php
+++ b/application/controllers/HostController.php
@@ -141,8 +141,9 @@ class HostController extends Controller
 
         $before = $this->params->shift('before', time());
         $previousTimestamp = $this->params->shift('last-entry');
-        $url = Url::fromRequest()->setParams(clone $this->params);
-        $url->setParam('name', $this->host->name);
+        $url = Url::fromRequest()->setParams(
+            (clone $this->params)->add('name', $this->host->name)
+        );
 
         $timestampControl = $this->createTimestampControl();
         $limitControl = $this->createLimitControl();
--- a/application/controllers/ServiceController.php
+++ b/application/controllers/ServiceController.php
@@ -291,8 +291,10 @@ class ServiceController extends Controller
 
         $before = $this->params->shift('before', time());
         $previousTimestamp = $this->params->shift('last-entry');
-        $url = Url::fromRequest()->setParams(clone $this->params);
-        $url->setParams(['name' => $this->service->name, 'host.name' => $this->service->host->name]);
+        $url = Url::fromRequest()->setParams(
+            (clone $this->params)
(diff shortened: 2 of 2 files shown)


application/controllers/ServiceController.php
```diff
@@ -290,7 +290,8 @@ public function historyAction(): Generator
 
         $before = $this->params->shift('before', time());
         $url = Url::fromRequest()->setParams(clone $this->params);
-        $url->setParams(['name' => $this->service->name, 'host.name' => $this->service->host->name]);
+        $url->setParam('name', $this->service->name)
+            ->setParam('host.name', $this->service->host->name);
```
sukhwinder33445: I would suggest adding the parameters to the cloned parameters using `addValues()`, as this avoids having to call the setter again.
```suggestion
$url = Url::fromRequest()->setParams(
            (clone $this->params)
                ->addValues(['name' => $this->service->name, 'host.name' => $this->service->host->name])
        );
```

BastianLedererIcinga: Looks good, I made the change.
As you suggested in person I also adjusted `HostController::historyAction` for consistency.

application/controllers/HostController.php
```diff
@@ -139,8 +139,10 @@ public function historyAction(): Generator
         ));
 
         $before = $this->params->shift('before', time());
-        $url = Url::fromRequest()->setParams(clone $this->params);
-        $url->setParam('name', $this->host->name);
+        $url = Url::fromRequest()->setParams(
+            (clone $this->params)
+                ->addValues(['name' => $this->host->name])
```
sukhwinder33445: Please use `add()` for a single param.

sukhwinder33445 (dismissed):

nilmerg:
please rebase

BastianLedererIcinga:
> please rebase

Done

nilmerg (approved):

PR #1366, opened 2026-04-26
Preserve state labels in plugin output

fsalomon:
- Keep bracketed state labels as text inside plugin output (state-ball replacements)
  - Hide the text visually while preserving the existing state-ball styling
  - Update PluginOutput expectations for rendered state labels
  
Nice when copy&pasting plugin output to a code agent.

### DIFF
--- a/library/Icingadb/Util/PluginOutput.php
+++ b/library/Icingadb/Util/PluginOutput.php
@@ -34,13 +34,13 @@ class PluginOutput extends HtmlString
     protected const TEXT_REPLACEMENTS = [
         "\t",
         "\n",
-        '<span class="state-ball ball-size-m state-ok">[OK]</span>',
-        '<span class="state-ball ball-size-m state-warning">[WARNING]</span>',
-        '<span class="state-ball ball-size-m state-critical">[CRITICAL]</span>',
-        '<span class="state-ball ball-size-m state-unknown">[UNKNOWN]</span>',
-        '<span class="state-ball ball-size-m state-up">[UP]</span>',
-        '<span class="state-ball ball-size-m state-down">[DOWN]</span>',
-        '<span class="state-ball ball-size-m state-info">[INFO]</span>',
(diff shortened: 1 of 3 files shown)


fsalomon:
bump CLA

bobapple:
@cla-bot check

nilmerg (changes requested):
Thanks. Though, the alignment of the state-balls and their text is off now.

**Before**
Image

**After**
Image

You can avoid this by using our native way to hide text visually:

`<i class="sr-only">[OK]</i>`

fsalomon:
The `sr-only` approach works for screen readers and keeps the alignment intact, but it breaks the in-panel copy button. Using your screenshot's example, clicking the copy button now produces:

```
WARNING - states: warning=1 ok=2
[WARNING]
 Dual Humidity Port 1: 43%
[OK]
 Dual Temperature Port 1: 26C
[OK]
 Temp Sensor Test: 21C
```

instead of:

```
WARNING - states: warning=1 ok=2
[WARNING] Dual Humidity Port 1: 43%
[OK] Dual Temperature Port 1: 26C
[OK] Temp Sensor Test: 21C
```

Cause: `CopyToClipboard.js` in `icinga-php-library` reads via `element.innerText`, which is rendering-aware. Since `.sr-only` uses `position: fixed`, the inner `<i>` is auto-blockified (per CSS 2.1 §9.7), so `innerText` inserts a newline around each label. A native Ctrl+C over the same DOM is fine — `Range.toString()` returns the inline form.

Fix would be in `icinga-php-library`:

```diff
-copyText = clipboardSource.innerText;
+const range = document.createRange();
+range.selectNodeContents(clipboardSource);
+copyText = range.toString();
```

Happy to open that PR. Should I keep the `sr-only` markup here and let the upstream fix land separately, or revert to inline text in this PR?

nilmerg:
Sounds and works fine! It even returns rendered content indented as shown in case it's a list. Feel free to open an appropriate PR!

nilmerg (approved):

PR #1369, opened 2026-04-28
Add `V2/Source` hook for `icinga-notifications-web`

### LINKED ISSUE #1365: Implement updated integration for Icinga Notifications Web
Icinga Notifications Web v1.0 will again change how source integrations hook into rule filter configuration. So a different hook implementation needs to be provided. The actual implementation is still be designed, though. This should be worked on closely with whoever does that.

sukhwinder33445:
This PR adds a new `V2/Icinga2Source` hook implementation to integrate with the `V2/SourceHook` interface from `icinga-notifications-web`.

### Supporting changes required by the hook:
                                                                                                                                                                                                           
- Introduce `QueryValuesProvider` — Decouples value suggestion logic from `ObjectSuggestions` into a standalone Generator-based class. This allows the hook's `getValueSuggestions()` to reuse the same suggestion infrastructure without a `Suggestions` context.                                                                                                                                                 
- `QueryColumnsProvider`: Add `setShowRelationLabels()` setter — Exposes the relation label display option so the hook can control it.
- `QueryColumnsProvider`: Make `collectFilterColumns()` protected — No longer needs to be public; the override in `SearchControls` that called it statically is removed as the optimization it provided is no longer necessary.                           
- `ObjectSuggestions`: Remove label-to-path resolution for columns with spaces to ensure consistency between the suggestions in the search bar and suggestions provided by the `Icinga2Source`.

resolves: #1365
### Requires:
- https://github.com/Icinga/icinga-notifications-web/pull/469
- https://github.com/Icinga/icingadb-web/pull/1375
- https://github.com/Icinga/ipl-orm/pull/161


library/Icingadb/ProvidedHook/Notifications/V2/Source.php
```diff
@@ -0,0 +1,83 @@
...
+                    $condition->metaData()->set('columnLabel', $this->allowedColumns[$condition->getColumn()]);
+                } elseif (preg_match('/^(host|service)\.vars\.(.*)/i', $condition->getColumn(), $m)) {
+                    $prefix = $m[1] === 'host' ? $this->translate('Host') : $this->translate('Service');
+                    $condition->metaData()->set('columnLabel', $prefix . ' ' . $m[2]);
+                }
+            }
+        });
+        $editor->on(SearchEditor::ON_VALIDATE_COLUMN, function (Condition $condition) {
+            if (
+                ! isset($this->allowedColumns[$condition->getColumn()])
```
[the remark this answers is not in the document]
sukhwinder33445: Ahmm OK.

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
```diff
@@ -0,0 +1,83 @@
...
+        $this->allowedColumns = [
+            'host.name' => $this->translate('Host Name'),
+            'hostgroup.name' => $this->translate('Hostgroup Name'),
+            'host.user.name' => $this->translate('Contact Name'),
+            'host.usergroup.name' => $this->translate('Contactgroup Name'),
+            'service.name' => $this->translate('Service Name'),
+            'servicegroup.name' => $this->translate('Servicegroup Name'),
+            'service.user.name' => $this->translate('Contact Name'),
+            'service.usergroup.name' => $this->translate('Contactgroup Name')
+        ];
```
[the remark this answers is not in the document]
sukhwinder33445: This optimization is not required now, maybe later.

application/controllers/SuggestController.php
```diff
@@ -16,13 +14,7 @@ class SuggestController extends CompatController
...
     {
         $suggestions = (new ObjectSuggestions())
-            ->setModel(
-                match ($this->params->getRequired('type')) {
-                    'host', Source::TYPE_ALL => Host::class,
-                    'service' => Service::class,
-                    default => $this->httpBadRequest('Invalid type')
-                }
-            )
+            ->setModel(new Host())
```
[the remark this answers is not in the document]
sukhwinder33445: Changed.

application/controllers/SuggestController.php
```diff
@@ -16,13 +14,7 @@ class SuggestController extends CompatController
...
         $suggestions = (new ObjectSuggestions())
-            ->setModel(
-                match ($this->params->getRequired('type')) {
-                    'host', Source::TYPE_ALL => Host::class,
-                    'service' => Service::class,
-                    default => $this->httpBadRequest('Invalid type')
-                }
-            )
+            ->setModel(Host::class)
             ->onlyWithCustomVarSources(['host', 'service'])
```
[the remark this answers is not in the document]
sukhwinder33445: `V1/SourceHook` is no longer important.

library/Icingadb/ProvidedHook/Notifications/V2/Source.php
```diff
@@ -0,0 +1,83 @@
...
+                }
+            }
+        });
+        $editor->on(SearchEditor::ON_VALIDATE_COLUMN, function (Condition $condition) {
+            if (
+                array_key_exists($condition->getColumn(), $this->allowedColumns)
+                && ! preg_match('/^(?:host|service)\.vars\./i', $condition->getColumn())
+            ) {
+                throw new SearchException($this->translate('Is not a valid column'));
+            }
```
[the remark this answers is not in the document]
sukhwinder33445: Fixed.

BastianLedererIcinga:
I reduced the `SourceHook` interface in https://github.com/Icinga/icinga-notifications-web/pull/469, feel free to comment if you'd like anything changed about it.
Adjusting to the new implementation shouldn't be too complicated, the callables that are currently being registered on the `SearchEditor`  in `Source::getRuleFilterEditor()` can be used to implement the new interface functions with minor adjustments.

library/Icingadb/ProvidedHook/Notifications/V2/Source.php
```diff
@@ -0,0 +1,134 @@
...
+
+        if (is_int($key) && count($array) > 1) {
+            return [...$keys, '*'];
+        }
+
+        $keys[] = $key;
+
+        return is_array($array[$key])
+            ? $this->flattenKeys($array[$key], $keys)
+            : $keys;
```
BastianLedererIcinga: There are two problems with this function.
- When the function runs into an empty dictionary or array, `$key` becomes `null`, which causes an error.

- If a customvar contains several keys at the same level, this function always walks into the first one, no matter what the user input was, so the returned `jsonPath` is valid, but points at the wrong field. You will have to take the value of `$column` into account to get the correct result.

BastianLedererIcinga (dismissed):

library/Icingadb/ProvidedHook/Notifications/V2/Source.php
sukhwinder33445: The first problem is fixed with the `!empty()` check for the array.
The second one is a bug where multiple suggestions can be generated for the same customvar_id, because a customvar can be multi-dimensional and have multiple entries as flattened vars in the `customvar_flat` table. Since `$column` is user-defined and not guaranteed to exist in the database, it cannot be used. This should be fixed in a separate PR later.

sukhwinder33445: As discussed offline, I misunderstood your suggestion, but have now implemented it.

library/Icingadb/ProvidedHook/Notifications/V2/Source.php
```diff
@@ -0,0 +1,149 @@
...
+                $remaining = '';
+            } elseif (str_starts_with($originalColumn, $strK . '.')) {
+                $remaining = substr($originalColumn, strlen($strK) + 1);
+            } elseif (str_starts_with($originalColumn, $strK . '[*]')) {
+                $remaining = substr($originalColumn, strlen($strK));
+            } else {
+                continue;
+            }
+
+            $keys = [...$keys, $strK];
```
BastianLedererIcinga: I think pushing onto `$keys` is more readable, the array spread is rather confusing.
```suggestion
            $keys[] = $strK;
```

library/Icingadb/ProvidedHook/Notifications/V2/Source.php
```diff
@@ -0,0 +1,149 @@
...
+            return $column; // the var no longer exists
+        }
+
+        $values = json_decode($var->value, true);
+        if (! is_array($values)) {
+            return $column; // already flat customvar e.g host.vars.app
+        }
+
+        $path = preg_replace('/^(?:host|service)\.vars\./', '', $column);
+        $path = substr($path, strlen($var->name) + 1);
```
BastianLedererIcinga: This is not reliable, take this example:
```
object Host "simple_array_host" {
    check_command = "dummy"
    address = "127.0.0.1"

    vars.simple_array = ["a", "b", "c"]
}
```
`$column` becomes `host.vars.simple_array[*]`

You strip the `host.vars.` and are left with `simple_array[*]` and now this line of code turns that to `*]`, because it assumes that the name it always followed by a dot.

I would suggest to include the `$var->name` and an optional dot after it in the `preg_replace`.

library/Icingadb/ProvidedHook/Notifications/V2/Source.php
```diff
@@ -0,0 +1,149 @@
...
+            ->onlyWithCustomVarSources(['host', 'service'])
+            ->withFixedColumns($this->allowedColumns);
+    }
+
+    public function getMetadataKeys(): array
+    {
+        return ['customvarid'];
+    }
+
+    private function flattenKeys(array $array, string $originalColumn, array $keys = []): array
```
BastianLedererIcinga: Even when fixing the handling of `$path` for the mentioned example, `flattenKeys` does not handle the case of `$originalColumn` being `[*]`, so it that case would need extra handling by the function or before calling it.

BastianLedererIcinga (commented):
This is what I noticed for the current state, but some of it may no longer apply once the changes regarding jsonPath are implemented.

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -126,11 +112,11 @@ protected function fetchValueSuggestions($column, $searchTerm, Filter\Chain $sea
         $query = $model::on($this->getDb());
         $query->limit(static::DEFAULT_LIMIT);
 
-        if (strpos($column, ' ') !== false) {
+        if (str_contains($column, ' ')) {
             // $column may be a label
-            list($path, $_) = Seq::find(
+            [$path, $_] = Seq::find(
                 $this->fixedColumns
-                    ?? QueryColumnsProvider::collectFilterColumns($query->getModel(), $query->getResolver()),
```
nilmerg, line 133: that's how my IDE indents by default?

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -139,57 +125,13 @@ protected function fetchValueSuggestions($column, $searchTerm, Filter\Chain $sea
             }
         }
```
nilmerg, line 140: Since you didn't take this over to the new implementation, and since my check back then only really supports English (or rather ignores the fact that labels not always include a space) this can be removed entirely. It's not a important functionality anyway. It also reduces the required visibility of `\Icinga\Module\Icingadb\Data\QueryColumnsProvider::collectFilterColumns` as this (and other similar usage) is the only case where it's required outside.

library/Icingadb/Data/QueryValuesProvider.php
```diff
@@ -0,0 +1,116 @@
...
+        if ($isCustomVar) {
+            $columnPath = $targetPath . '.flatvalue';
+            $this->query->filter(Filter::like($targetPath . '.flatname', $columnName));
+        }
+
+        $inputFilter = Filter::like($columnPath, $this->searchTerm);
+        $this->query->columns($columnPath);
+        $this->query->orderBy($columnPath);
+
+        // This had so many iterations, if it still doesn't work, consider removing it entirely :(
```
nilmerg: I think by now this comment is obsolete.

library/Icingadb/Data/QueryValuesProvider.php
```diff
@@ -0,0 +1,116 @@
...
+
+        $this->query->filter($this->searchFilter);
+
+        $this->applyRestrictions($this->query);
+        if ($this->hasBaseFilter()) {
+            $this->query->filter($this->getBaseFilter());
+        }
+
+        try {
+            return (new ObjectSuggestionsCursor($this->query->getDb(), $this->query->assembleSelect()->distinct()))
```
nilmerg: I think we should dismiss `\Icinga\Module\Icingadb\Util\ObjectSuggestionsCursor` now. The main reason it exists is because of the base class `\ipl\Sql\Cursor`. It allows pagination and generally limiting results. But none of those are really required to be controllable by the receiver of the traversable, are they? So the remaining thing left is why `ObjectSuggestionsCursor` exists, and this can be done here as well by turning this method into a generator.

nilmerg (changes requested):
Didn't look at the hook implementation yet.

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
sukhwinder33445: You are right, it was not indented correctly. But this code has been removed now.

library/Icingadb/Data/QueryValuesProvider.php
sukhwinder33445: Removed.

library/Icingadb/Data/QueryValuesProvider.php
```diff
@@ -0,0 +1,124 @@
...
+        foreach ($this->query->getDb()->yieldAll($select, PDO::FETCH_COLUMN) as $value) {
+            // TODO(lippserd): This is a quick and dirty fix for PostgreSQL binary datatypes for which PDO returns
+            // PHP resources that would cause exceptions since resources are not a valid type for attribute values.
+            // We need to do it this way as the suggestion implementation bypasses ORM behaviors here and there.
+            if (is_resource($value)) {
+                $value = stream_get_contents($value);
+            }
+
+            yield ['search' => $value, 'label' => $value];
+        }
```
nilmerg, line 108: I think there's no reason anymore to bypass the orm now. This was only done to be able to use the _cursed_ cursor implementation… So please just enable distinct on the select base and iterate over the orm query instead. In order to know which property to access, without inspecting `$columnPath` again, give it an alias in the above call to `columns()`.

library/Icingadb/ProvidedHook/Notifications/V2/Source.php
```diff
@@ -0,0 +1,217 @@
...
+
+    /** @var array<string, array<string, string[]>> The selected custom vars */
+    private array $selectedCustomVars = [];
+
+    public function __construct()
+    {
+        $this->allowedColumns = [
+            'host.name' => $this->translate('Host Name'),
+            'hostgroup.name' => $this->translate('Hostgroup Name'),
+            'host.user.name' => $this->translate('Contact Name'),
```
nilmerg: We're providing an integration for notifications here. Routing notifications based on the old contacts and contactgroups is rather unnecessary, don't you think? So we should probably remove them here, although this means the set of columns is different from how restrictions work. But we want users to migrate anyway so it doesn't help if they retain their old configs just to keep their rules working.

run.php
```diff
@@ -16,6 +16,7 @@
 
 if ($this::exists('notifications')) {
     $this->provideHook('Notifications/v1/Source');
+    $this->provideHook('Notifications/v2/Source');
```
BastianLedererIcinga: Notifications Web will expect the hook's type to be included in the name: https://github.com/Icinga/icinga-notifications-web/pull/469#discussion_r3234562254

So please make sure the hook's name is `Notifications/v2/Icinga2Source`, either by renaming the class or using the `$implementation` parameter of `provideHook`

library/Icingadb/Data/QueryValuesProvider.php
sukhwinder33445: Done.

library/Icingadb/ProvidedHook/Notifications/V2/Source.php
sukhwinder33445: Done.

library/Icingadb/Data/QueryValuesProvider.php
```diff
@@ -0,0 +1,125 @@
...
+            $targetPath = substr($targetPath, 0, -4) . 'customvar_flat';
+        }
+
+        if (str_contains($targetPath, '.')) {
+            try {
+                $this->query->with($targetPath); // TODO: Remove this, once ipl/orm does it as early
+            } catch (InvalidRelationException $e) {
+                throw new SearchException(sprintf($this->translate('"%s" is not a valid relation'), $e->getRelation()));
+            }
+        }
```
nilmerg: you can drop this now as well, but you still have to catch the exception but below in the foreach.

sukhwinder33445: Done.

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
```diff
@@ -0,0 +1,216 @@
...
+            'hostgroup.name' => $this->translate('Hostgroup Name'),
+            'service.name' => $this->translate('Service Name'),
+            'servicegroup.name' => $this->translate('Servicegroup Name')
+        ];
+    }
+
+    public function getSourceType(): string
+    {
+        return 'icinga2';
+    }
```
nilmerg: that's obsolete now

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
```diff
@@ -0,0 +1,216 @@
...
+
+                        do {
+                            $path = $this->createPath(array_splice($values, -1), [sprintf("['%s']", $varName)]);
+                            $path = $this->replacePrefix($prefix . implode('', $path));
+                            if (! in_array($path, $paths[$column] ?? [], true)) {
+                                $paths[$column][] = $path;
+                            }
+                        } while (! empty($values));
+                    }
+                }
```
nilmerg: I'm convinced this is a re-implementation of `CustomvarFlat::unFlattenVars()`, but I'm not quite sure as the code here is really hard to understand at first.

My quick attempt to at least verify `unFlattenVars` is a candidate showed success:

```php
                $traversable = function ($column) use ($prefix) {
                    foreach ($this->selectedCustomVars[$column] as $varName => $vars) {
                        yield (object) [
                            'flatname' => substr($column, strlen($prefix) + 1),
                            'flatvalue' => 'true',
                            'customvar' => (object) [
                                'name' => $varName,
                                'value' => reset($vars)
                            ]
                        ];
                    }
                };
                $struct = (new CustomvarFlat())->unFlattenVars($traversable($column));
```

Note that this is really dirty code, but `$struct` represents the path variations applicable to a single column and allows to be easily transformed to a set of json paths.

nilmerg (changes requested):

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
sukhwinder33445: Removed.

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
sukhwinder33445: This works. Thanks. I have a case where `unFlattenVars()` returns an unexpected result. I'm currently debugging it.

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    public function enrichCondition(Condition $condition): void
+    {
+        $column = $condition->getColumn();
+        if (isset($this->allowedColumns[$column])) {
+            $condition->metaData()->set('columnLabel', $this->allowedColumns[$column]);
+        } elseif (preg_match('/^(host|service)\.vars\.(.*)/i', $column, $m)) {
+            $prefix = $m[1] === 'host' ? $this->translate('Host') : $this->translate('Service');
```
nilmerg, line 102: This doesn't drop the `[*]` suffix like we usually do when re-rendering a condition.

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
```diff
@@ -0,0 +1,222 @@
...
+            $this->applyRestrictions($vars);
+
+            $found = false;
+            foreach ($vars as $var) {
+                $found = true;
+                $this->selectedCustomVars[$column][$var->name][$var->value] = $var->path;
+            }
+
+            if (! $found) {
+                throw new SearchException($this->translate('Given customvar does not exist yet, try later'));
```
nilmerg: Sounds a bit rude to me. There's surely a friendlier approach. Also, *customvar* is an unnecessary abbreviation, use *custom variable*.

What I mean with friendlier:
*try later*; either indicates the user doesn't know what he's doing or it incorrectly signals it may be resolved automatically
_does not exist **yet**_; is also not quite correct as you imply that it _will_ change.

Just state the fact that it isn't found and needs to be defined in order to be referenced in a rule.

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    public function assertValidCondition(Condition $condition): void
+    {
+        $column = $condition->getColumn();
+        if (isset($this->allowedColumns[$column])) {
+            return;
+        }
+
+        if (preg_match('/^(host|service)\.vars\.(.*)/i', $column, $m)) {
```
nilmerg: this must prevent columns with wildcards other than a trailing `[*]`.

nilmerg (changes requested):

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
sukhwinder33445: Done.

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
sukhwinder33445: Done.

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
sukhwinder33445: Done.

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
```diff
@@ -0,0 +1,220 @@
...
+            foreach ($this->selectedCustomVars[$column] as $varName => $vars) {
+                foreach ($vars as $value => $flatname) {
+                    yield (object) [
+                        'flatname' => $flatname,
+                        'flatvalue' => 'true',
+                        'customvar' => (object) [
+                            'name' => $varName,
+                            'value' => $value
+                        ]
+                    ];
```
nilmerg: Yield a proper model here please, i.e. `CustomvarFlat`. Once https://github.com/Icinga/ipl-orm/pull/161 is done you can pass just the arrays as-is to the constructor.

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
nilmerg: I think this should result in an error that describes the problem, rather than stating this is invalid. If you detect a wildcard, tell the user this is forbidden in custom variables.

nilmerg (changes requested):

library/Icingadb/Web/Control/SearchBar/ObjectSuggestions.php
```diff
@@ -126,70 +111,13 @@ protected function fetchValueSuggestions($column, $searchTerm, Filter\Chain $sea
...
-        $query->filter($searchFilter);
-        $this->applyBaseFilter($query);
-
-        try {
-            return (new ObjectSuggestionsCursor($query->getDb(), $query->assembleSelect()->distinct()))
-                ->setFetchMode(PDO::FETCH_COLUMN);
-        } catch (InvalidColumnException $e) {
-            throw new SearchException(sprintf(t('"%s" is not a valid column'), $e->getColumn()));
-        }
+        return $provider;
```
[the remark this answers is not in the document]
nilmerg: was always allowed, the consumer had to expect arrays prior as well

library/Icingadb/Data/QueryValuesProvider.php
```diff
@@ -0,0 +1,110 @@
...
+use ipl\Orm\Exception\InvalidColumnException;
+use ipl\Orm\Exception\InvalidRelationException;
+use ipl\Orm\Query;
+use ipl\Stdlib\BaseFilter;
+use ipl\Stdlib\Filter;
+use ipl\Web\Control\SearchBar\SearchException;
+use IteratorAggregate;
+
+/**
+ * Provide column and custom variable suggestions for a given query
```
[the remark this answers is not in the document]
sukhwinder33445: done.

nilmerg (approved):
Please rebase

library/Icingadb/ProvidedHook/Notifications/V2/Icinga2Source.php
```diff
@@ -0,0 +1,226 @@
...
+            if (preg_match('/^\[\d+]$/', $key)) {
+                $segments[] = '[*]';
+            } else {
+                $segments[] = $key;
+            }
+
+            $paths[] = implode('', $segments);
+        }
+
+        return $paths;
```
nilmerg:
```suggestion
            $paths[implode('', $segments)] = true;
        }

        return array_keys($paths);
```

nilmerg (changes requested):

nilmerg (approved):

nilmerg (approved):

PR #1370, opened 2026-04-29
Fix tactical overview export

### LINKED ISSUE #1334: json and csv export in tactical overview does not work
## Describe the bug
The Json / csv export does not work on tactical overview, it works in the monitoring module

## To Reproduce

1. login
2. click on overview-> tactical overview -> options (v) -> csv or json
3. or navigate to 
* https://demo.icinga.com/icingadb/tactical?format=csv
* https://demo.icinga.com/icingadb/tactical?format=json
4. get error

## Expected behavior
return valid json and consider the filters if set

## Your Environment

https://demo.icinga.com

nilmerg:
This introduces a new model `TacticalStateSummary` that represents a union of `HoststateSummary` and `ServicestateSummary`. The union of both is required to have the default export functionality also yield results for both types of states as it only evaluates the first query. Now, CSV/JSON exports also contain service states. Both can be differentiated either by looking at the column data types (`null` for `hosts_*` columns means it's a service state summary) or by simply checking the `type` column which is either `'host_state'` or `'service_state'`.

fixes #1334

### DIFF
--- a/application/controllers/TacticalController.php
+++ b/application/controllers/TacticalController.php
@@ -6,15 +6,13 @@
 namespace Icinga\Module\Icingadb\Controllers;
 
 use GuzzleHttp\Psr7\ServerRequest;
-use Icinga\Module\Icingadb\Model\HoststateSummary;
 use Icinga\Module\Icingadb\Model\ServicestateSummary;
+use Icinga\Module\Icingadb\Model\TacticalStateSummary;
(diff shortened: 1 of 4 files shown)


sukhwinder33445 (commented):
Works fine.

TacticalStateSummary: Please remove unused imports and add return types to the methods.

nilmerg:
> add return types to the methods.

Didn't add one to `::on()` as it calls the base implementation and has no control over the type.

sukhwinder33445 (approved):

Issue #1326, opened 2026-01-23
Custom variable containing dots is displayed incorrect


Virsacer:
## Describe the bug
Custom variable containing dots is displayed incorrect.
This was adressed in https://github.com/Icinga/icingadb-web/issues/757
But it is still partially present in current version...

## To Reproduce
The behaviour is different on two hosts, where the variable differs only by one letter:

    vars.vhosts = {"xxxxn-xxxx01.mgmt.xxxxxx.com"={http_port="4443"},}
    vars.vhosts = {"xxxxv-xxxx01.mgmt.xxxxxx.com"={http_port="4443"},}

The second host "v" also has a second variable in similar style (which is also displayed incorrectly):

    vars.sslcerts = {"wildcard_xxxxxx.com"={http_address="10.x.x.xxx"},"wildcard_xxxxxxxxxxxxxxxxxx.de"={http_address="10.x.x.xxx"},"wildcard_xxxxxxxxx.com"={http_address="10.x.x.xxx"},}

## Expected behavior
Display them like the first host:
![image](https://github.com/user-attachments/assets/dd950f42-4d54-46c6-9dde-97b1bb5c607a)

## Screenshots
Actual display of second host:
![image](https://github.com/user-attachments/assets/e3363b28-72fc-4370-b4f8-46e3c2cd6d40)

## Your Environment
* Icinga DB Web version (System - About): 1.3
* Icinga Web 2 version (System - About): 2.12.6
* Icinga 2 version (`icinga2 --version`): 2.15.1
* Icinga DB version (`icingadb --version`): 1.5.1
* PHP version used (`php --version`): 8.3.19
* Server operating system and version: Oracle Linux 9.7

nilmerg:
Hi, I cannot reproduce this. Using your config the second host looks like this:

<img width="672" height="730" alt="Image" src="https://github.com/user-attachments/assets/b10fe7fe-5c0f-4336-bddf-d1fb5084c529" />

Config:

```
object Host "first_host_with_vhosts" {
    import "test-host"

    vars.issue_no = 1326
    vars.vhosts = {"xxxxn-xxxx01.mgmt.xxxxxx.com"={http_port="4443"},}
}

object Host "second_host_with_vhosts" {
    import "test-host"

    vars.issue_no = 1326
    vars.vhosts = {"xxxxv-xxxx01.mgmt.xxxxxx.com"={http_port="4443"},}
[...]

PR #1375, opened 2026-05-20
`CustomvarFlat`: Don't ingnore special chars when splitting

sukhwinder33445:
Two issues were fixed in the split regex:

1. Special characters before `.` were ignored:
The old regex only split on `.` when directly preceded by a wordcharacter (\w), causing splits to be missed when `.` was preceded by special characters like whitespace, / or $.
Result before: `foo /.bar.one` => `['foo /.bar', 'one']`
Result after: `foo /.bar.one` => `['foo /', 'bar', 'one']`

2. Brackets mid-word were incorrectly split:
The old regex split before any `[`, causing keys like `con[0]cat` to be split incorrectly.
Now splits before `[` only when it's a standalone array index — meaning `]` is followed by `.`, another `[`, or end of string.
    1. Dot (`.`) : `disk[0].size => ['disk', '[0]', 'size']`
    2. Bracket(`[`): `disk[0][1] => ['disk', '[0]', '[1]']`
    3. String end: `disk[0] => ['disk', '[0]']`

    Result before: `con[0]cat` => `['con', '[0]cat']`
    Result after: `con[0]cat` => `['con', '[0]', 'cat']`

### DIFF
--- a/library/Icingadb/Model/CustomvarFlat.php
+++ b/library/Icingadb/Model/CustomvarFlat.php
@@ -185,7 +185,7 @@ class CustomvarFlat extends Model
                 $path = array_merge(
                     [$realName],
                     $sourcePath
-                        ? preg_split('/(?<=\w|])\.|(?<!^|\.)(?=\[)/', $sourcePath)
+                        ? preg_split('/(?<=.)\.|(?<!^|\.)(?=\[\d+](?:[.\[]|$))/', $sourcePath)
                         : []
                 );
             } else {
--- a/test/php/library/Icingadb/Model/CustomvarFlatTest.php
+++ b/test/php/library/Icingadb/Model/CustomvarFlatTest.php
@@ -53,7 +53,19 @@ class CustomvarFlatTest extends TestCase
         ["real_list[0]","one","real_list","[\"one\",\"two\",\"three\"]"],
         ["[1].2.[3].4.[5].6","123456","[1].2","{\"[3].4\":{\"[5].6\":123456}}"],
         ["ex.ample.com","cba","ex.ample.com","\"cba\""],
-        ["[4]","four","[4]","\"four\""]
+        ["[4]","four","[4]","\"four\""],
+        [
+            "disks.disk /.disk_partitions",
+            "result",
+            "disks",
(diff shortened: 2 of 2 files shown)


library/Icingadb/Model/CustomvarFlat.php
```diff
@@ -185,7 +185,7 @@ public function unFlattenVars(Traversable $flattenedVars): array
                 $path = array_merge(
                     [$realName],
                     $sourcePath
-                        ? preg_split('/(?<=\w|])\.|(?<!^|\.)(?=\[)/', $sourcePath)
+                        ? preg_split('/(?<=.)\.|(?<!^|\.)(?=\[)/', $sourcePath)
                         : []
```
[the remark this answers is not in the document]
sukhwinder33445: Added the test case.

library/Icingadb/Model/CustomvarFlat.php
```diff
@@ -185,7 +185,7 @@ public function unFlattenVars(Traversable $flattenedVars): array
                 $path = array_merge(
                     [$realName],
                     $sourcePath
-                        ? preg_split('/(?<=\w|])\.|(?<!^|\.)(?=\[)/', $sourcePath)
+                        ? preg_split('/(?<=.)\.|(?<!^|\.)(?=\[[\d*]+](?:[.\[]|$))/', $sourcePath)
```
nilmerg: `[\d*]+` -> `\d+`

nilmerg (changes requested):
This might also fix #1326 as a digit doesn't match `\w` either. Please verify and add a test case.

library/Icingadb/Model/CustomvarFlat.php
sukhwinder33445: Done.

sukhwinder33445:
> This might also fix #1326 as a digit doesn't match `\w` either. Please verify and add a test case.

`\w` also matches digits, so the mentioned issue will not be fixed by this PR. I was unable to reproduce it using the custom variables from the linked issue, with or without this PR.

nilmerg (approved):

PR #1377, opened 2026-05-21
QueryColumnsProvider: Fully use up limit with custom variables

nilmerg:
Previously custom variables that didn't match any of the desired sources were eating up the query's limit and so there is a chance that not even a single variable was listed. This especially occurs with the updated notifications integration, as the available database columns are very limited in number. But even in other views this may have been noticed by users in case they have many variables that are similarly named but have no overlap with database columns. (i.e. input that doesn't match any database columns but many vars)


nilmerg:
Tested myself with MySQL and PostgreSQL.

sukhwinder33445 (approved):
Works fine, tested on MySQL and PostgreSQL.

PR #1380, opened 2026-06-01
`QueryColumnsProvider`: Always quote the relation name

sukhwinder33445:
With https://github.com/Icinga/icingadb-web/pull/1377, we introduced a HAVING clause using the MAX function, which require quoting for the reserved keyword (user).

And as we produce sql queries by hand (with ipl-sql) and use strings that were not meant to be used this way, the relation name should always be quoted.

### DIFF
--- a/library/Icingadb/Data/QueryColumnsProvider.php
+++ b/library/Icingadb/Data/QueryColumnsProvider.php
@@ -424,6 +424,8 @@ class QueryColumnsProvider implements IteratorAggregate
                     );
                 }
 
+                $name = $this->getDb()->getAdapter()->quoteIdentifier($name);
+
                 $aggregates[$name] = new Expression("MAX($name)");
                 $scalarQueries[$name] = $select
                     ->resetColumns()->columns(new Expression('1'))


nilmerg (changes requested):
Good find! But this is a case where quoting needs to be done initially not during post-processing. What `Backend` does, related to postgres quoting, is only meant for cases where ipl/orm is responsible. The `MAX(user)` expression mentioned here isn't created using ipl/orm and hence needs to be quoted when created.

sukhwinder33445:
Previously, the selected `user` column also delivered wrong values. The query created by `QueryColumnsProvider::queryCustomvarConfig()` looked like the following:
```SELECT flatname, ... (MAX(user)) AS user, ...```
Which resulted in the `user` column returning the PostgreSQL user (`CURRENT_USER`) instead of the expected value:
Screenshot 2026-06-01 at 09 31 40

Now as the HAVING clause looks like `HAVING (MAX(user) = 1)`, this forces PostgreSQL to evaluate `text = integer`, making the type error explicit.

nilmerg (changes requested):
`ObjectAuthorization` does a similar thing but quotes always. Of course there it is about role names that may allow sql injection, but looking from the perspective that we produce sql queries *by hand* (with ipl-sql) and use strings that were not meant to be used this way, I think this should also quote always.

nilmerg (approved):

PR #1388, opened 2026-06-25
Fix `SearchControls` trait

BastianLedererIcinga:
https://github.com/Icinga/icingadb-web/pull/1046 introduced two problems to the `SearchControls` trait.

Returning `false` from `SearchControls::callHandleRequest()` prevents the ipl-web `SearchControls` trait from calling `handleRequest()` in both `createSearchBar()` and `createSearchEditor()`.
An implementation of `createSearchEditor()` is added, which explicitly calls `$editor->handleRequest()` as `createSearchBar()` already does. Without this no `SearchEditor` in the entire module is usable.

An other problem is that moving the call of `$searchbar->handleRequest()` also delayed its `assemble()`, so this if block was entered before the assemble, resulting in the wrapper being added twice.

```
if (($wrapper = $searchBar->getWrapper()) && ! $wrapper->getWrapper()) {
    // TODO: Remove this once ipl-web v0.7.0 is required
    $searchBar->addWrapper(Html::tag('div', ['class' => 'search-controls']));
}
```
which caused a visual bug:

Screenshot from 2026-06-25 11-01-42

The statement should be safe to remove together with the less that has the same `TODO`.

### DIFF
--- a/library/Icingadb/Common/SearchControls.php
+++ b/library/Icingadb/Common/SearchControls.php
@@ -9,6 +9,7 @@ use Icinga\Module\Icingadb\Hook\SearchColumnsProviderHook;
 use ipl\Html\Html;
 use ipl\Orm\Query;
 use ipl\Web\Control\SearchBar;
+use ipl\Web\Control\SearchEditor;
 use ipl\Web\Url;
 use ipl\Web\Widget\ContinueWith;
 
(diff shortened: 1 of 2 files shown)


nilmerg (approved):
Thanks. :)

PR #1392, opened 2026-07-15
Allow to manage and subscribe to incidents

### LINKED ISSUE #1358: Allow to manage and subscribe to Icinga Notification Incidents
The host and service object detail should [integrate some of Icinga Notifications Web's features](https://github.com/Icinga/icinga-notifications-web/issues/464).

### Manage and Subscribe

They're very similar to the other quick actions so let's add them there. Though, Manage/Unmanage are somewhat redundant compared with Acknowledge/Remove Acknowledgement. To keep with the terms the user is mostly familiar with, acknowledging a host/service means also managing its incident. So only Subscribe/Unsubscribe are additional quick actions.

#### Permissions

As with other quick actions, the subscribe/unsubscribe action should be covered by a new permission.

### Recipients

The current *Notifications* section and the current recipients in Icinga Notifications are _somewhat_ similar. *Notifications* lists contacts and groups that *may* receive notifications, *Recipients* lists contacts that *did* receive notifications.

There is also the chance that both are configured. Both cases mean we cannot simply show recipients where contacts and groups currently show up. Either because they'd collide with the current configuration or the user misinterprets their meaning.

A mockup by @flourish86 will follow, showing possible solutions to this.

BastianLedererIcinga:
resolve #1358
requires https://github.com/Icinga/icinga-notifications-web/pull/545

Add a new `Subscribe` button to `QuickActions`, to that allows to subscribe to the incident matching the Host/Service.

The existing Acknowledge/Unacknowledge option also serves as manage/unmanage of the matching incident.

Users that have no notification contact configured will see an error message when trying to subscribe, and a warning when acknowledging.

A new `Notification Recipients` section is added to both Host- and Sevicedetail, showing the current recipients and subscribers. The manager is highlighted with the redesigned ack badge suggested in https://github.com/Icinga/icingadb-web/issues/1358#issuecomment-4682061584.


sukhwinder33445:
Please mention the required PR https://github.com/Icinga/icinga-notifications-web/pull/472 in the description.

library/Icingadb/Notifications/SubscriptionForm.php
```diff
@@ -0,0 +1,80 @@
...
+            try {
+                foreach (IncidentFinder::forObject($this->object) as $incident) {
+                    if ($this->subscribe) {
+                        $incident->addSubscriber($username);
+                    } else {
+                        $incident->removeSubscriber($username);
+                    }
+                }
+            } catch (InvalidArgumentException) {
+                Notification::error(
```
sukhwinder33445, line 67: Please add a sucess message, other QuickActions have it.

library/Icingadb/Notifications/ManagesIncidents.php
```diff
@@ -0,0 +1,50 @@
...
+                        $incident->removeManager($username);
+                    }
+                }
+            }
+        } catch (InvalidArgumentException) {
+            if ($manage) {
+                Notification::warning(
+                    t('Cannot manage matching incident, no notification contact configured for your account.')
+                );
+            }
```
sukhwinder33445: Add a sucess message please.

library/Icingadb/Common/HostLinks.php
```diff
@@ -74,4 +74,14 @@ public static function services(Host $host): Url
...
+
+    public static function subscribe(Host $host): Url
+    {
+        return Url::fromPath('icingadb/host/subscribe', ['name' => $host->name]);
+    }
+
+    public static function unsubscribe(Host $host): Url
+    {
+        return Url::fromPath('icingadb/host/unsubscribe', ['name' => $host->name]);
+    }
```
sukhwinder33445, line 86: These methods are not used.

library/Icingadb/Common/ServiceLinks.php
```diff
@@ -106,4 +106,14 @@ public static function toggleFeatures(Service $service, Host $host): Url
...
+
+    public static function subscribe(Service $service, Host $host): Url
+    {
+        return Url::fromPath('icingadb/service/subscribe', ['name' => $service->name, 'host.name' => $host->name]);
+    }
+
+    public static function unsubscribe(Service $service, Host $host): Url
+    {
+        return Url::fromPath('icingadb/service/unsubscribe', ['name' => $service->name, 'host.name' => $host->name]);
+    }
```
sukhwinder33445, line 118: These methods are not used.

library/Icingadb/Notifications/SubscribeIncidents.php
```diff
@@ -0,0 +1,46 @@
...
+    protected function handleSubscription(bool $subscribe, Host|Service $object, Url $redirectUrl): void
+    {
+        if (
+            Icinga::app()->getModuleManager()->hasEnabled('notifications')
+            && $this->isGrantedOn('icingadb/notifications/subscribe', $object)
+        ) {
+            $form = (new SubscriptionForm($subscribe, $object))
+                ->setAction((string) Url::fromRequest())
+                ->on(Form::ON_SUBMIT, function () use ($redirectUrl) {
+                    $this->getResponse()->setAutoRefreshInterval(1);
```
sukhwinder33445: You redirect immediately, i think this is superfluous.

library/Icingadb/Notifications/SubscriptionForm.php
```diff
@@ -0,0 +1,80 @@
...
+     * @param Host|Service $object The object whose incidents should be subscribed/unsubscribed
+     */
+    public function __construct(bool $subscribe, Host|Service $object)
+    {
+        $this->subscribe = $subscribe;
+        $this->object = $object;
+        $this->setAttribute(
+            'title',
+            $subscribe
+                ? t('Subscribe to the related incident')
```
sukhwinder33445: Always use `Translation` trait in forms please.

library/Icingadb/Widget/Detail/QuickActions.php
```diff
@@ -48,11 +51,51 @@ protected function assemble()
...
+                }
+
+                if ($role !== 'manager') {
+                    $action = $role === 'subscriber' ? 'unsubscribe' : 'subscribe';
+
+                    $this->addHtml(
+                        Html::tag(
+                            'li',
+                            (new SubscriptionForm($action === 'subscribe', $this->object)
+                            )->setAction($this->getLink($action))
```
sukhwinder33445: Fix indentation please.

sukhwinder33445 (changes requested):
Please fix the alignment of the new ack badge.

Image

library/Icingadb/Common/HostLinks.php
BastianLedererIcinga: They are used here:
https://github.com/Icinga/icingadb-web/blob/0f0683c8f8cf72436cf377b22f4cb7d37da1f4de/library/Icingadb/Widget/Detail/QuickActions.php#L144-L151

Our IDEs just don't understand it because it is done via $action.

sukhwinder33445:
One more thing I noticed: once you subscribe to and then unsubscribe from an incident in the incident detail view, subscribing again is not possible and throws an error, because the unsubscribe removes the incident_contact DB entry entirely.
Using the `QuickAction` introduced with this PR always works.

BastianLedererIcinga:
> One more thing I noticed: once you subscribe to and then unsubscribe from an incident in the incident detail view, subscribing again is not possible and throws an error, because the unsubscribe removes the incident_contact DB entry entirely. Using the `QuickAction` introduced with this PR always works.

As discussed in person, this an unrelated bug, caused by the `IncidentQuickActions` not setting the `changed_at` column.

library/Icingadb/Notifications/SubscriptionForm.php
```diff
@@ -0,0 +1,88 @@
...
+                );
+            } catch (InvalidArgumentException) {
+                Notification::error(
+                    $this->translate('Cannot subscribe, there is no notification contact configured for your account.')
+                );
+            }
+        });
+    }
+
+    protected function assemble()
```
sukhwinder33445: Add return type please.

sukhwinder33445 (commented):
LGTM

sukhwinder33445 (approved):

library/Icingadb/Notifications/SubscribeIncidents.php
```diff
@@ -0,0 +1,45 @@
...
+     * @return void
+     */
+    protected function handleSubscription(bool $subscribe, Host|Service $object, Url $redirectUrl): void
+    {
+        if (
+            Icinga::app()->getModuleManager()->hasEnabled('notifications')
+            && $this->isGrantedOn('icingadb/notifications/subscribe', $object)
+        ) {
+            $form = (new SubscriptionForm($subscribe, $object))
+                ->setAction((string) Url::fromRequest())
```
nilmerg: If I'm not mistaken, an action is only mandatory for modal forms, and this isn't part of a modal.

library/Icingadb/Notifications/SubscribeIncidents.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+trait SubscribeIncidents
+{
+    /**
+     * Subscribe or unsubscribe all incidents matching the given tags
+     *
+     * @param bool $subscribe
+     * @param Host|Service $object
+     * @param Url $redirectUrl The URL to redirect to after the form has been handled
+     * @return void
```
nilmerg, line 28: empty line before `@return`

library/Icingadb/Notifications/ManagesIncidents.php
```diff
@@ -0,0 +1,66 @@
...
+                Notification::success(sprintf(
+                    $manage
+                        ? tp('Now managing the matching incident', 'Now managing %d matching incidents', $count)
+                        : tp(
+                            'No longer managing the matching incident',
+                            'No longer managing %d matching incidents',
+                            $count
+                        ),
+                    $count
+                ));
```
nilmerg: This renders now two success messages in a row which is sub-optimal. But I think we should rather describe what *managing an incident* actually means anyway. But not in a success message either. So please drop this.

library/Icingadb/Notifications/ManagesIncidents.php
```diff
@@ -0,0 +1,66 @@
...
+            }
+        } catch (InvalidArgumentException) {
+            if ($manage) {
+                Notification::warning(
+                    t('Cannot manage matching incident, no notification contact configured for your account.')
+                );
```
nilmerg: I'm more and more convinced that we should provide a way to verify this prior. The problem now is that you will be able to acknowledge the problem in Icinga 2 but the *incident* keeps unchanged. We don't know whether this is intentional right now, so let's give us this option!

To un-/manage, the following conditions must be met:
* The current user must be a notifications contact (https://github.com/Icinga/icinga-notifications-web/issues/529)
* The current user must have the permission `icingadb/notifications/manage`

So in case the user has the permission but there's no contact, it shouldn't be possible to acknowledge or subscribe. The particular actions in the object detail should be disabled and provide a title to describe why.

library/Icingadb/Notifications/ManagesIncidents.php
```diff
@@ -0,0 +1,66 @@
...
+    protected function manageIncidents(bool $manage): void
+    {
+        if (! Icinga::app()->getModuleManager()->hasEnabled('notifications')) {
+            return;
+        }
+
+        $username = $this->getAuth()->getUser()->getUsername();
+        $count = 0;
+        try {
+            foreach ($this->getObjects() as $object) {
```
nilmerg: for bulk actions, this does nothing as the underlying query does not yield any result at this stage.

nilmerg (changes requested):
Looking at this I noticed a flaw in my previous idea how the new component is integrated here. It relies on a few presumptions that are either incompatible or simply wrong:

* The user acknowledging needs a matching notifications contact
  * This is not always true and we have no idea to identify this as intentional or not, an additional permission should solve this
* The acknowledgement form still allows to declare an expiry time
  * But there's no way for Icinga DB to remove a manager, https://github.com/Icinga/icinga-notifications/issues/478
* The acknowledgement form still allows to disable the sticky flag
  * Let's hide this option for users that have the new manage permission and are a contact in notifications
* Enabling icinga notifications web does not mean it's configured, but this should be a requirement and be easy to answer to decide what to offer
  * The module being enabled is not enough, [db web knows about a working notifications component](https://github.com/Icinga/icingadb/issues/1141), so this should be the condition instead
* *incident* is already an outdated term so I don't want to see it here. It is also of no meaning to users which have no access to the configuration UI anyway, so we should rather focus on proper representation what the effect of acknowledge now is and what subscribe means compared to the legacy notifications feature. A seamless integration so to say, without the need for the user to know anything about Icinga Notifications.

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -414,6 +423,94 @@ protected function createNotifications(): array
...
+                        )
+                    );
+                }
+
+                $subscribers[] = $subscriberItem;
+            }
+
+            foreach ($incident->getRecipients() as $recipient) {
+                $icon = new Icon($recipientIcons[$recipient['type']]);
+                $name = $recipient['full_name'] ?? $recipient['name'];
```
nilmerg: What this, and the loop over the subscribers above, shows me is that the only thing of interest here is the `full_name` array key. It doesn't make much sense to yield an array key in only 1 of 3 cases with a value set. Let me suggest a minor but breaking change now, in order to get this straight before the stable release or other integrations will be worked on: https://github.com/Icinga/icinga-notifications-web/issues/531

application/forms/Command/Object/AcknowledgeProblemForm.php
```diff
@@ -117,62 +140,64 @@ protected function assembleElements()
...
-                'description'   => t('If the acknowledgement should expire, check this option.')
-            ]
-        );
-        $decorator->decorate($this->getElement('expire'));
-
-        if ($this->getElement('expire')->isChecked()) {
-            $expireTime = new DateTime();
-            $expireTime->add(new DateInterval($config->get('settings', 'acknowledge_expire_time', 'PT1H')));
-
+        if (! $canManage) {
```
nilmerg: I forgot to mention it as it can't get more obvious: The *Send Notification* toggle is also obsolete in this case. 🤣

library/Icingadb/Common/Backend.php
```diff
@@ -196,4 +203,40 @@ public static function supportsDependencies(): bool
...
+     * Whether Icinga DB reports a healthy Icinga Notifications component
+     *
+     * @return bool
+     */
+    public static function notificationsHealthy(): bool
+    {
+        if (self::$notificationsHealthy === null) {
+            self::$notificationsHealthy = self::supportsNotifications()
+                && Instance::on(self::getDb())
+                    ->columns('notifications_healthy')
```
nilmerg, line 236:
```suggestion
                    ->columns('notifications_healthy')
                    ->filter(Filter::equal('responsible', true'))
```

configuration.php
nilmerg, line 1: I believe the descriptions for the new permissions should mention the new component and that it's a prerequisite. They are configured by administrators and they should know about it.

library/Icingadb/Widget/Detail/MultiselectQuickActions.php
```diff
@@ -132,12 +135,40 @@ protected function assemble()
...
+                )
+            ) {
+                $title = t('Acknowledge this problem and tag it as being handled');
+            } elseif (Incidents::canManage($this->getAuth()->getUser())) {
+                $title = t(
+                    'Acknowledge this problem and tag it as being handled, so that only you and anyone'
+                    . ' who subscribes will receive its notifications'
+                );
+            } else {
+                $disabled = true;
```
nilmerg: We should also disable the remove ack button, as the contact's username might be gone or changed in the meantime.

library/Icingadb/Widget/Detail/MultiselectQuickActions.php
```diff
@@ -132,12 +135,40 @@ protected function assemble()
...
+                    . 'being handled');
+            } elseif (
+                ! $this->isGrantedOnType(
+                    'icingadb/notifications/manage',
+                    $this->type,
+                    $this->getBaseFilter(),
+                    false
+                )
+            ) {
+                $title = t('Acknowledge this problem and tag it as being handled');
```
nilmerg: This should be same title as if the component isn't available. It's as if it wouldn't after all, in case the permission is not granted.

library/Icingadb/Widget/Detail/QuickActions.php
```diff
@@ -44,12 +49,85 @@ protected function assemble()
...
+                            );
+                        } else {
+                            $disabled = true;
+                            $title = t(
+                                'You cannot acknowledge this problem, as there is no Icinga Notifications contact'
+                                . ' configured for your account'
+                            );
+                        }
+                    } else {
+                        $title = t('Acknowledge this problem and tag it as being handled');
```
nilmerg: same here, should be the same as if the component isn't available

library/Icingadb/Widget/Detail/QuickActions.php
```diff
@@ -44,12 +49,85 @@ protected function assemble()
...
+                        $this->addHtml(
+                            Html::tag(
+                                'li',
+                                (new SubscriptionForm($action === 'subscribe', $this->object))
+                                    ->setAction($this->getLink($action))
+                            )
+                        );
+                    }
+                }
             }
```
nilmerg, line 127: This looks like another candidate to add to Icinga Notifications Web instead: A way to get the current roles of a user. If it returns null, there is no incident to subscribe to. Why role**s**? Because while for Icinga DB there can only be one incident right now, this might not be true for other sources. I'll prepare something first myself, then adjust this please.

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -399,8 +400,17 @@ protected function createNotifications(): array
             );
         }
 
+        if (
+            Icinga::app()->getModuleManager()->hasEnabled('notifications')
+            && Backend::notificationsHealthy()
+            && ! $userList->hasContent()
+            && ! $usergroupList->hasContent()
```
nilmerg: Please check this first. It's a minor optimization, but there's no point to query the database if either of the lists have content.

nilmerg (changes requested):

nilmerg:
Please rebase with https://github.com/Icinga/icingadb-web/pull/1404 and create a stack.

library/Icingadb/Widget/Detail/QuickActions.php
nilmerg: https://github.com/Icinga/icinga-notifications-web/pull/540

library/Icingadb/Common/Backend.php
```diff
@@ -196,4 +204,41 @@ public static function supportsDependencies(): bool
...
+
+        return self::$supportsNotifications;
+    }
+
+    /**
+     * Whether Icinga DB reports a healthy Icinga Notifications component
+     *
+     * @return bool
+     */
+    public static function notificationsHealthy(): bool
```
nilmerg: Had to reconsider what I previously wrote. The component being *healthy* is sure a nice thing to have, but it's not a requirement to manage or subscribe. The component being **configured** is what's required. That's not just a slight difference in wording, it means that *acknowledging* or *subscribing* is not dependent at all on a *healthy* component. Icinga Notifications Web is responsible to designate these roles and for this only the database needs to be working.

- Rename this to `notificationsSetUp`
- Also check whether the module is enabled, since both checks are always coupled right now
- Add a todo to check instead what `icingadb_config` has to offer (I don't want this to depend on https://github.com/Icinga/icingadb/issues/1108 right now), for now it's still fine to check the healthiness

application/forms/Command/Object/AcknowledgeProblemForm.php
```diff
@@ -55,18 +62,33 @@ public function __construct()
 
     protected function assembleElements()
     {
+        $canManage = Icinga::app()->getModuleManager()->hasEnabled('notifications')
+            && Backend::notificationsHealthy()
+            && $this->getAuth()->hasPermission('icingadb/notifications/manage')
+            && Incidents::canManage($this->getAuth()->getUser());
```
nilmerg: The user being able to manage (`Incidents::canManage`) is not a reason to decide what options to show, the user cannot normally visit this form in this case after all. (The action is disabled)

So, let's drop this condition. If it'd be false, submitting the form shows an error anyway.

nilmerg: Oh and again, please check the cheapest things first in such cases.

1. permission
2. hasEnabled
3. db stuff

So that PHP can early abort and wont needlessly evaluate costly conditions.

application/forms/Command/Object/AcknowledgeProblemForm.php
```diff
@@ -208,8 +232,8 @@ protected function getCommands(Iterator $objects): Traversable
         $command = new AcknowledgeProblemCommand();
         $command->setComment($this->getValue('comment'));
         $command->setAuthor($this->getAuth()->getUser()->getUsername());
-        $command->setNotify($this->getElement('notify')->isChecked());
-        $command->setSticky($this->getElement('sticky')->isChecked());
+        $command->setNotify(! $this->hasElement('notify') || $this->getElement('notify')->isChecked());
```
nilmerg: I'm not sure it is okay to default to `true` here in case the element is missing. It implies that Icinga 2 should still transmit notifications to legacy recipients?

library/Icingadb/Notifications/ManagesIncidents.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param bool $manage
+     *
+     * @return void
+     */
+    protected function manageIncidents(bool $manage): void
+    {
+        if (
+            Icinga::app()->getModuleManager()->hasEnabled('notifications')
+            && Backend::notificationsHealthy()
+            && Incidents::canManage($this->getAuth()->getUser())
```
nilmerg: Shouldn't this rather check the permission? If it comes this far and there's no contact available I'd expect an error.

I know the permission is being checked already in the loop, but for each object. It's perfectly fine to evaluate it first in a global context as if that's false, it won't be true otherwise anyway.

library/Icingadb/Notifications/SubscribeIncidents.php
nilmerg, line 1: I just wondered, why isn't this action available for bulk processing?

nilmerg (changes requested):

library/Icingadb/Notifications/SubscribeIncidents.php
BastianLedererIcinga: I think I initially didn't include it because prior to the addition of `Incidents::canSubscribe()`, it would have been quite a mess, and I never reconsidered it after that.

But while implementing it I realized that guarding the presence of the subscribe/unsubscribe is still a problem.
In `QuickActions` we check if a matching incident exists, but doing the same in `MultiSelectQuickActions` would need to run that check once per selected object, just to know which button(s) to show, so it is currently guarded by the presence of problems, which can lead to subscribe/unscubscribe buttons that have no effect when pressed.

application/forms/Command/Object/AcknowledgeProblemForm.php
BastianLedererIcinga: Fair point, since the description states 

> only you and anyone who subscribes to it will receive its notifications.

It probably makes sense to default to `false`

BastianLedererIcinga:
The added `SubscribeIncidents` in `RedundancygroupController` can't be tested without local modifications, because the controller is completely broken due to recent changes, I will create seperate issues for these problems.

When testing subscribe and unsubscribe in a detail view, the subscribe button may not change to unsubscribe. This is a bug in `Incidents::getRoles()`, https://github.com/Icinga/icinga-notifications-web/pull/541 should fix it.

library/Icingadb/Common/Backend.php
```diff
@@ -196,4 +205,44 @@ public static function supportsDependencies(): bool
...
+            } else {
+                self::$supportsNotifications = self::getDbSchemaVersion() >= 8;
+            }
+        }
+
+        return self::$supportsNotifications;
+    }
+
+    /**
+     * Whether Icinga DB reports a healthy Icinga Notifications component
```
nilmerg: This does not represent the intended result yet.

library/Icingadb/Notifications/IncidentFinder.php
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Find the open incidents related to the given object
+     *
+     * @param Host|Service $object
+     *
+     * @return Incidents
+     */
+    public static function forObject(Host|Service $object): Incidents
+    {
+        return Incidents::find(self::buildTags($object));
```
nilmerg: We should rename `Incidents::find(…)` to  `Incidents::get(array $tags): ?Incident`.

Regarding the bulk processing actions, I'd suggest to introduce `Incidents::matchAll(Traversable<array>): ?static` and `Incidents::getAll(…)`. `matchAll` can be used to signal the intent to perform partial matches.

`IncidentFinder::forObjects(…): ?Incidents` is introduced to use `getAll`.

library/Icingadb/Notifications/ManagesIncidents.php
```diff
@@ -0,0 +1,42 @@
...
+    protected function manageIncidents(bool $manage): void
+    {
+        if ($this->getAuth()->hasPermission('icingadb/notifications/manage') && Backend::notificationsSetUp()) {
+            $username = $this->getAuth()->getUser()->getUsername();
+            foreach ($this->getObjects() as $object) {
+                if (! $this->isGrantedOn('icingadb/notifications/manage', $object)) {
+                    continue;
+                }
+
+                foreach (IncidentFinder::forObject($object) as $incident) {
```
nilmerg: You should use `IncidentFinder::forObjects` here and pass it a `CallbackFilterIterator` that limits the objects to those which match the permission.

library/Icingadb/Notifications/SubscriptionForm.php
```diff
@@ -0,0 +1,115 @@
...
+        $this->subscribe = $subscribe;
+        $this->objects = $objects;
+        $this->setAttribute(
+            'title',
+            $subscribe
+                ? $this->translate('Receive notifications about this problem')
+                : $this->translate('Stop receiving notifications about this problem')
+        );
+
+        $this->on(self::ON_SUBMIT, function () {
```
nilmerg: I think this fits better inside the trait. Though, notifications should still be triggered here and the count of objects can be identified by … well, counting them, just like the command forms do.

library/Icingadb/Widget/Detail/MultiselectQuickActions.php
```diff
@@ -123,6 +126,14 @@ protected function assemble()
         $activeChecks = "{$this->getColumnPrefix()}_active_checks_enabled";
         $passiveChecks = "{$this->getColumnPrefix()}_passive_checks_enabled";
 
+        $affectsIncidents = Backend::notificationsSetUp()
+            && $this->isGrantedOnType(
+                'icingadb/notifications/manage',
+                $this->type,
+                $this->getBaseFilter(),
+                false
+            );
```
nilmerg: Remember the order of conditions ;)

library/Icingadb/Widget/Detail/MultiselectQuickActions.php
```diff
@@ -149,12 +177,45 @@ protected function assemble()
...
+
+                $this->add(Html::tag('li', $removeAckForm));
+            }
+        }
 
-            $this->add(Html::tag('li', $removeAckForm));
+        if (
+            $this->summary->$acks + $this->summary->$unacknowledged > 0
+            && Backend::notificationsSetUp()
+            && $this->isGrantedOnType('icingadb/notifications/subscribe', $this->type, $this->getBaseFilter(), false)
```
nilmerg: same here ;)

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -414,6 +423,96 @@ protected function createNotifications(): array
...
     }
 
+    protected function createNotificationRecipients(): array
+    {
+        if (! Backend::notificationsSetUp()) {
+            return [];
+        }
+
+        // TODO: The query notifications-web creates here can take several seconds for large amount of incidents,
+        //       it should be optimized
```
nilmerg: That's covered by https://github.com/Icinga/icinga-notifications-web/issues/528

library/Icingadb/Widget/Detail/QuickActions.php
```diff
@@ -34,22 +40,84 @@ public function __construct($object)
 
     protected function assemble()
     {
+        $affectsIncidents = Backend::notificationsSetUp()
+            && $this->isGrantedOn('icingadb/notifications/manage', $this->object);
```
nilmerg: again, order please, https://github.com/Icinga/icingadb-web/pull/1392#discussion_r3764419220 was about all conditions like this.

nilmerg (changes requested):

nilmerg:
Please also extend `doc/05-Upgrading.md` and explain the way both, new and legacy, configurations are supported. Also document the migration path I'd recommend as just discussed.

library/Icingadb/Notifications/SubscriptionForm.php
BastianLedererIcinga: Only now while rereading your comments I realized that I moved more to the trait than you asked for. But I actually don't see any advantage in keeping the notification in the form. It removes the distinction between the count of problems vs. the count of incidents that could actually be subscribed.

library/Icingadb/Notifications/IncidentFinder.php
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Get the matching incident for each of the given objects
+     *
+     * @param iterable<Host|Service> $objects
+     *
+     * @return Incidents
+     */
+    public static function forObjects(iterable $objects): Incidents
+    {
+        return Incidents::getAll(Seq::map($objects, fn (Host|Service $object) => self::buildTags($object)));
```
nilmerg: please don't insert a space after `fn`

library/Icingadb/Notifications/SubscriptionForm.php
nilmerg: The reason is the distinction between rendering and processing. All render related parts are part of the form, while processing logic is part of the trait. In case the rendering needs to change, I'd look in the form first and then wonder where the notification is being triggered.

nilmerg (changes requested):

library/Icingadb/Notifications/IncidentFinder.php
```diff
@@ -0,0 +1,66 @@
...
+use Icinga\Module\Notifications\Integrations\Incidents;
+use ipl\Stdlib\Seq;
+
+class IncidentFinder
+{
+    /**
+     * Find the open incident related to the given object
+     *
+     * If no incident for the object is found, calling any function on the returned instance will throw an
+     * {@see IncidentNotFoundException}
```
nilmerg: Document this using `@throws`

library/Icingadb/Notifications/SubscribeIncidents.php
```diff
@@ -0,0 +1,91 @@
...
+                            }
+                        }
+                    } catch (InvalidArgumentException) {
+                        Notification::error(
+                            $this->translate(
+                                'Cannot subscribe, as there is no Icinga Notifications contact configured'
+                                . ' for your account.'
+                            )
+                        );
+                    }
```
nilmerg: That's already guarded by the disabled button, isn't it? I see no reason to *fail softly* here.

nilmerg (changes requested):
Still have to test this, though. 😩 Didn't look at the UI in detail as well yet.

library/Icingadb/Notifications/SubscribeIncidents.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    protected function handleSubscription(bool $subscribe, array|Query $objects, Url $redirectUrl): void
+    {
+        if (Backend::notificationsSetUp()) {
+            $form = (new SubscriptionForm($subscribe))
+                ->setObjectCount(count($objects))
+                ->on(Form::ON_SUBMIT, function () use ($subscribe, $objects, $redirectUrl) {
+                    $username = $this->getAuth()->getUser()->getUsername();
+                    $granted = new CallbackFilterIterator(
+                        is_array($objects) ? new ArrayIterator($objects) : $objects->getIterator(),
```
nilmerg: Actually, I'd prefer the previously used `IteratorIterator`. `getIterator` may be valid, but that's only an implementation detail that is responsible for `Query` being `Traversable`, and shouldn't be relied upon.

public/css/widget/notification-recipients.less
nilmerg, line 1: Just noticed…

Image

Compare this with the mockup in the issue and you'll see that the *configured contacts* have no pill whatsoever and also no comma. Don't change anything now, I want first @flourish86's opinion on what to settle, as I'm unsure if it's fine to change the legacy section visually now.

nilmerg (changes requested):

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -414,6 +424,94 @@ protected function createNotifications(): array
...
+        if (! Backend::notificationsSetUp()) {
+            return [];
+        }
+
+        $incident = IncidentFinder::forObject($this->object);
+
+        $subscribers = [];
+        $recipients = [];
+        try {
+            foreach ($incident->getSubscribers() as $subscriber) {
```
nilmerg: And, once `getSubscribers()` and `getRecipients()` do not return a generator anymore, please only wrap `getSubscribers()` in a try…catch block, and not the entire loop.

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -414,6 +424,96 @@ protected function createNotifications(): array
...
+        }
+
+        $incident = IncidentFinder::forObject($this->object);
+
+        $subscriberList = [];
+        $recipientList = [];
+        try {
+            $subscribers = $incident->getSubscribers();
+        } catch (IncidentNotFoundException) {
+            // No incident for this object was found
```
nilmerg: *irony on* A developer added an inline comment *irony off*

public/css/widget/notification-recipients.less
flourish86: Just discussed this with @nilmerg, which clarified a few things.

Please leave the Contact elements as they are. The mockup came from my wrong assumption, that they currently look this way.

For the Recipients it would be better to use the design from Variant A (notice the comma being @text-color-light). It differentiates them better from Contact elements, which open a detail view, when you click them.

By Contrast clicking a Recipient element has no effect.

[The post with the mockup](https://github.com/Icinga/icingadb-web/issues/1358#issuecomment-4682061584)

library/Icingadb/Notifications/SubscribeIncidents.php
```diff
@@ -0,0 +1,81 @@
...
+        if (Backend::notificationsSetUp()) {
+            $form = (new SubscriptionForm($subscribe))
+                ->setObjectCount(count($objects))
+                ->on(Form::ON_SUBMIT, function () use ($subscribe, $objects, $redirectUrl) {
+                    $username = $this->getAuth()->getUser()->getUsername();
+                    $granted = new CallbackFilterIterator(
+                        is_array($objects) ? new ArrayIterator($objects) : new IteratorIterator($objects),
+                        fn($object) => $this->isGrantedOn('icingadb/notifications/subscribe', $object)
+                    );
+                    foreach (IncidentFinder::forObjects($granted) as $incident) {
```
nilmerg: Guard this once it throws similar to https://github.com/Icinga/icingadb-web/blob/2da207bedd5a4e5b081f9e6c5c157220695d00b1/application/forms/Command/Object/AcknowledgeProblemForm.php#L213-L214

public/css/widget/notification-recipients.less
nilmerg: ^ this

nilmerg (changes requested):

public/css/widget/notification-recipients.less
flourish86: Another thing came up:

(Active) Subscribers are in this case always single Icinga Web users, so that's why another differentiation between Recipient and active subscribers make sense (no icon). They have in common, that they're both not clickable, so the "text"-based visualization for both makes sense.

All in all this is what it should finally look like.

Frame 31

Please also note that the issue suggests tooltips for clarifying the meaning.

library/Icingadb/Widget/Detail/ObjectDetail.php
```diff
@@ -414,6 +424,95 @@ protected function createNotifications(): array
...
+                );
+            }
+
+            $subscriberList[] = $subscriberItem;
+        }
+
+        $recipientIcons = [
+            'contact' => Icons::USER,
+            'contactgroup' => Icons::USERGROUP,
+            'schedule' => 'calendar',
```
nilmerg: `Icons::SCHEDULE`, please

nilmerg (changes requested):
One, very small, thing left. You can directly sqash and rebase now.

nilmerg (approved):
🎉

PR #1396, opened 2026-07-23
Allow assisted configuration of notifications

### LINKED ISSUE #1356: Allow to configure Icinga Notifications
The module configuration should allow to configure Icinga Notifications. This is only about an assisted configuration that only works by setting the environment up how we recommend it by default: On the same system, connected by a unix domain socket (UDS).

The database table `icingadb.icingadb_instance` provides several new columns to support this and guarantee reliably their validity:

* `icingadb_service_user` The user the icingadb daemon runs as
* `notifications_discovered_socket_path` The path to the UDS that Icinga Notifications is listening on

Icinga DB Web is supposed to use these to provide a single click configuration for the user to quickly set up the new notification component.

The *Database* tab should be a suitable location for this right now. Though, it should probably be renamed to *General Settings*. A toggle should provide this functionality.

Upon save it manages the source configuration in Icinga Notifications using [a procedure provided by Icinga Notificiations Web](https://github.com/Icinga/icinga-notifications-web/issues/463).

It also needs to configure Icinga DB itself and thus a [new table](https://github.com/Icinga/icingadb/issues/1108) is required. This is a key-value based table in which it is possible to insert entries that relate to environment variables used by the daemon to already load configuration with.

In an HA environment it will be required to configure both instances, in which case Icinga DB Web will find two entries in `icingadb_instance` for whose it needs to provide appropriate configuration pairs. This also means that two distinct source configurations might be required, unless the same `icingadb_service_user` is in use.

In case these settings are configured locally already, i.e. locked, the assisted configuration described here is not possible. It is also not possible if any of the two `icingadb_instance` columns are not available, which might be the case if the schema is outdated or if just no UDS was discovered. The toggle thus might be disabled and a reason must be shown alongside it.

Disabling the toggle must not cause the source in Icinga Notifications being deleted, it just remains as it is. Only the configuration of Icinga DB must be reset so that it stops connecting to the UDS.

BastianLedererIcinga:
resolve #1356

This PR introduces the _General Settings_ section, which contains the functionality of the previous _Database_ section and also allows to configure Icinga Notifications. `GeneralConfigForm` is introduced to support this.

For the configuration of Icinga Notifications a toggle is added which, allows to enable/disable the Icinga Notifications integration, as long as it runs on the same system as Icinga DB and the two are connected by a Unix domain socket.

To achieve this this PR relies on [a new table and modifications to the `instance` table](https://github.com/Icinga/icingadb/issues/1108).

Enabling the toggle simply writes the discovered socket path(s) to in `icingadb_config` and ensures Icinga Notifications has a matching source for each configured service user. Disabling it removes the rows from `icingadb_config`, but will not touch the sources.
Once row is written per endpoint found in `icingadb_instance`, if the `endpoint_id` is null a a default value of 20 zero bytes is used, matching the daemons fallback.

The toggle is disabled with a callout explaining why if:

- Instances of multiple environments exist
- The schema is outdated
- The daemon did not provide a socket path
- `icingadb_config` has a row with `locked` =  `y`, because a static configuration already exists.
- no responsible `icingadb_instance` is found at all
- `ICINGADB_NOTIFICATIONS_SYNCHRONIZE_WITH_DATABASE` is false
- No database is configure yet


application/controllers/ConfigController.php
```diff
@@ -28,11 +36,20 @@ public function databaseAction()
         $form = (new DatabaseConfigForm())
             ->setIniConfig(Config::module('icingadb'));
 
+        $notificationsForm = null;
+        if (Icinga::app()->getModuleManager()->hasEnabled('notifications')) {
```
sukhwinder33445: I would suggest to use `Module::exists('notifications')` instaed.

application/controllers/ConfigController.php
sukhwinder33445: Please rename the action, url and tab as well.
- `generalSettingsAction()`
- url: `'icingadb/config/general-settings'`
- Tab: `...->getConfigTabs()->activate('general-settings');`

application/controllers/ConfigController.php
```diff
@@ -28,11 +36,20 @@ public function databaseAction()
...
+            $notificationsForm = $this->createNotificationsForm();
+        }
+
         $form->handleRequest();
+        $notificationsForm?->handleRequest($this->getServerRequest());
 
         $this->mergeTabs($this->Module()->getConfigTabs()->activate('database'));
 
         $this->addFormToContent($form);
+        if ($notificationsForm !== null) {
```
sukhwinder33445: Why not add the form in a single condition?

```suggestion
// old code
if (Module::exists('notifications')) {
    // create and add the form here
}

Comment (sukhwinder33445) [reply application/controllers/ConfigController.php]: You can move the tab activation to the beginning of method body.

File: configuration.php [hunk 3]
```diff
@@ -545,8 +545,8 @@
     }
 
     $this->provideConfigTab('database', [
-        'label' => t('Database'),
-        'title' => t('Configure the database backend'),
+        'label' => t('General Settings'),
+        'title' => t('Configure general settings'),
         'url'   => 'config/database'
```
sukhwinder33445: Change the url please.

application/forms/NotificationsConfigForm.php
```diff
@@ -0,0 +1,123 @@
...
+use ipl\Web\Compat\CompatForm;
+use ipl\Web\Url;
+use ipl\Web\Widget\Callout;
+use ipl\Web\Widget\Icon;
+use ipl\Web\Widget\Link;
+
+class NotificationsConfigForm extends CompatForm
+{
+    use CsrfCounterMeasure;
+    use Translation;
```
sukhwinder33445, line 26: Please add `FormUid::createUidElement()` to distinguish this form, as there are two forms at the same address now.

application/forms/NotificationsConfigForm.php
```diff
@@ -0,0 +1,123 @@
...
+                fn (string $message) => HtmlElement::create('li', null, $message),
+                $this->lockMessages
+            ));
+
+            $this->addHtml(new Callout(
+                CalloutType::Info,
+                $reason,
+                $this->translate('Assisted configuration is not possible')
+            ));
+        }
```
sukhwinder33445: I'd suggest moving this before the description, so this info appears first and the description appears directly before the toggle.

library/Icingadb/Common/NotificationsIntegration.php
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @return void
+     */
+    public function enable(Instance $instance): void
+    {
+        $this->storeConfig($instance);
+        Source::get($instance->icingadb_service_user)
+            ->setName('Icinga DB')
+            ->setType('icinga2')
+            ->save();
```
sukhwinder33445: Is it acceptable if `$this->storeConfig($instance);` fails, but ‘Source’ is saved successfully?

library/Icingadb/Model/IcingaDBConfig.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @inheritDoc
+     */
+    public function getColumns(): array
+    {
+        return [
+            'key',
+            'value',
+            'locked',
+            'endpoint_id',
```
sukhwinder33445, line 48: Remove trailing comma please.

library/Icingadb/Model/Config.php
```diff
@@ -0,0 +1,63 @@
...
+use ipl\Orm\Behavior\Binary;
+use ipl\Orm\Behavior\BoolCast;
+use ipl\Orm\Behaviors;
+use ipl\Orm\Relations;
+
+/**
+ * @property string $key
+ * @property string $value
+ * @property bool $locked
+ * @property string $endpoint_id
```
sukhwinder33445, line 18: Add the relation please,  `@property Query|Instance $icingadb_instance`.

sukhwinder33445 (changes requested):
The general Settings tab now shows two forms; please add a separator to visually distinguish them. For example , see the `/account` page.

library/Icingadb/Common/NotificationsIntegration.php
```diff
@@ -0,0 +1,100 @@
...
+        $key = $this->db->quoteIdentifier('key');
+        $condition = [
+            $key . ' = ?'     => static::CONFIG_KEY,
+            'endpoint_id = ?' => $this->encodeEndpointId($instance->endpoint_id),
+        ];
+
+        $exists = $this->db->select(
+            (new Select())->from('icingadb_config')->columns([$key])->where($condition)
+        )->fetch();
+        if ($exists) {
```
sukhwinder33445: Any reason you are not using the model here `IcingaDBConfig::on($this->db)`

BastianLedererIcinga: Good idea, we can even use the relation on the `$instance` parameter directly.

library/Icingadb/Common/NotificationsIntegration.php
BastianLedererIcinga: Yes I think it is, but it's not what would currently happen.

If `storeConfig()` throws a `PDOException` we never save `Source`, so nothing is persisted to the db at all, which is fine since the Error shows something went wrong.

The more problematic case is what happens when `storeConfig()` runs succesfully, but `Source::save()` fails.
Icinga DB is correctly configured, but Icinga Notifications has no Source for it.

The error is displayed correctly, but the inserted row in `icingadb_config` is what drives the state of the toggle. So after a page reload, the toggle will be shown in active state, despite the fact that there was no source created.

So I think the best behavior is to first run `Source::save()` and then `storeConfig()`.
If `Source::save()` succeeds but `storeConfig()` does not, Icinga Notifications has the correctly configured source, and Icinga DB has no row added to `icingadb_config`.
This is the same state we have after running `NotificationsIntegration::disable()`.

application/forms/NotificationsConfigForm.php
```diff
@@ -0,0 +1,129 @@
...
+            ]
+        );
+
+        if (! $this->locked) {
+            $this->addElement('submit', 'save', ['label' => $this->translate('Save')]);
+        }
+
+        $this->addElement($this->createUidElement());
+        $this->addCsrfCounterMeasure(Session::getSession()->getId());
+    }
```
sukhwinder33445: Please add these above the ‘Submit’ button; I suggest adding them as the first elements, as we have done in most forms.

application/controllers/ConfigController.php
```diff
@@ -61,4 +78,75 @@ protected function mergeTabs(Tabs $tabs): self
...
+
+        $form->setEnabled($toggleState)
+            ->on(
+                NotificationsConfigForm::ON_SUBMIT,
+                function (NotificationsConfigForm $form) use ($instances, $integration) {
+                    $enable = $form->getValue('notifications-enabled') === 'y';
+
+                    try {
+                        foreach ($instances as $instance) {
+                            $enable ? $integration->enable($instance) : $integration->disable($instance);
```
sukhwinder33445: As discussed offline, I'd suggest writing the code directly and dropping the `NotificationsIntegration` class. It looks a bit superfluous to me.

library/Icingadb/Model/IcingaDBConfig.php
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * @property string $key
+ * @property string $value
+ * @property bool $locked
+ * @property string $endpoint_id
+ *
+ * @property Query|Instance $icingadb_instance
+ */
+class IcingaDBConfig extends Model
```
sukhwinder33445: Please rename the class to `Config`.

application/forms/NotificationsConfigForm.php
```diff
@@ -0,0 +1,129 @@
...
+
+class NotificationsConfigForm extends CompatForm
+{
+    use CsrfCounterMeasure;
+    use FormUid;
+    use Translation;
+
+    protected $defaultAttributes = [
+        'class' => 'icinga-form icinga-controls notifications-config-form',
+        'name'  => 'notifications-config-form',
```
sukhwinder33445: Please remove the trailing comma.

application/forms/NotificationsConfigForm.php
```diff
@@ -0,0 +1,129 @@
...
+use ipl\Web\Widget\Link;
+
+class NotificationsConfigForm extends CompatForm
+{
+    use CsrfCounterMeasure;
+    use FormUid;
+    use Translation;
+
+    protected $defaultAttributes = [
+        'class' => 'icinga-form icinga-controls notifications-config-form',
```
sukhwinder33445: Please use array for classes: `'class' =>  ['icinga-form', 'icinga-controls', 'notifications-config-form']`

This makes replacing or removing individual classes easier.

application/forms/NotificationsConfigForm.php
```diff
@@ -0,0 +1,129 @@
...
+        'name'  => 'notifications-config-form',
+    ];
+
+    /** @var bool Whether notifications are currently enabled */
+    private bool $enabled = false;
+
+    /** @var bool Whether the configuration is locked, because it is statically configured */
```
sukhwinder33445: Please fix the PHPDoc.

sukhwinder33445 (dismissed):

sukhwinder33445 (commented):
Looks good so far.

I'll approve this PR as soon as the IcingaDB part has been implemented and this branch can be properly tested.

library/Icingadb/Model/Config.php
```diff
@@ -0,0 +1,63 @@
...
+ * @property string $env_value
+ * @property bool $locked
+ * @property string $endpoint_id
+ * @property string $environment_id
+ */
+class Config extends Model
+{
+    /**
+     * @inheritDoc
+     */
```
nilmerg: Please drop all doc-comments like this and read-up on what `@inheritDoc` really does to understand why.

library/Icingadb/Model/Config.php
```diff
@@ -0,0 +1,63 @@
...
+use ipl\Orm\Behavior\BoolCast;
+use ipl\Orm\Behaviors;
+use ipl\Orm\Relations;
+
+/**
+ * @property string $env_key
+ * @property string $env_value
+ * @property bool $locked
+ * @property string $endpoint_id
+ * @property string $environment_id
```
nilmerg, line 20: Please document the relation property too.

application/controllers/ConfigController.php
nilmerg, line 1: Two separate forms, with differently labelled submit buttons, I expected more. Not only because there are two sets of submit buttons, but also due to a very specific risk:

Upon initial setup, which button is the user supposed to click first? **Not** the second one.

So, due to UX reasons, a single submit button must be present only. Use fieldsets to separate the two sections. And yes, I'm aware that this implies to refactor the `DatabaseConfigForm` to `ipl\Html\Form`. But there's an alternative now which comes in handy: `\Icinga\Web\Form\ConfigForm`

This in turn implies a dependency to Icinga Web v2.14, but that's acceptable as there are other reasons that may require this raise.

Since `\Icinga\Web\Form\ConfigForm` relies on `FormElements::getValues()` and complains in case of arrays (by using fieldsets) this may only be an option for the notifications configuration (the fieldset must be `ignored`) and the database dropdown must be added normally.

So please introduce a new form instead which combines both: `GeneralConfigForm`

It should also include the routines to lock the form as required and to store the settings in the database. The controller shouldn't contain this much logic.

nilmerg (changes requested):

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,349 @@
...
+
+    protected function onSuccess(): void
+    {
+        $this->saveNotificationsConfig();
+        parent::onSuccess();
+    }
+
+    protected function assemble(): void
+    {
+        $this->addCsrfCounterMeasure(Session::getSession()->getId());
```
nilmerg:
```suggestion
        $this->setCsrfCounterMeasureId(Session::getSession()->getId());
```

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,349 @@
...
+    {
+        $configuredResource = $this->config->get('icingadb', 'resource');
+        if (Str::isEmpty($configuredResource)) {
+            $this->lockNotifications(t('No database has been configured yet.'));
+
+            return;
+        }
+
+        if ($this->getPopulatedValue('icingadb__resource', $configuredResource) !== $configuredResource) {
+            $this->lockNotifications(t('The database configuration has changed. Please save it first.'));
```
nilmerg: Can we avoid this? `\Icinga\Module\Icingadb\Common\Backend::setDb` exists for a reason.

nilmerg (changes requested):

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,349 @@
...
+
+        if (Module::exists('notifications')) {
+            $this->readNotificationsConfig();
+            $this->addNotificationsSection();
+        }
+    }
+
+    private function addDatabaseSection(): void
+    {
+        $this->addHtml(HtmlElement::create('h1', null, t('Database')));
```
nilmerg: `h2` please.

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,349 @@
...
+            )),
+            Text::create(' '),
+            new Link(
+                [t('Consult the documentation for more information'), ' ', new Icon('arrow-up-right-from-square')],
+                Url::fromPath(
+                    'https://icinga.com/docs/icinga-db/latest/doc/03-Configuration/#notifications-configuration'
+                ),
+                ['target' => '_blank']
+            )
+        ]));
```
nilmerg: Fieldsets have a legend and a description, to facilitate this the constructor has to call `\ipl\Web\Compat\CompatForm::applyDefaultElementDecorators`

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,349 @@
...
+                            'ul',
+                            ['class' => 'lock-reasons'],
+                            array_map(
+                                fn(string $message) => HtmlElement::create('li', null, $message),
+                                $this->lockMessages
+                            )
+                        ),
+                    t('Assisted configuration is not possible')
+                )
+            );
```
nilmerg, line 161: should be part of the fieldset's description

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,349 @@
...
+use ipl\Web\Widget\Callout;
+use ipl\Web\Widget\Icon;
+use ipl\Web\Widget\Link;
+use RuntimeException;
+use Throwable;
+
+class GeneralConfigForm extends ConfigForm
+{
+    use CsrfCounterMeasure;
+    use Database;
```
nilmerg: The base form uses the `Translation` trait so please replace all `t()` usages.

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,349 @@
...
+use ipl\Web\Url;
+use ipl\Web\Widget\Callout;
+use ipl\Web\Widget\Icon;
+use ipl\Web\Widget\Link;
+use RuntimeException;
+use Throwable;
+
+class GeneralConfigForm extends ConfigForm
+{
+    use CsrfCounterMeasure;
```
nilmerg: that's already used by the the base form

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,349 @@
...
+        try {
+            if (! Backend::supportsNotifications()) {
+                $this->lockNotifications(
+                    t('The Icinga DB schema is outdated. Please update to configure notifications.')
+                );
+
+                return;
+            }
+
+            $instances = Instance::on($this->getDb())->columns(
```
nilmerg, line 227: I feel this should only fetch the `responsible` instance. Otherwise it may cause false-negative or false-positive results. Of course, this questions `$this->serviceUsers` etc. but that's fine.

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,349 @@
...
+
+                if ($instance->notifications_discovered_socket_path === null) {
+                    $this->lockNotifications(t('No Unix socket for Icinga Notifications was discovered.'));
+                }
+
+                if ($this->environmentId === null) {
+                    $this->environmentId = $instance->environment_id;
+                } elseif ($instance->environment_id !== $this->environmentId) {
+                    $this->lockNotifications(t('The Icinga DB instances belong to different environments.'));
+                }
```
nilmerg: Why is this a reason to lock the configuration?

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,349 @@
...
+
+                $endpointId = $instance->endpoint_id ?? $this->defaultEndpointId();
+                $socketPath = $instance->notifications_discovered_socket_path;
+
+                // Instances sharing an endpoint id also share a single config row, hence they must agree on its content
+                if (isset($this->endpoints[$endpointId]) && $this->endpoints[$endpointId] !== $socketPath) {
+                    $this->lockNotifications(t('The Icinga DB instances are not configured identically.'));
+                }
```
nilmerg: Surprised to see this and need an explanation for it before I can form an opinion.

public/css/form/general-config-form.less
nilmerg, line 1: not sure any of this still needed once the form uses the decorators properly.

nilmerg (changes requested):

application/forms/GeneralConfigForm.php
BastianLedererIcinga: I can't think of one in the current implementation, I suppose I forgot to remove it same as the comment below.

application/forms/GeneralConfigForm.php
BastianLedererIcinga: As discussed the `FieldsetDecorator` does not support `ValidHtml` as `description`, so I only used it for the `legend` and left `description` as is.

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,377 @@
...
+                if ($instance->notifications_discovered_socket_path === null) {
+                    $this->lockNotifications(
+                        $this->translate('No Unix socket for Icinga Notifications was discovered.')
+                    );
+                }
+
+                $endpointId = $instance->endpoint_id ?? $this->defaultEndpointId();
+                $this->environments[$instance->environment_id][$endpointId] =
+                    $instance->notifications_discovered_socket_path;
+                $this->serviceUsers[$instance->icingadb_service_user] = true;
```
nilmerg: There can only be one *responsible* instance hence why I wrote that it questions `$this->serviceUsers` etc. You won't be able to cache what you'll find here to re-use it upon save. `saveNotificationsConfig` needs to re-issue the query and fetch all instances. I only suggested to filter for responsible here due to the fact that this shows the current state.

public/css/form/general-config-form.less
```diff
@@ -0,0 +1,23 @@
...
+
+  input[type="checkbox"]:checked + .toggle-switch.disabled .toggle-slider {
+    background-color: @gray-light;
+    border-color: @gray-light;
+
+    &::before {
+      background-color: @gray-lighter;
+      border-color: @gray-light;
+    }
+  }
```
nilmerg: That should be redundant, where did you get that from?

nilmerg (changes requested):
Managed to get this:

Image

And the reason is that there is no source configured due to a database change. The toggle should not be on in this case and an appropriate warning should appear.

We should also think about a notice that tells the user the configuration looks fine, although the connection is unhealthy.

public/css/form/general-config-form.less
BastianLedererIcinga: From `icingaweb2/public/css/icinga/forms.less`, which has styles for disabled controls, but they don't apply to a toggle in enabled state.

nilmerg: I'd say, it looks correct as it does without this, until https://github.com/Icinga/icingaweb2/issues/5234 is done.

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,409 @@
...
+            }
+
+            if ($enable) {
+                foreach (array_keys($serviceUsers) as $serviceUser) {
+                    $source = Source::get($serviceUser);
+                    if ($source->getName() === null) {
+                        $source->setName('Icinga DB');
+                    }
+
+                    $source->setType('icinga2')->save();
```
nilmerg: `save()` should only be called in case the transaction below succeeds.

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,409 @@
...
+            return;
+        }
+
+        $enable = $this->getValue('notifications')['enabled'] === 'y';
+
+        try {
+            $serviceUsers = [];
+            $endpoints = [];
+            $environmentId = null;
+            foreach (
```
nilmerg, line 325: use `$query->first()` please

application/forms/GeneralConfigForm.php
```diff
@@ -0,0 +1,409 @@
...
+                            )
+                        ),
+                    $this->translate('Assisted configuration is not possible')
+                )
+            );
+        }
+
+        if ($this->sourceMissing) {
+            $notifications->addHtml(
+                new Callout(
```
nilmerg, line 166: please include a hint that storing may fix this

nilmerg (changes requested):

application/forms/GeneralConfigForm.php
BastianLedererIcinga: As discussed I added a comment as this is required for the HA case.

nilmerg (approved):
😓

PR #1399, opened 2026-07-26
Add Configurable Sortcontrols

gbin2265:
Add custom sort controls to a host/service/hostgroup/servicegroup/... based on an ini file.

This avoids having to patch the controller source code with extra custom lines on every icingadb-web upgrade.

; Extra sort options for Icinga DB Web list views.
;
; Location:  /etc/icingaweb2/modules/icingadb/sortcontrols.ini
; Ownership: root:icingaweb2, mode 0660 (must be readable by the web server user)
;
; One section per sort option. The section name is only an identifier and must be
; unique within this file; it is never shown in the UI. Prefixing it with the view
; keeps related options together.
;
;   type   Comma separated list of views this option applies to. Either a controller
;          name (hosts, services, hostgroups, servicegroups, comments, downtimes,
;          notifications, contacts, contactgroups, history) or controller/action for
;          the lists inside a detail view (host/services, host/history, host/parents,
;          host/children, service/history, hostgroup/hosts, servicegroup/services).
;   sort   The full sort spec, exactly as it would appear in the ?sort= URL parameter.
;          Multiple columns are allowed, separated by commas.
;   title  Label shown in the "Sort By" dropdown. Optional; defaults to the sort spec.
;          Not translatable, since it does not end up in the .po catalogs.
;
; Options appear in the dropdown in the order they are listed here, after the ones
; that ship with the module.

; Example

[hostgroups_srv_critical_warning_unhandled]
type  = "hostgroups"
sort  = "services_critical_unhandled desc,services_warning_unhandled desc"
title = "Srv Unhandled Critial,Warning"

### DIFF
--- a/library/Icingadb/Web/Controller.php
+++ b/library/Icingadb/Web/Controller.php
@@ -34,12 +34,14 @@ use Icinga\Util\Environment;
 use Icinga\Util\Json;
 use ipl\Html\Html;
 use ipl\Html\ValidHtml;
+use ipl\Orm\Common\SortUtil;
 use ipl\Orm\Query;
 use ipl\Orm\UnionQuery;
 use ipl\Stdlib\Filter;
 use ipl\Web\Compat\CompatController;
 use ipl\Web\Control\LimitControl;
 use ipl\Web\Control\PaginationControl;
+use ipl\Web\Control\SortControl;
 use ipl\Web\Filter\QueryString;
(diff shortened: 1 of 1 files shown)


gbin2265:
I have created a new pull request for `ipl-web`.

This moves the `SortControl` implementation to a higher level, making it reusable by other modules as well.

See Pull Request: https://github.com/Icinga/ipl-web/pull/404

PR #1408, opened 2026-08-26
Correctly handle the `null`-return of `Query::first()`

nilmerg:


nilmerg:
That's bullshit. The *null coalesce* operator handles this fine already, counterintuitively IMHO…

